PR #5, opened 2018-09-26
Add union, equals

lavalamp:
(also starts intersection and difference but now I have to go to a
meeting)


apelisse:
I don't really have anything to complain about. I'm curious how you're going to create the sets `NewFromDiff`, `NewFromConfig` as they will introduce dependency on openapi.

/lgtm
/approve

PR #8, opened 2018-09-26
Minor cleanup

lavalamp:
Address feedback.

### DIFF
--- a/fieldpath/set_test.go
+++ b/fieldpath/set_test.go
@@ -192,41 +192,44 @@ func TestSetIntersectionDifference(t *testing.T) {
 		MakePathOrDie("bar", "c2"),
 	)
 
-	i := NewSet(
-		MakePathOrDie("a1"),
-		MakePathOrDie("foo", 1),
-		MakePathOrDie("b1", KeyByFields("name", value.StringValue("first"))),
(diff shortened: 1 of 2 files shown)


fieldpath/set_test.go
```diff
@@ -192,41 +192,44 @@ func TestSetIntersectionDifference(t *testing.T) {
...
-		MakePathOrDie("b2", KeyByFields("name", value.StringValue("first"))),
-		MakePathOrDie("bar", "c2"),
-	)
-
-	got = s2.Difference(s1)
-
-	if !got.Equals(s2DiffS) {
-		t.Errorf("s2 - s1: expected: \n%v\n, got \n%v\n", s2DiffS, got)
-	}
+	t.Run("intersection", func(t *testing.T) {
```
apelisse: Quite minor, but I'm not really convinced that this is a good use-case for `Run`. In general I'd keep tests as small and isolated as possible. Whatever.

lavalamp: Hopefully we can agree that this is better than importing ginkgo ;)

apelisse: :joy: Absolutely!

PR #9, opened 2018-09-28
Add: typed value & schema validation

lavalamp:
Test coverage of `typed` package is > 90%

Major thing lacking in the schema now is oneof/union support


typed/validate_test.go
```diff
@@ -0,0 +1,201 @@
...
+	"fmt"
+	"testing"
+
+	"github.com/kubernetes-sigs/structured-merge-diff/schema"
+	"github.com/kubernetes-sigs/structured-merge-diff/value"
+
+	"gopkg.in/yaml.v2"
+)
+
+type validationTestCase struct {
```
apelisse: Yeah, I suspect you could have more dedicated unit-tests

typed/validate_test.go
```diff
@@ -0,0 +1,201 @@
...
+	name:         "associative list",
+	rootTypeName: "myRoot",
+	schema: `types:
+- name: myRoot
+  struct:
+    fields:
+    - name: list
+      type:
+        namedType: myList
+    - name: list2
```
apelisse: maybe give a more explicit name? "set"?

typed/validate_test.go
```diff
@@ -0,0 +1,201 @@
...
+		`{"key":"foo","value":true}`,
+		`{"key":"foo","value":true}`,
+	},
+	invalidObjects: []string{
+		`{"key":true,"value":1}`,
+		`{"key":1,"value":{}}`,
+		`{"key":false,"value":null}`,
+		`{"key":null}`,
+		`{"key":[1, 2]}`,
+		`{"key":{"foo":true}}`,
```
apelisse: what about the lack of key?

apelisse: which makes me think, do you have a concept of "required"/"optional"?

typed/validate_test.go
```diff
@@ -0,0 +1,201 @@
...
+      type:
+        scalar: numeric
+- name: myValue
+  map:
+    elementType:
+      scalar: string
+`,
+	validObjects: []string{
+		`{"list":[]}`,
+		`{"list":[null]}`,
```
apelisse: that seems a wrong

typed/validate_test.go
```diff
@@ -0,0 +1,201 @@
...
+		`{"list":[{}]}`,
```
apelisse: associative keys are not required?

typed/validate.go
```diff
@@ -0,0 +1,267 @@
...
+
+	return []ValidationError{v.error("invalid atom")}
+}
+
+func (v validation) doScalar(t schema.Scalar, value value.Value) []ValidationError {
+	switch t {
+	case schema.Numeric:
+		if value.Float == nil && value.Int == nil {
+			// TODO: should the schema separate int and float?
+			return []ValidationError{v.error("expected numeric (int or float)")}
```
apelisse: Would it be useful to say what it had? "expected numeric, got %v"?

typed/validate.go
```diff
@@ -0,0 +1,267 @@
...
+	case value.Map != nil:
+		// OK
+	default:
+		return []ValidationError{v.error("expected struct")}
+	}
+
+	allowedNames := map[string]struct{}{}
+	m := *value.Map
+	for i := range t.Fields {
+		f := &t.Fields[i]
```
apelisse: Do you need the pointer? Could you `range _, f := range t.Fields {` instead?

typed/validate.go
```diff
@@ -0,0 +1,267 @@
...
+						Name:  fieldName,
+						Value: value.Value{Null: true},
+					})
+				}
+				return pe, nil
+			}
+			return pe, errors.New("associative list with keys has an element that's a non-map type")
+		}
+
+		switch {
```
apelisse: Not completely clear that we are in the "set" mode here, maybe add a comment (or separate into individual functions)

apelisse:
Also, is this going to tighten validation on objects and is it fine?

typed/validate.go
lavalamp: I don't need the pointer but I don't want to use the loop variable since a reference might outlive the loop iteration (in an error message).

typed/validate.go
lavalamp: will add

typed/validate.go
```diff
@@ -0,0 +1,267 @@
...
+type ValidationError struct {
+	Path  fieldpath.Path
+	Error string
+}
+
+func (ve ValidationError) FormatMessage() string {
+	return fmt.Sprintf("%s: %v", ve.Path, ve.Error)
+}
+
+// Validate returns an error with a list of every spec violation.
```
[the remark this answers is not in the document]
lavalamp: Keeping the constructor with the use. I can move it.

typed/validate_test.go
```diff
@@ -0,0 +1,201 @@
...
+	"github.com/kubernetes-sigs/structured-merge-diff/value"
+
+	"gopkg.in/yaml.v2"
+)
+
+type validationTestCase struct {
+	name           string
+	rootTypeName   string
+	schema         string
+	validObjects   []string
```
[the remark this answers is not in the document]
lavalamp: It's annoying to specify schemas, so this is optimized around minimizing that. Also I think it's important to specify both positive and negative cases for each schema.

lavalamp:
re: tightening: yes, and we need to tolerate non-conformant objects and we'll need to fix the validation tightening problem before we can turn this on everywhere. See: https://github.com/kubernetes/kubernetes/issues/64841#issuecomment-413035093

lavalamp:
Rearranged commits slightly. More tests.

typed/validate_test.go
```diff
@@ -0,0 +1,201 @@
...
+		`{"key":1,"value":{}}`,
+		`{"key":false,"value":null}`,
+		`{"key":null}`,
+		`{"key":[1, 2]}`,
+		`{"key":{"foo":true}}`,
+	},
+}, {
+	name:         "associative list",
+	rootTypeName: "myRoot",
+	schema: `types:
```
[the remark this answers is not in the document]
lavalamp: improved very slightly.

typed/validate.go
```diff
@@ -205,15 +219,15 @@ func listItemToPathElement(list schema.List, index int, child value.Value) (fiel
 	return pe, nil
 }
 
-func (v validation) doList(t schema.List, value value.Value) (errs []ValidationError) {
+func (v validation) doList(t schema.List, value value.Value) (errs ValidationErrori) {
```
apelisse: `ValidationErrori`?

lavalamp: Huh, I thought I pushed the fix for this before making the PR, weird.

PR #10, opened 2018-09-28
Add SetFromValue (sans schema)

lavalamp:


fieldpath/fromvalue.go
```diff
@@ -0,0 +1,112 @@
...
+
+// AssociativeListCandidateFieldNames lists the field names which are
+// considered keys if found in a list element.
+var AssociativeListCandidateFieldNames = []string{
+	"key",
+	"id",
+	"name",
+}
+
+// GuessBestListPathElement guesses whether item is an associative list
```
apelisse: In an ideal world, we would have great tool and people would do the right thing, and we wouldn't have to guess anything, because we would have all the information we need to do what must be done.

But we're not in an ideal world, is this a good intermediary state that brings us closer to one? I'm not completely convinced. Guessing usually involves lots of "bugs" and confusion that I'd rather avoid if necessary.

lavalamp: The important thing is, people can fix the bugs by providing a schema.

Most CRDs are currently schemaless; apply needs to do something, it can't fail completely for them.

apelisse: > The important thing is, people can fix the bugs by providing a schema.

fieldpath/set_test.go
```diff
@@ -95,7 +111,7 @@ func TestSetIterSize(t *testing.T) {
 		if s2.Size() != addedCount {
 			t.Errorf("added %v items to set, but size is %v", addedCount, s2.Size())
 		}
-		if addedCount > 0 != s2.Empty() {
```
apelisse: :-)

fieldpath/fromvalue_test.go
```diff
@@ -0,0 +1,78 @@
...
+func TestFromValue(t *testing.T) {
+	table := []struct {
+		objYAML string
+		set     *Set
+	}{
+		{`a: a`, NewSet(MakePathOrDie("a"))},
+		{`{"a": [{"a": null}]}`, NewSet(
+			MakePathOrDie("a", 0, "a"),
+		)}, {`{"a": [{"id": a}]}`, NewSet(
+			MakePathOrDie("a", KeyByFields("id", value.StringValue("a")), "id"),
```
apelisse: Looks good to me.

fieldpath/fromvalue_test.go
```diff
@@ -0,0 +1,78 @@
...
+			MakePathOrDie("a", KeyByFields("key", value.StringValue("a")), "key"),
+		)}, {`{"a": [{"name": "a", "key": "b"}]}`, NewSet(
+			MakePathOrDie("a", KeyByFields(
+				"name", value.StringValue("a"),
+				"key", value.StringValue("b"),
+			), "key"),
+			MakePathOrDie("a", KeyByFields(
+				"name", value.StringValue("a"),
+				"key", value.StringValue("b"),
+			), "name"),
```
apelisse: yes

fieldpath/fromvalue_test.go
```diff
@@ -0,0 +1,78 @@
...
+		)}, {`{"a": [5]}`, NewSet(
+			MakePathOrDie("a", 0),
```
apelisse: So we own the index, not the value?

PR #11, opened 2018-10-02
add typed version of fieldset

lavalamp:
Follow-up to #10 (merge that one first).

### DIFF
--- a/schema/elements.go
+++ b/schema/elements.go
@@ -74,6 +74,9 @@ type Struct struct {
 	// one or more fields in the above list. A given field from the above
 	// list may be referenced in exactly 0 or 1 places in the below list.
 	// Unions []Union `yaml:"unions,omitempty"`
+
+	// Atomic structs must not have individual items merged.
+	Atomic bool `yaml:"atomic,omitempty"`
 }
 
 // StructField pairs a field name with a field type.
--- a/typed/toset_test.go
+++ b/typed/toset_test.go
@@ -1,0 +1,235 @@
+/*
+Copyright 2018 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package typed
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/kubernetes-sigs/structured-merge-diff/fieldpath"
+	"github.com/kubernetes-sigs/structured-merge-diff/schema"
+	"github.com/kubernetes-sigs/structured-merge-diff/value"
+
+	"gopkg.in/yaml.v2"
+)
+
+type objSetPair struct {
+	object string
+	set    *fieldpath.Set
+}
+
+type fieldsetTestCase struct {
+	name         string
+	rootTypeName string
+	schema       string
+	pairs        []objSetPair
+}
+
+var (
+	// Short names for readable test cases.
+	_NS  = fieldpath.NewSet
+	_P   = fieldpath.MakePathOrDie
+	_KBF = fieldpath.KeyByFields
+	_SV  = value.StringValue
+	_BV  = value.BooleanValue
+	_IV  = value.IntValue
+	_FV  = value.FloatValue
+)
+
+var fieldsetCases = []fieldsetTestCase{{
+	name:         "simple pair",
+	rootTypeName: "stringPair",
+	schema: `types:
+- name: stringPair
+  struct:
+    fields:
+    - name: key
+      type:
+        scalar: string
+    - name: value
+      type:
+        untyped: {}
+`,
+	pairs: []objSetPair{
+		{`{"key":"foo","value":1}`, _NS(_P("key"), _P("value"))},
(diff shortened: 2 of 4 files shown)


typed/toset_test.go
```diff
@@ -0,0 +1,236 @@
...
+      type:
+        untyped: {}
+`,
+	pairs: []objSetPair{
+		{`{"key":"foo","value":1}`, _NS(_P("key"), _P("value"))},
+		{`{"key":"foo","value":{"a": "b"}}`, _NS(_P("key"), _P("value"))},
+		{`{"key":"foo","value":null}`, _NS(_P("key"), _P("value"))},
+		{`{"key":"foo"}`, _NS(_P("key"))},
+		{`{"key":"foo","value":true}`, _NS(_P("key"), _P("value"))},
+		{`{"key":"foo","value":true}`, _NS(_P("key"), _P("value"))},
```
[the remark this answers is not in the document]
apelisse:
```
> diff <(echo '{`{"key":"foo","value":true}`, _NS(_P("key"), _P("value"))}') <(echo '{`{"key":"foo","value":true}`, _NS(_P("key"), _P("value"))}')
>
```

Look similar to me :-)

typed/toset_test.go
```diff
@@ -0,0 +1,236 @@
...
+
+type fieldsetTestCase struct {
+	name         string
+	rootTypeName string
+	schema       string
+	pairs        []objSetPair
+}
+
+var (
+	// Short names for readable test cases.
```
apelisse: > readable

:joy:

typed/toset_test.go
```diff
@@ -0,0 +1,236 @@
...
+          elementType:
+            scalar: string
+          elementRelationship: associative
+    - name: setBool
+      type:
+        list:
+          elementType:
+            scalar: boolean
+          elementRelationship: associative
+    - name: setNumeric
```
apelisse: I like these names

typed/typed.go
```diff
@@ -59,13 +59,35 @@ func (tv TypedValue) Validate() error {
 	return errs
 }
 
+// ToFieldSet creates a set containing every leaf field mentioned in tv, plus
+// any validation errors. A (perhaps partial) field set is returned even if
+// there are validation errors.
```
apelisse: > A (perhaps partial) field set is returned even if there are validation errors.

Why?

typed/typed.go
```diff
@@ -59,13 +59,35 @@ func (tv TypedValue) Validate() error {
...
+func (tv TypedValue) ToFieldSet() (*fieldpath.Set, error) {
```
apelisse: Are the two methods different enough that we don't want to merge them? AFAIU, `err := tv.Validate()` is equivalent to `_, err := tv.ToFieldSet()`?

typed/toset_test.go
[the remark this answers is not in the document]
lavalamp: Fixed

schema/elements.go
```diff
@@ -74,6 +74,9 @@ type Struct struct {
 	// one or more fields in the above list. A given field from the above
 	// list may be referenced in exactly 0 or 1 places in the below list.
 	// Unions []Union `yaml:"unions,omitempty"`
+
+	// Atomic structs must not have individual items merged.
+	Atomic bool `yaml:"atomic,omitempty"`
```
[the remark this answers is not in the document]
lavalamp: I think we should change these to be an enum to match fields before we use this?

typed/typed.go
lavalamp: To let us roll this out gradually even if there's old objects that don't validate against the schema. Open to better ways to do this.

I guess we could require objects to validate against the schema if you're using the new apply?

typed/typed.go
lavalamp: I think it's good to keep the interfaces separate, so we know the caller's intent if we ever change the library.

apelisse: I don't disagree with that, you shouldn't have to copy the implementation though:
```
func (tv) Validate() err {
  _, err := tv.ToFieldSet()
  return err
}
```

typed/typed.go
apelisse: That sounds reasonable, yes. Also, I suspect people who use apply already go through client-side validation (you can skip it, but it's enabled by default).

schema/elements.go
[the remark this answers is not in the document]
apelisse: We don't know yet if it can be something else. I have read somewhere that we should avoid bools because they usually end-up being more than 2 states? :-)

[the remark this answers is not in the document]
lavalamp: Yes, it's a code smell.

typed/toset_test.go
```diff
@@ -0,0 +1,236 @@
...
+var (
+	// Short names for readable test cases.
+	_NS  = fieldpath.NewSet
+	_P   = fieldpath.MakePathOrDie
+	_KBF = fieldpath.KeyByFields
+	_SV  = value.StringValue
+	_BV  = value.BooleanValue
+	_IV  = value.IntValue
+	_FV  = value.FloatValue
+)
```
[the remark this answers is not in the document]
lavalamp: I have drunk the go coolaid on this one (concise names help readability).

typed/typed.go
lavalamp: There's some modicum of overhead to constructing a field set. I suppose I could make a constructor to reduce the duplication.

schema/elements.go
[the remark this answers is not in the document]
lavalamp: I'll change in a follow-up.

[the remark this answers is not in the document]
lavalamp: See #13

typed/typed.go
lavalamp: Fixed

PR #12, opened 2018-10-03
Rework test suite/framework to allow a sequence of actions

kwiesmueller:
Started my implementation of the multi action/vector test suite.
As I was not 100% sure about the test design this might have broken plans on how people write tests against the suite, so please give feedback on that.


kwiesmueller:
The current wip approach uses Sequence to store the initial and the expected state as well as Vectors to run against the state.
Vectors can fail on their own and are meant as actions on the current state possibly mutating it.

Currently Vector is an interface which might interfere with @lavalamp 's plan to supply tests as YAML. Got to evaluate/discuss that.
Aside from this it is quite simple to implement Vectors for common cases (apply, put, post) and then combine them in various sequences.

The previous implementation concept got removed in favor of the Vector interface which might not suit the previous concept (wasn't sure there).

Would be happy to talk about this in person or do a small coworking session to learn your plans for this cc. @lavalamp @apelisse .

testvectors/sequence.go
```diff
@@ -0,0 +1,108 @@
...
+)
+
+// YAMLObject is an object encoded in YAML.
+type YAMLObject string
+
+// SchemaDefinition is an object schema. (TODO: get correct type; for now
+// assume this is a yaml-formatted string that can be deserialized.)
+type SchemaDefinition string
+
+// Sequence is a testcase collection consisting of 1+n Vectors which are being
```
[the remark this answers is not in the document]
apelisse: Yeah, I don't know what a vector is. Vector sounds like a type, not like a variable.

testvectors/vector.go
```diff
@@ -0,0 +1,73 @@
...
+// in applying them to various implementations.
+type Vector interface {
+	Name() string
+	Valid() bool
+	// Run the Vector on cur with schema to return the resulting object and
+	// failing the test on unexpected conflicts
+	Run(t *testing.T, cur YAMLObject, schema SchemaDefinition) YAMLObject
+}
+
+// TestVector is an implementation of Vector which just returns the defined ReturnObject
```
apelisse: It's unclear what all of these things are going to be used for?

testvectors/sequence.go
```diff
@@ -0,0 +1,108 @@
...
+	SchemaName string `yaml:"schemaName"`
+
+	InitialState YAMLObject `yaml:"initialState"`
+	Vectors      []Vector   `yaml:"vectors"`
+
+	ExpectedState YAMLObject `yaml:"expectedState"`
+}
+
+// AppendTestVectors adds the given vectors to the global list.
+func (s *Sequence) AppendTestVectors(vectors ...Vector) {
```
apelisse: `AppendTestVectors`, but it doesn't take `TestVectors` as parameters?

testvectors/example.go
```diff
@@ -31,38 +26,31 @@ type:
...
+		ExpectedState: `items:
+		- a
+		- b
+		- e
+		- c
+		- d`,
 	}
 
-	AppendTestVectors(tests...)
+	exampleSequence.AppendTestVectors(&TestVector{
```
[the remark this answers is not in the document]
kwiesmueller: Tried to keep this from the previous setup as I didn't want to delete everything you did. But yes I totally agree.

testvectors/vector.go
kwiesmueller: Those are example/test implementations of the Vector (now Action) interface being used in [example.go](testvectors/example.go) to show how it works.

testvectors/example.go
```diff
@@ -16,13 +16,8 @@ limitations under the License.
...
 package testvectors
 
-import (
-	"fmt"
-
-	"gopkg.in/yaml.v2"
-)
-
 func init() {
+
```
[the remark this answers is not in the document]
kwiesmueller: used to doing some of those things. done.

testvectors/example.go
```diff
@@ -31,38 +26,31 @@ type:
...
-      - e
-      - c
-      - d
-`), &tests)
-	if err != nil {
-		panic(fmt.Sprintf("test cases defined wrong: %v", err))
+	exampleSequence := &Sequence{
+		Name:       "example",
+		SchemaName: "listordering-primitiveInlineList",
+		InitialState: `items:
```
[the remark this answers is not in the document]
kwiesmueller: If you look at it now, that I defined the actions inline it get's harder to read again.
So either we define the actions outside the sequence or (as you already said it would be preferred) get rid of the interface and make it yaml compatible again.
I still am not sure how to represent this in a good way and allow various different actions, though.

tests/framework/state.go
```diff
@@ -0,0 +1,44 @@
...
+
+// State of the current test in terms of live object
+type State struct {
+	Live           YAMLObject `yaml:"live"`
+	Implementation Implementation
+}
+
+// Apply the passed in object to the current state
+func (s *State) Apply(obj YAMLObject, workflow string, force bool) (err error) {
+	s.Live, err = s.Implementation.Apply(s.Live, obj, workflow, force)
```
lavalamp: Don't update live if there's an error? (all three functions)

tests/framework/state.go
```diff
@@ -0,0 +1,44 @@
...
+}
+
+// Update the current state with the passed in object
+func (s *State) Update(obj YAMLObject, workflow string) (err error) {
+	s.Live, err = s.Implementation.NonApply(s.Live, obj, workflow)
+	return err
+}
+
+// Patch the current state with the passed in object
+func (s *State) Patch(obj YAMLObject, workflow string) (err error) {
```
lavalamp: Suggest we omit this function in the first draft or actually do a json patch before sending obj to the implementation.

tests/framework/impl.go
```diff
@@ -0,0 +1,25 @@
...
+*/
+
+package framework
+
+// Implementation defines the interface required by the actual merge code to be used in Kubernetes and this test framework
+type Implementation interface {
+	// Apply returns the new post merge object and errors (including conflicts if any occur)
+	Apply(live, config YAMLObject, workflow string, force bool) (YAMLObject, error)
+	// NonApply returns the new object and errors (including conflicts if any occur)
+	NonApply(live, new YAMLObject, workflow string) (YAMLObject, error)
```
lavalamp: Let's rename this Update.

tests/framework/impl.go
```diff
@@ -0,0 +1,25 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package framework
+
+// Implementation defines the interface required by the actual merge code to be used in Kubernetes and this test framework
+type Implementation interface {
+	// Apply returns the new post merge object and errors (including conflicts if any occur)
+	Apply(live, config YAMLObject, workflow string, force bool) (YAMLObject, error)
```
lavalamp: Somehow we need to pass the schema to the implementation as well; we could construct it knowing the schema and pass `value.Value`s for the objects, or we could pass typed.TypedValue for the objects. Thoughts?

tests/examples/example_test.go
```diff
@@ -0,0 +1,60 @@
...
+// TestExample shows how to use the test framework
+func TestExample(t *testing.T) {
+	state := &framework.State{Implementation: &MockImplementation{}}
+
+	err := state.Apply(`
+list:
+- a
+- b
+- c
+`, "default", false)
```
lavalamp: Check error here?

tests/framework/state_test.go
```diff
@@ -0,0 +1,79 @@
...
+
+package framework_test
+
+import (
+	"testing"
+
+	"sigs.k8s.io/structured-merge-diff/tests/framework"
+)
+
+type MockImplementation struct {
```
lavalamp: might as well keep it private?

testvectors/example.go
```diff
@@ -31,38 +25,32 @@ type:
...
-      items:
-      - a
-      - b
-      - e
-      - c
-      - d
-`), &tests)
-	if err != nil {
-		panic(fmt.Sprintf("test cases defined wrong: %v", err))
+	exampleSequence := &Sequence{
```
lavalamp: Maybe we should drop the sequence concept completely and make this look more like the example test?

kwiesmueller: Oh sorry, this is the old one and I forgot to delete. So everything in testvectors would be thrown away for now in favor of tests/...

tests/framework/state_test.go
kwiesmueller: Right, will change.
It originally was but I wanted to try something and did not change it back.

tests/framework/state.go
kwiesmueller: Not sure why I did it that way but thinking about it again I agree. Especially for printing the live vs. new on errors inside the test this information would be helpful.

tests/framework/impl.go
kwiesmueller: Yeah, I thought about this too but wanted to build what we outlined in the doc first.

I tried to think about ways that might suite you. As this was what we outlined in the doc I thought setting the spec on the implementation when initializing would be best.
I have to look at the types you introduced and the merge code to have more of an opinion though.

tests/framework/conflicts.go
```diff
@@ -0,0 +1,74 @@
...
+	for _, co := range c {
+		errs = append(errs, co.Error())
+	}
+	return strings.Join(errs, ", ")
+}
+
+// CheckExpectedConflicts and return unexpected ones
+// TODO: decide how we want this to return. either fail fast and return all conflicts directly
+// or check/filter all conflicts and only return unexpected ones as error (or as Conflicts).
+// Also, do we want to fail if expected conflicts do not occur or should this be a different function like `CheckAllExpectedConflicts` and `CheckAnyExpectedConflicts`?
```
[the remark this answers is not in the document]
lavalamp: I think it's ok to merge leaving the questions in. I think they may become clear once we have the actual conflict type in :)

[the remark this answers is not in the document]
apelisse: Yeah, I'll submit something :-)

tests/framework/impl.go
lavalamp: That's fine for the moment at least. We can add it into the implementation.

tests/framework/state.go
```diff
@@ -0,0 +1,44 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package framework
+
+// YAMLObject is an object encoded in YAML.
+type YAMLObject string
```
lavalamp: We will probably swap this out for a value.Value, but that can come in a followup.

PR #13, opened 2018-10-03
Validation fixes

lavalamp:
Builds on #11; the last three commits are new.

### DIFF
--- a/schema/elements.go
+++ b/schema/elements.go
@@ -54,12 +54,28 @@ type Atom struct {
 
 // Scalar (AKA "primitive") has a single value which is either numeric, string,
 // or boolean.
+// TODO: split numeric into float/int? Something even more fine-grained?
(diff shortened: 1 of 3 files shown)


schema/elements.go
```diff
@@ -62,6 +63,21 @@ const (
 	Boolean = Scalar("boolean")
 )
 
+// ElementRelationship is an enum of the different possible relationships
+// between the elements of container types.
+type ElementRelationship string
```
apelisse: Topology? Didn't we spend time arguing about this word? :-)

lavalamp: Did we? Do you think "ElementTopology" is clearer to devs who haven't spent a bunch of time thinking about schemas?

apelisse: No forget it, it's fine like this.

PR #14, opened 2018-10-03
Update ownerclaims for apply and non-apply cases

jennybuckley:


fieldpath/fromdiff.go
```diff
@@ -0,0 +1,26 @@
...
+
+package fieldpath
+
+import (
+	"github.com/kubernetes-sigs/structured-merge-diff/value"
+)
+
+// SetFromDiff creates a set containing every leaf field that has a
+// different value between v and u.
+func SetFromDiff(v, u value.Value) *Set {
```
lavalamp: This doesn't need the schema?

merge/object.go
```diff
@@ -0,0 +1,41 @@
...
+*/
+
+package merge
+
+import (
+	"github.com/kubernetes-sigs/structured-merge-diff/fieldpath"
+	"github.com/kubernetes-sigs/structured-merge-diff/value"
+)
+
+type Version string
```
lavalamp: Godoc all the things.

merge/object.go
```diff
@@ -0,0 +1,41 @@
...
+	*fieldpath.Set
+	Version Version
+}
+
+type Converter interface {
+	Convert(object *Object, version Version) *Object
+}
+
+type Object struct {
+	value.Value
```
lavalamp: Typed?

merge/object.go
```diff
@@ -0,0 +1,41 @@
...
+}
+
+func (o *Object) OwnerSet() map[string]*VersionedSet          { return nil }
+func (o *Object) SetOwnerSet(owners map[string]*VersionedSet) {}
+func (o *Object) GroupVersion() Version                       { return Version("v1") }
```
lavalamp: I think the intention is that apiserver can construct one of these that works with kubernetes api objects?

Maybe use a fieldpath to locate the version field and another fieldpath to locate the workflow id map?

merge/object.go
```diff
@@ -0,0 +1,41 @@
...
+
+func ConflictsToError(conflicts map[string]*VersionedSet) error { return nil }
```
lavalamp: Define a type for conflicts? I don't understand where the map comes in.

merge/object.go
lavalamp: Or that stuff could go in the schema?

merge/object.go
```diff
@@ -0,0 +1,41 @@
...
+func MergeApplied(live, config *Object) *Object                 { return nil }
```
lavalamp: Should this take objects or typed.TypedValue?

merge/update.go
```diff
@@ -0,0 +1,90 @@
...
+	}
+
+	for owner, conflictSet := range conflicts {
+		owners[owner].Set = owners[owner].Set.Difference(conflictSet.Set)
+	}
+
+	return owners, nil
+}
+
+func (s *SetUpdater) NonApply(live, nw *Object, owner string) *Object {
```
lavalamp: Maybe Update, Put, or Patch?

merge/update.go
```diff
@@ -0,0 +1,90 @@
...
+limitations under the License.
+*/
+
+package merge
+
+import (
+	"github.com/kubernetes-sigs/structured-merge-diff/fieldpath"
+)
+
+type SetUpdater struct {
```
lavalamp: Applier? WorkflowOperations?

Godoc :)

merge/update.go
lavalamp: Hm, there's already an Update. Godoc comments would help a lot :)

merge/update.go
```diff
@@ -0,0 +1,90 @@
...
+func (s *SetUpdater) NonApply(live, nw *Object, owner string) *Object {
+	owners, _ := s.Update(live, nw, owner, true)
+	changed := fieldpath.SetFromDiff(live.Value, nw.Value)
+	removed := fieldpath.SetFromValue(live.Value).Difference(fieldpath.SetFromValue(nw.Value))
+	owners[owner].Set = owners[owner].Set.Union(changed).Difference(removed)
+	nw.SetOwnerSet(owners)
+	return nw
+}
+
+func (s *SetUpdater) Apply(live, config *Object, owner string, force bool) (*Object, error) {
```
lavalamp: is owner the same as workflow? Maybe use the same name everywhere (I vote for workflow but don't care all that much).

fieldpath/fromdiff.go
jennybuckley: It probably does need the schema eventually but this way it keeps consistency with SetFromValue.

merge/object.go
jennybuckley: I think the type can be inferred from the Value, if we look at the apiVersion field

jennybuckley: But seeing your comment below I like the idea of this being implemented in the apiserver better

merge/object.go
jennybuckley: map from workflowid to fields that it owns which are conflicting with the apply request

jennybuckley: So we can return the owner of the conflicting fields to the user

merge/update.go
kwiesmueller: Just asked this question to @apelisse as he was using owner in #25 
I could understand the favor for owner as it is easier to understand and does cover more cases (while workflow would sound weird for users or operators in the future.

apelisse:
This has been merged through a different PR (#32)

Issue #7, opened 2018-09-26
Turn on CI


lavalamp:
We should block merges for anything where `go test ...` doesn't pass.

mvladev:
I've tried to enable the CI, but unfortunately the code can't compile (missing external dependency)

```
I1004 20:15:29.264] Call:  /workspace/test-infra/jenkins/../scenarios/execute.py go test ./...
W1004 20:15:29.280] Run: ('go', 'test', './...')
W1004 20:15:29.338] value/unstructured.go:22:2: cannot find package "gopkg.in/yaml.v2" in any of:
W1004 20:15:29.339] 	/usr/local/go/src/gopkg.in/yaml.v2 (from $GOROOT)
W1004 20:15:29.339] 	/go/src/gopkg.in/yaml.v2 (from $GOPATH)
```

The `test-infra` build image is using go `v1.11.1`, so `go.mod` can be used.

lavalamp:
Thanks for the report, we'll definitely set up go.mod for this.

(now just to go figure out how that works...)

mvladev:
@lavalamp I've added that in #16

mvladev:
Hmm what should `test-ifra` call when it clones this repo runs the presubmit (PR) jobs?

Most of the other projects have some small bash script/Makefile:

- https://github.com/kubernetes/test-infra/blob/f00550bff994d708f4a9becb9d1fc78eec0dd9ef/config/jobs/kubernetes/cluster-registry/cluster-registry-config.yaml#L19 
- https://github.com/kubernetes/test-infra/blob/106203e89955131005390b269662fbce58dd63f2/config/jobs/GoogleCloudPlatform/k8s-multicluster-ingress/k8s-multicluster-ingress-config.yaml#L18-L26

lavalamp:
I really want to have zero bash in this repo. go test ./...?

mvladev:
This repo is available at https://k8s-testgrid.appspot.com/
More specifically - https://k8s-testgrid.appspot.com/sig-api-machinery-structured-merge-diff

Reopen this PR if needed.

lavalamp:
@mvladev, thanks a ton!!

PR #16, opened 2018-10-04
Rename packages to sigs.k8s.io/structured-merge-diff

mvladev:
`sigs.k8s.io` redirects to `github.com/kubernetes-sigs`

Added `go.mod` and `vendor` folder to resolve missing dependencies.

Prerequisite for #7

### DIFF
--- a/fieldpath/element.go
+++ b/fieldpath/element.go
@@ -21,7 +21,7 @@ import (
 	"sort"
 	"strings"
 
-	"github.com/kubernetes-sigs/structured-merge-diff/value"
+	"sigs.k8s.io/structured-merge-diff/value"
 )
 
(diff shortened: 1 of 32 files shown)


apelisse:
I'm not opposed to this change, but I'd like the canonical name to also bet set, to prevent double imports: https://golang.org/doc/go1.4#canonicalimports

lavalamp:
wow, nifty!

Antoine, I agree, but we can do that in a follow up and I think we'll want to change the name anyway.

/lgtm
/approve

apelisse:
Unrelated, but @mvladev if you want to be awesome, it'd be nice to set-up travis too ;-)

PR #18, opened 2018-10-09
Fix Set.Difference: parent removal affects children

lavalamp:
Reported in person by @apelisse.


lavalamp:
/hold

fixing this...

PR #20, opened 2018-10-10
Create Conflict type as an error

apelisse:


merge/conflict.go
```diff
@@ -0,0 +1,36 @@
...
+
+package merge
+
+import (
+	"fmt"
+
+	"sigs.k8s.io/structured-merge-diff/fieldpath"
+)
+
+// Conflict is a conflict on a specific field with the current owner of
```
lavalamp: Won't you want a list of conflicts as the error type?

apelisse: We could, we don't have too. I suspect the way we want to print the conflicts might lead that decision (though it doesn't have to, we can aggregate/split at any time). It might look like this:
```
conflict with 'random-id': .spec.replicas
conflict with 'random-id': .spec.strategy.type
```

or it could look like that:
```
conflict with 'random-id': 
- .spec.replicas
- .spec.strategy.type
```

Both look fine to me (I like the former better though). 

From a code standpoint, I'd keep each conflict as an individual entity.

Eventually, we'll want to create a list of conflicts from the intersection of owner fieldsets, which means we'll need a: `ConflictsFromFieldset(fieldset) []Conflict`.

apelisse:
Actually, what might be missing is the version of the object that we conflicted with.

lavalamp:
If you look at the validation errors I made, I had both.

apelisse:
Yeah, I don't exactly see the value of that right now, but sure

apelisse:
PTAL, I'm aggregating the conflicts, tell me what you think

merge/conflict.go
```diff
@@ -0,0 +1,91 @@
...
+		messages = append(messages, fmt.Sprintf("conflicts with %q:", owner))
+		for _, path := range m[owner] {
+			messages = append(messages, fmt.Sprintf("- %v", path))
+		}
+	}
+	return strings.Join(messages, "\n")
+}
+
+// NewFromSets creates a list of conflicts error from a map of owner to set of fields.
+func NewFromSets(sets map[string]*fieldpath.Set) Conflicts {
```
lavalamp: The map desperately needs to be a named type; doesn't have to happen in this PR.

merge/conflict_test.go
```diff
@@ -0,0 +1,56 @@
...
+			_P("list", _KBF("key", _SV("a"), "id", _IV(2)), "id"),
+		),
+		"Alice": _NS(
+			_P("value"),
+			_P("list", _KBF("key", _SV("a"), "id", _IV(2)), "key"),
+		),
+	})
+	wanted := `conflicts with "Alice":
+- .value
+- .list[id=2,key="a"].key
```
lavalamp: This test is a change detector if anything about the field path serialization changes :/

I guess it's not a huge problem, it just makes it hard to do things if there's a bunch of tests like this.

I'd probably just search the error message for the Alice & Bob lines if you want to make a followup, I don't feel super strongly.

PR #24, opened 2018-10-13
Untested way to get Value from a Path

apelisse:


fieldpath/element.go
```diff
@@ -70,6 +71,24 @@ func (e PathElement) String() string {
...
+	case e.FieldName != nil:
+		return v.Map.Follow(*e.FieldName)
+	case len(e.Key) > 0:
+		return v.List.FollowKeys(e.Key)
+	case e.Value != nil:
+		return v.List.FollowValue(*e.Value)
+	case e.Index != nil:
+		return v.List.FollowIndex(*e.Index)
+	default:
+		return value.Value{}, errors.New("invalid path element")
```
kwiesmueller: wrapping the error to provide path context would be helpful for testing and later user errors

value/value.go
```diff
@@ -44,11 +46,54 @@ type Field struct {
...
 	Items []Value
 }
 
+// FollowKeys returns the first value with the corresponding keys or an
+// error if no item can be found.
+func (l *List) FollowKeys(keys []Field) (Value, error) {
+	if l == nil {
+		return Value{}, errors.New("can't lookup keys in nil-list")
+	}
+	return Value{}, fmt.Errorf("couldn't find item for keys %v in list", keys)
```
kwiesmueller: Nothing happening here yet?

value/value.go
```diff
@@ -44,11 +46,54 @@ type Field struct {
...
+	if l == nil {
+		return Value{}, errors.New("can't lookup index in nil-list")
+	}
+	if index < 0 {
+		return Value{}, errors.New("can't lookup negative index in list")
+	}
+	if index >= len(l.Items) {
+		return Value{}, fmt.Errorf("index out of range: %d/%d", index, len(l.Items))
+	}
+	return l.Items[index], nil
```
kwiesmueller: return error if item not found?

kwiesmueller: nvm thought I am looking at map

fieldpath/element.go
```diff
@@ -70,6 +71,24 @@ func (e PathElement) String() string {
 	}
 }
 
+// Follow tries tries to find the value corresponding to the
+// PathElement. If it can't follow the PathElement, an error is
+// returned.
+func (e PathElement) Follow(v value.Value) (value.Value, error) {
+	switch {
+	case e.FieldName != nil:
+		return v.Map.Follow(*e.FieldName)
```
lavalamp: I'm not sure about this--
* Do we need this function on untyped values?
* Is it good to add a follow function to every type? (e.g., why not just call get here)
* It seems you'll need to verify the type of the value before blindly accessing something in the one-of
* ...so maybe even if we need this for untyped values, it should be implemented in the typed package?

apelisse: > Do we need this function on untyped values?

Hard to tell. Right now I mostly need it to get the version of the object.

> Is it good to add a follow function to every type? (e.g., why not just call get here)

I could have done it, this function would have been more complicated, while I can spread the complexity to each individual type (better encapsulation?). Not super strong.

> It seems you'll need to verify the type of the value before blindly accessing something in the one-of

This is still checking if it works or not. I'm checking that we do have a map when we're requesting a field etc. Also if we are acting on typed object, we maybe can assume that it's been validated before.

One of my problem is that I'd like this to work on value and typedvalue, and when we're using typedvalue, it'd be nice if the type information would come with it.

lavalamp: I'm still thinking it'd be better to put this in the typed package; it can still work on untyped objects (we can allow it to go inside untyped blobs).

lavalamp: I think the thing that mostly bothers me is that this functionality crosses package boundaries.

apelisse:
Yeah, I'm not happy with this either, still thinking about it, but I don't want to lose time doing that now. I'll come back to it later, thanks for your answer.

apelisse:
I don't think we want or need this anymore.

PR #25, opened 2018-10-15
Create signature for Updater

apelisse:
Create the signature for Updater, the object that will be used to update
the owners Fieldset and merge applied configurations.

### DIFF
--- a/merge/update.go
+++ b/merge/update.go
@@ -1,0 +1,52 @@
+/*
+Copyright 2018 The Kubernetes Authors.
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+    http://www.apache.org/licenses/LICENSE-2.0
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package merge
+
+import (
+	"sigs.k8s.io/structured-merge-diff/fieldpath"
+	"sigs.k8s.io/structured-merge-diff/typed"
+)
+
+// APIVersion describes the version of an object or of a fieldset.
+type APIVersion string
+
+// VersionedSet associates a version to a set.
+type VersionedSet struct {
+	*fieldpath.Set
+	APIVersion APIVersion
+}
+
+// Owners is a map from workflow-id to VersionedSet (what they own in
+// what version).
+type Owners map[string]*VersionedSet
+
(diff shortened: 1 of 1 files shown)


apelisse:
@lavalamp @kwiesmueller

merge/update.go
```diff
@@ -0,0 +1,53 @@
...
+// Updater is the object used to compute updated FieldSets and also
+// merge the object on Apply.
+type Updater struct{}
+
+// Update is the method you should call once you've merged your final
+// object on CREATE/UPDATE/PATCH verbs. newObject must be the object
+// that you intend to persist, and liveObject must be the original
+// object. owners is modified and needs to be injected in the object
+// before persisting.
+func (s *Updater) Put(liveObject, newObject typed.TypedValue, owners Owners, owner string) error {
```
kwiesmueller: Should be Update or the comment should be different

merge/update.go
```diff
@@ -0,0 +1,53 @@
...
+type Owners map[string]*VersionedSet
+
+// Updater is the object used to compute updated FieldSets and also
+// merge the object on Apply.
+type Updater struct{}
+
+// Update is the method you should call once you've merged your final
+// object on CREATE/UPDATE/PATCH verbs. newObject must be the object
+// that you intend to persist, and liveObject must be the original
+// object. owners is modified and needs to be injected in the object
```
lavalamp: Is it better to modify owners or construct & return a new one? I think most of the types in this repo behave better if treated as immutable. (same question on `Apply`, too)

merge/update.go
```diff
@@ -0,0 +1,53 @@
...
+// what version).
+type Owners map[string]*VersionedSet
+
+// Updater is the object used to compute updated FieldSets and also
+// merge the object on Apply.
+type Updater struct{}
+
+// Update is the method you should call once you've merged your final
+// object on CREATE/UPDATE/PATCH verbs. newObject must be the object
+// that you intend to persist, and liveObject must be the original
```
lavalamp: that you intend to persist `(after applying the patch if this is for a PATCH call)`, and liveObject must be the original object `(empty if this is a CREATE call)`.

merge/update.go
apelisse: Yeah, so we don't have a good mechanism to change the typedvalue right now, but I think it makes more sense to return the object with the owners field set already.

apelisse:
PTAL

kwiesmueller:
After looking at it again I wanted to ask why Owners isn't part of the Object being updated (which would make returning them easier as well).

apelisse:
> After looking at it again I wanted to ask why Owners isn't part of the Object being updated (which would make returning them easier as well).

I remember why now. The reason is that the types are not exactly the same. What we will store in the API is not the same type that we use to do the set operations. Which means we need a layer to convert from one to another, and I don't think that layer will live in this repository.

PR #26, opened 2018-10-16
Signigicant changes to the test framework

apelisse:
- Remove the conflicts comparison since it's easier to just use
reflect.DeepEqual to compare the list of errors.
- Use TypedValue as parameter for the state, since passing YAML doesn't
carry the schema information
- Add the Owners field that matches the implementation
- Hard-code the implementation rather than using a mock, since we don't
intend to have multiple implementations anyway.


tests/examples/example_test.go
```diff
@@ -19,47 +19,60 @@ package tests
...
 list:
 - a
 - b
 - c
-`, "default", false)
-	err = framework.CheckExpectedConflicts(err, nil)
+`, "lists")
+	if err != nil {
+		t.Fatalf("Failed to parse YAML: %v", err)
+	}
```
lavalamp: Idea for followup:
```
config := state.ParseOrDie(t, parser, `yaml string here`)
```

tests/framework/state.go
```diff
@@ -16,28 +16,35 @@ limitations under the License.
...
+	Updater *merge.Updater
 }
 
-// Apply the passed in object to the current state
-func (s *State) Apply(obj typed.YAMLObject, workflow string, force bool) error {
-	new, err := s.Implementation.Apply(s.Live, obj, workflow, force)
+// Update the current state with the passed in object
+func (s *State) Update(obj typed.TypedValue, owner string) error {
+	owners, err := s.Updater.Update(s.Live, obj, s.Owners, owner)
 	if err == nil {
```
lavalamp: Nit: It's more natural to do `if err != nil { return err }`

PR #27, opened 2018-10-16
Start a command line tool

lavalamp:


smd/main.go
```diff
@@ -0,0 +1,117 @@
...
+		return nil, fmt.Errorf("unable to read schema %q: %v", o.schemaPath, err)
+	}
+	base.parser, err = typed.NewParser(typed.YAMLObject(b))
+	if err != nil {
+		return nil, fmt.Errorf("schema %q has errors:\n%v", o.schemaPath, err)
+	}
+
+	if o.typeName == "" {
+		types := base.parser.Schema.Types
+		if len(types) == 0 {
```
apelisse: NIT: Should you test that in both cases? It's kind of surprising that this error only happens if you didn't specify the typename.

typed/parser.go
```diff
@@ -33,18 +33,19 @@ type YAMLParser interface {
 	FromYAML(object YAMLObject, typename string) (TypedValue, error)
 }
 
-type parser struct {
-	schema schema.Schema
+// Parser implements YAMLParser and allows introspecting the schema.
```
apelisse: You should probably drop the `YAMLParser` too then?

smd/main.go
lavalamp: They'll get a different error--whatever type they mention won't exist.

apelisse: Yeah, I understand that, I just think the inconsistency is sort of surprising

apelisse: Not a big deal :-)

typed/parser.go
lavalamp: It's OK to offer the interface (e.g. for testability). I can remove it though if you want.

apelisse: I created it because I wanted to hide the type more than because I wanted to fake it, so feel free to remove it

smd/main.go
lavalamp: I could force people to always state the type, but I thought this was nicer.

typed/parser.go
lavalamp: ok one second.

lavalamp: done

apelisse:
great!
/lgtm

PR #28, opened 2018-10-17
Add compare and merge commands

lavalamp:


lavalamp:
I have to say, the test story in this repo is 1000x better than in the main repo.

apelisse:
> I have to say, the test story in this repo is 1000x better than in the main repo.
:joy:

smd/main.go
```diff
@@ -22,27 +22,52 @@ import (
...
+	"io"
 	"io/ioutil"
 	"log"
+	"os"
 
 	"sigs.k8s.io/structured-merge-diff/typed"
 )
 
+var (
+	ErrTooManyOperations = errors.New("exactly one of --merge, --compare, or --validate must be provided")
```
apelisse: Just my opinion, but you could have created sub-commands rather than options

smd/main.go
```diff
@@ -77,26 +102,140 @@ func (o *options) resolve() (operation, error) {
 
 	switch {
 	case o.validatePath != "":
+		if o.merge || o.compare {
```
apelisse: The fact that you're using the case statement in a specific order is sort of confusing (certainly is very counter-intuitive to me since I don't see a switch statement as a succession of if conditions). I don't see how this is different from a `if else if else if`?

smd/main.go
```diff
@@ -77,26 +102,140 @@ func (o *options) resolve() (operation, error) {
...
 	}
-	_, err = v.parser.FromYAML(typed.YAMLObject(b), v.typeName)
+
+	got, err := lhs.Compare(rhs)
 	if err != nil {
-		return fmt.Errorf("unable to validate file %q:\n%v", v.fileToValidate, err)
+		return err
+	}
+
+	if got.Added.Empty() && got.Modified.Empty() && got.Removed.Empty() {
```
apelisse: unrelated but this is done a lot in tests, we might want to add a method like `Comparison.AreSame() bool` maybe

lavalamp:
I can split it up.

I don't like Cobra and I also don't want to add another dependency.

kwiesmueller:
Yeah me neither, but wanted to make sure we are not bypassing any standards I don't know about :innocent:

smd/main.go
lavalamp: Yeah, that starts to sound like work, this was just to make it easy to explore with files...

smd/main.go
lavalamp: done

smd/main.go
lavalamp: improved slightly

lavalamp:
PTAL

lavalamp:
BTW the deployment doesn't validate because there's errors in the spec:

```
$ go run ./smd/main.go --schema=smd/testdata/k8s-schema.yaml --type-name=io.k8s.api.apps.v1beta1.Deployment --validate=smd/testdata/k8s-deployment.yaml
2018/10/18 16:07:37 Couldn't execute operation: unable to validate file "smd/testdata/k8s-deployment.yaml":
errors:
  .spec.strategy.rollingUpdate.maxUnavailable: expected string, got 0
  .spec.template.spec.containers[name="kubedns"].livenessProbe.httpGet.port: expected string, got 10054
  .spec.template.spec.containers[name="kubedns"].ports: duplicate entries for key [containerPort=10053]
  .spec.template.spec.containers[name="kubedns"].readinessProbe.httpGet.port: expected string, got 8081
  .spec.template.spec.containers[name="dnsmasq"].livenessProbe.httpGet.port: expected string, got 10054
  .spec.template.spec.containers[name="dnsmasq"].ports: duplicate entries for key [containerPort=53]
  .spec.template.spec.containers[name="sidecar"].livenessProbe.httpGet.port: expected string, got 10054
exit status 1
```

lavalamp:
Fixed int or string, now just down to the two duplicate key complaints. I'll have to actually fix the schema to fix that one :/

I think this is ready for merging.

PR #32, opened 2018-10-18
Add update code

apelisse:
Improve a few things in the test framework and add the code for the updater. This is still mostly missing tests, but the goal is that we can share that duty as soon as possible.


lavalamp:
lgtm, can you maybe squash down e.g. the fixup commits (the last two at least)?

apelisse:
> lgtm, can you maybe squash down e.g. the fixup commits (the last two at least)?

Sure!

apelisse:
squashed the last two1

PR #33, opened 2018-10-18
update readme

lavalamp:
This was an item on our task list. I don't know what we meant, but I guess this is an improvement.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,8 +1,55 @@
 # Structured Merge and Diff
 
-This repo contains code for implementing the Kubernetes "apply" operation.
+This repo contains code which implements the Kubernetes "apply" operation.
 
-It also contains test vectors for testing an implementation of apply.
+## What is the apply operation?
+
+We model resources in a control plane as having multiple "managers". Each
+manager is typically trying to manage only one aspect of a resource. The goal is
+to make it easy for disparate managers to make the changes they need without
+messing up the things that other managers are doing. In this system, both humans
+and machines (aka "controllers") act as managers.
+
+To do this, we explicitly track (using the fieldset data structure) which fields
+each manager is currently managing.
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,8 +1,54 @@
...
+For APPLY, the user is explicitly stating which fields they wish to manage (and
+therefore requesting deletion of any fields that they used to manage but stop
+mentioning).
+
+Any time a manager begins managing some new field, that field is removed from
+all other managers. If the manager is using the APPLY command, we call these
+conflicts, and will not proceed unles the user passes the "force" option. This
+prevents accidentally setting fields which some other entity is managing.
+
+PUT/PATCH always "force". They are mostly used by automated systems, which won't
```
apelisse: s/won't/wouldn't/ ?

README.md
```diff
@@ -1,8 +1,54 @@
...
+do anything productive with a new error type.
+
+## Components
+
+The operation has a few building blocks:
+
+* We define a simple schema type in the schema package. (Vastly simpler than
```
apelisse: I don't hate the word "simple", but I think "targeted" is more accurate, so I'd replace that part in parenthesis to explain why you think it's simple "because it's targeted to our specific need"

README.md
```diff
@@ -1,8 +1,54 @@
...
+  OpenAPI.)
+* We define a "field set" data structure, in the fieldpath package. A field path
+  locates a "leaf" field in an object. A field set is a group of such paths.
```
apelisse: > A field path locates a "leaf" field in an object.

Not entirely true

README.md
lavalamp: I think it's correct as written--this is a statement of fact about the automated systems that already exist, not a subjunctive hypothesis.

lavalamp:
Updated, PTAL

PR #34, opened 2018-10-19
add tests for inlineList from design doc

kwiesmueller:
Tests as described here:
https://docs.google.com/document/d/1-SIcJArYuFE99NbfiOJrkfxCYuLTb9wmbS0HnPMYLEo/edit#heading=h.ybed98pe7iai

### DIFF
--- a/merge/list_test.go
+++ b/merge/list_test.go
@@ -1,0 +1,282 @@
+package merge_test
+
+import (
+	"testing"
+
+	"sigs.k8s.io/structured-merge-diff/typed"
+
+	"sigs.k8s.io/structured-merge-diff/merge"
+)
+
+func TestInlineListMerge(t *testing.T) {
+	parser, err := typed.NewParser(`types:
+- name: lists
+  struct:
+    fields:
+    - name: items
+      type:
+        list:
+          elementType:
+            scalar: string
+          elementRelationship: associative`)
+	if err != nil {
+		t.Fatalf("Failed to create parser: %v", err)
+	}
+
+	t.Run("disjointItemsOrder", func(t *testing.T) {
+		state := &State{
+			Updater:  &merge.Updater{},
+			Parser:   parser,
+			Typename: "lists",
+		}
+
+		lhs := typed.YAMLObject(`
+items:
+- a
+- b
+- c
+`)
+
+		rhs := typed.YAMLObject(`
+items:
+- d
+- e
+`)
+
+		output := typed.YAMLObject(`
+items:
+- a
+- b
+- c
+- d
+- e
+`)
+
+		twoWayMerge(t, state, lhs, rhs, output)
+	})
+
+	t.Run("overlappingItemsOrder", func(t *testing.T) {
+		state := &State{
+			Updater:  &merge.Updater{},
+			Parser:   parser,
+			Typename: "lists",
+		}
+
+		lhs := typed.YAMLObject(`
+items:
+- a
+- b
+- c
+`)
+
+		rhs := typed.YAMLObject(`
+items:
+- a
+- b
+- d
+`)
+
+		output := typed.YAMLObject(`
+items:
+- a
+- b
+- c
+- d
+`)
+
+		twoWayMerge(t, state, lhs, rhs, output)
+	})
+
+	t.Run("disjointOverlappingItemsOrder", func(t *testing.T) {
+		state := &State{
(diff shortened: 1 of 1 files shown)


merge/list_test.go
```diff
@@ -0,0 +1,282 @@
...
+- c
+- d
+- b
+- e
+`)
+
+		twoWayMerge(t, state, lhs, rhs, output)
+	})
+
+	t.Run("orderConstraintCheck", func(t *testing.T) {
```
kwiesmueller: @apelisse @lavalamp this test is currently failing, not sure why:
`sigs.k8s.io/structured-merge-diff/merge/list_test.go:272: Wanted err = <nil>, got .items[=true]: expected string, got true`

merge/list_test.go
```diff
@@ -0,0 +1,282 @@
...
+      type:
+        list:
+          elementType:
+            scalar: string
+          elementRelationship: associative`)
+	if err != nil {
+		t.Fatalf("Failed to create parser: %v", err)
+	}
+
+	t.Run("disjointItemsOrder", func(t *testing.T) {
```
apelisse: I'd write a different test rather than abusing the `t.Run` thing maybe.

merge/list_test.go
```diff
@@ -0,0 +1,282 @@
...
+		state := &State{
+			Updater:  &merge.Updater{},
+			Parser:   parser,
+			Typename: "lists",
+		}
+
+		lhs := typed.YAMLObject(`
```
apelisse: lhs and rhs make little sense in this package I think, since it's about apply multiple operations in a row.

merge/list_test.go
```diff
@@ -0,0 +1,282 @@
...
+		output := typed.YAMLObject(`
+items:
+- a
+- b
+- c
+- d
+- e
+`)
+
+		twoWayMerge(t, state, lhs, rhs, output)
```
apelisse: I'd rather have the code copy-pasted and obvious than have the logic hidden somewhere else

merge/list_test.go
```diff
@@ -0,0 +1,282 @@
...
+- b
+- c
+`)
+
+		twoWayMerge(t, state, lhs, rhs, output)
+	})
+}
+
+func twoWayMerge(t *testing.T, state *State, lhs, rhs, output typed.YAMLObject) {
+	err := state.Apply(lhs, "default", false)
```
apelisse: Applying twice in a row is definitely a legitimate use-case, though I don't know if I'd call it "twoWayMerge". Ideally, I think we'd also want to check if Owners is what we would expect.

merge/list_test.go
apelisse: Yeah I don't get that error ...

merge/list_test.go
kwiesmueller: I thought that's what its meant for as it allows grouping and produces the according output:
```
--- FAIL: TestInlineListMerge (0.01s)
    --- FAIL: TestInlineListMerge/orderConstraintCheck (0.00s)
        /home/kwiesmueller/go/src/sigs.k8s.io/structured-merge-diff/merge/list_test.go:272: Wanted err = <nil>, got .items[=true]: expected string, got true
```

apelisse: I mostly use these for table-driven tests, but sure why not

merge/list_test.go
```diff
@@ -0,0 +1,282 @@
...
+- z
+- b
+- c
+`)
+
+		twoWayMerge(t, state, lhs, rhs, output)
+	})
+}
+
+func twoWayMerge(t *testing.T, state *State, lhs, rhs, output typed.YAMLObject) {
```
lavalamp: I think the way to interpret those tests is that lhs == initial condition and rhs == a thing that is applied. Which is effectively two applies in a row, exactly like you've written here. But the names are confusing in this context. So maybe:

s/twoWayMerge/singleApplyTest/
s/lhs/initialObject/
s/rhs/appliedObject/

apelisse: I guess that would work yes.

merge/list_test.go
lavalamp: Is the yaml parser interpreting a `y` as a boolean? If you add quotes around the `y` does it work?

apelisse:
I'm closing this since I think we have a better way to test these things, though the content might still be interesting.

PR #35, opened 2018-10-22
Move Owners logic to its file, add Difference

apelisse:
Create the difference method to simplify comparison in tests.

### DIFF
--- a/merge/owners.go
+++ b/merge/owners.go
@@ -1,0 +1,75 @@
+/*
+Copyright 2018 The Kubernetes Authors.
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+    http://www.apache.org/licenses/LICENSE-2.0
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package merge
+
+import "sigs.k8s.io/structured-merge-diff/fieldpath"
+
+// APIVersion describes the version of an object or of a fieldset.
+type APIVersion string
+
+// VersionedSet associates a version to a set.
+type VersionedSet struct {
+	*fieldpath.Set
+	APIVersion APIVersion
+}
+
+// Owners is a map from owner to VersionedSet (what they own in
+// what version).
+type Owners map[string]*VersionedSet
+
+// Difference returns a symmetric difference between two Owners. If a
+// given user's entry has version X in lhs and version Y in rhs, then
+// the return value for that user will be from rhs. If the difference for
+// a user is an empty set, that user will not be inserted in the map.
+func (lhs Owners) Difference(rhs Owners) Owners {
+	diff := Owners{}
+
+	for owner, left := range lhs {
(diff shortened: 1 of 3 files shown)


merge/owners.go
```diff
@@ -0,0 +1,72 @@
...
+	*fieldpath.Set
+	APIVersion APIVersion
+}
+
+// Owners is a map from owner to VersionedSet (what they own in
+// what version).
+type Owners map[string]*VersionedSet
+
+// Difference returns a symmetric difference between two Owners.  If the
+// version is different for a user, then the entire rhs set is returned
```
lavalamp: rhs here and parameter name should maybe be consistent?

merge/owners.go
```diff
@@ -0,0 +1,72 @@
...
+type VersionedSet struct {
+	*fieldpath.Set
+	APIVersion APIVersion
+}
+
+// Owners is a map from owner to VersionedSet (what they own in
+// what version).
+type Owners map[string]*VersionedSet
+
+// Difference returns a symmetric difference between two Owners.  If the
```
lavalamp: If a given user's entry has version X in lhs and version Y in rhs, then the return value for that user will be from rhs.

Clearer?

merge/owners.go
```diff
@@ -0,0 +1,72 @@
...
+// version is different for a user, then the entire rhs set is returned
+// with rhs version. Empty sets are not inserted at all, so a zero-length
+// check will verify that the difference is empty
```
lavalamp: I don't understand this last sentence. Does it mean "users which own no fields will be omitted from the return value"?

merge/owners_test.go
```diff
@@ -0,0 +1,139 @@
...
+			},
+			rhs: merge.Owners{
+				"one": &merge.VersionedSet{
+					Set:        _NS(_P("numeric"), _P("string"), _P("bool")),
+					APIVersion: "v2",
+				},
+			},
+			out: merge.Owners{
+				"one": &merge.VersionedSet{
+					Set:        _NS(_P("numeric"), _P("string"), _P("bool")),
```
lavalamp: Strings that are the names of types make the indices harder to understand--"foo", "bar", etc would be easier?

Also a test name or comment stating what the test case is testing might be helpful--I missed that this was testing the versioning behavior at first.

merge/owners_test.go
```diff
@@ -0,0 +1,139 @@
...
+			},
+			rhs: merge.Owners{
+				"one": &merge.VersionedSet{
+					Set:        _NS(_P("string"), _P("bool")),
+					APIVersion: "v1",
+				},
+			},
+			out: merge.Owners{
+				"one": &merge.VersionedSet{
+					Set:        _NS(_P("numeric"), _P("bool")),
```
lavalamp: I think I expected both to appear here?

merge/owners.go
```diff
@@ -0,0 +1,72 @@
...
+func (o Owners) Difference(other Owners) Owners {
+	diff := Owners{}
+
+	for owner, left := range o {
+		if right, ok := other[owner]; !ok {
+			if !left.Empty() {
+				diff[owner] = left
+			}
```
lavalamp: A `continue` here means you don't need the `else` which keeps things less indented and more readable. (also on the nested if/else)

merge/owners_test.go
lavalamp: nvm this is right.

PR #36, opened 2018-10-22
Add tests for leaf fields

apelisse:
Some tests for applying leaf fields, and updates from controllers.
Also some fixes that became obvious with tests :-)

### DIFF
--- a/merge/leaf_test.go
+++ b/merge/leaf_test.go
@@ -1,0 +1,285 @@
+/*
+Copyright 2018 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package merge_test
+
+import (
+	"reflect"
+	"testing"
+
+	"sigs.k8s.io/structured-merge-diff/merge"
+	"sigs.k8s.io/structured-merge-diff/typed"
+)
+
+var leafFieldsParser = func() *typed.Parser {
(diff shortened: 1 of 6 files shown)


merge/update_test.go
```diff
@@ -95,6 +95,14 @@ func (s *State) CompareLive(obj typed.YAMLObject) (*typed.Comparison, error) {
 	return s.Live.Compare(tv)
 }
 
+// DummyConverter doesn't convert, it just returns the same exact object no matter what.
+type DummyConverter struct{}
```
lavalamp: Does it need to be public?

typed/typed.go
```diff
@@ -110,6 +111,18 @@ func (c *Comparison) IsSame() bool {
 	return c.Removed.Empty() && c.Modified.Empty() && c.Added.Empty()
 }
 
+// String returns a human readable version of the comparison.
+func (c *Comparison) String() string {
+	return fmt.Sprintf(`Merged:
```
lavalamp: s/Merged/Difference/
?

lavalamp: er, how about "Merged Object:"?

merge/leaf_test.go
```diff
@@ -0,0 +1,285 @@
...
+
+import (
+	"reflect"
+	"testing"
+
+	"sigs.k8s.io/structured-merge-diff/merge"
+	"sigs.k8s.io/structured-merge-diff/typed"
+)
+
+var LeafFieldsParser = func() *typed.Parser {
```
lavalamp: Does it need to be public?

typed/typed.go
```diff
@@ -110,6 +111,18 @@ func (c *Comparison) IsSame() bool {
...
+%v
+- Modified:
```
lavalamp: Add " Fields" to each of these three?

typed/typed.go
```diff
@@ -110,6 +111,18 @@ func (c *Comparison) IsSame() bool {
 	return c.Removed.Empty() && c.Modified.Empty() && c.Added.Empty()
 }
 
+// String returns a human readable version of the comparison.
+func (c *Comparison) String() string {
```
lavalamp: Maybe call this function from the command line tool if you're already changing things?

lavalamp:
just nits :)

PR #37, opened 2018-10-23
Write tests for "sets"

apelisse:
Some of the tests are currently broken (because the code is broken) and
are marked as skipped.

### DIFF
--- a/merge/set_test.go
+++ b/merge/set_test.go
@@ -1,0 +1,432 @@
+/*
+Copyright 2018 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


merge/set_test.go
```diff
@@ -0,0 +1,431 @@
...
+	if err != nil {
+		t.Fatalf("Wanted err = %v, got %v", nil, err)
+	}
+
+	config = typed.YAMLObject(`
+list:
+- a
+- b
+- c`)
+	err = state.Apply(config, "default", false)
```
lavalamp: If user shouldn't own b, shouldn't they get a conflict here?

lavalamp:
/lgtm

merge first, fix later.

lavalamp:
argh this will probably make #42 need a rebase :(

PR #38, opened 2018-10-23
update godoc for schema

lavalamp:
Since the schema has drifted slightly from the document, try to document it (in godoc) very clearly.

### DIFF
--- a/schema/doc.go
+++ b/schema/doc.go
@@ -14,5 +14,15 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-// Package schema defines a data model for structured merges and diffs.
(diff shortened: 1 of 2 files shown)


lavalamp:
"Iff" is a common abbreviation of if-and-only-if, at least in papers talking about boolean algebra or philosophy :) Also we can probably get you in the org so you can lgtm for real, I forget the process, you might already meet the requirements.

@apelisse would you like to lgtm for real?

kwiesmueller:
Well, yeah...
My fault, wasn't expecting it in a comment so i thought about a typo, but makes sense.
Then :+1: 

About the reqs, I think one is to have supporters from 2 companies which I am not sure I have.

PR #39, opened 2018-10-23
Slight improvement (?) to Parser

lavalamp:


apelisse:
Yeah that makes sense to me. We'll probably want to go further down that road when we start testing cross-version operations.

apelisse:
Soon will come time to create tests for the parser :-)

PR #40, opened 2018-10-29
Add missing documentation files

lavalamp:
I didn't add one for the merge package because it's a bit unfocused at the moment.

### DIFF
--- a/value/doc.go
+++ b/value/doc.go
@@ -1,0 +1,21 @@
+/*
+Copyright 2018 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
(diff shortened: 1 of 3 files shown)


value/doc.go
```diff
@@ -0,0 +1,20 @@
...
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// Package value defines types for an in-memory representation of yaml or json
+// objects which conform to a schema (as defined by the sibling schema
```
apelisse: > representation of yaml or json objects which conform to a schema

There is no mention of a schema or verification for the schema in the package, or am I missing it?

lavalamp: clarified

apelisse:
Thanks!
/lgtm

PR #41, opened 2018-10-29
Fix "HumanReadable" function name.

lavalamp:


fieldpath/element.go
```diff
@@ -55,14 +55,14 @@ func (e PathElement) String() string {
 	case len(e.Key) > 0:
 		strs := make([]string, len(e.Key))
 		for i, k := range e.Key {
-			strs[i] = fmt.Sprintf("%v=%v", k.Name, k.Value.HumanReadable())
+			strs[i] = fmt.Sprintf("%v=%v", k.Name, k.Value.String())
```
apelisse: Isn't `String()` automatically called when you use `%v`? If so you can drop it everywhere

lavalamp: Yes, hazards of using a rename tool :) will fix

lavalamp:
omitted String() calls.

PR #45, opened 2018-11-02
Make tests more readable

lavalamp:
Also require tabs instead of spaces for object yaml.


apelisse:
I don't know if that makes it easier to read or format, but I don't have a strong opinion.
/lgtm

PR #49, opened 2018-11-05
Export iterator for SetNodeMap

jennybuckley:
Create iterator over the keys of SetNodeMap, mirroring https://github.com/kubernetes-sigs/structured-merge-diff/blob/master/fieldpath/element.go#L179-L184 for the elements of PathElementSet

### DIFF
--- a/fieldpath/set.go
+++ b/fieldpath/set.go
@@ -290,6 +290,13 @@ func (s *SetNodeMap) Difference(s2 *Set) *SetNodeMap {
 	return out
 }
 
+// Iterate calls f for each PathElement in the set.
+func (s *SetNodeMap) Iterate(f func(PathElement)) {
+	for _, n := range s.members {
+		f(n.pathElement)
+	}
+}
+
 func (s *SetNodeMap) iteratePrefix(prefix Path, f func(Path)) {
 	for _, n := range s.members {
 		pe := n.pathElement
--- a/fieldpath/set_test.go
+++ b/fieldpath/set_test.go
@@ -334,3 +334,29 @@ func TestSetIntersectionDifference(t *testing.T) {
 		}
 	})
 }
+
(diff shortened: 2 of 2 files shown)


apelisse:
/lgtm
This seems correct.

/hold
is that the right thing to do @lavalamp ?

jennybuckley:
If this works then https://github.com/jennybuckley/kubernetes/commit/5b0bc2f1cc1e7391cd532d3e3e6f6bfcc0c04c66 will be to add the conversion

lavalamp:
It's either this or construct & return a list, I think I slightly prefer a list but not to the extent that I'd ask for a change.

Either way a test would be nice :)

jennybuckley:
Okay I added a test for this iterator and also the existing one

fieldpath/element_test.go
```diff
@@ -29,3 +29,20 @@ func TestPathElementSet(t *testing.T) {
...
+func TestPathElementSetIterate(t *testing.T) {
+	set := &PathElementSet{}
+	toAdd := 5
+	for i := 0; i < toAdd; i++ {
+		p := i
+		pe := PathElement{Index: &p}
+		set.Insert(pe)
+	}
+	count := 0
+	set.Iterate(func(_ PathElement) {
```
lavalamp: (It doesn't hurt to add this but I think this was effectively tested by the test for `String()`?)

fieldpath/element_test.go
```diff
@@ -29,3 +29,20 @@ func TestPathElementSet(t *testing.T) {
...
+		count++
```
lavalamp: optional nit: I'd verify that you get the right items back, too, personally

lavalamp:
/lgtm

meant to do that earlier sorry

PR #51, opened 2018-11-27
Add FromJSON and ToJSON methods for Value

apelisse:


value/unstructured.go
```diff
@@ -53,6 +54,21 @@ func FromYAML(input []byte) (Value, error) {
 	return v, nil
 }
 
+// FromJSON is a helper function for reading a JSON document
+func FromJSON(input []byte) (Value, error) {
+	var decoded interface{}
+
+	if err := yaml.Unmarshal(input, &decoded); err != nil {
```
jennybuckley: I understand that json is a subset of yaml, but why is it using the yaml library here in the FromJSON function?

jennybuckley:
/lgtm
if using yaml was intentional

PR #54, opened 2019-01-11
skip converting to same version

jennybuckley:
When converting to the same version we already have, don't attempt to go through the hub version


apelisse:
I like the way it's done.
/lgtm

PR #55, opened 2019-01-15
Remove empty sets from managers

jennybuckley:
The three codepaths added are tested by:
- ```TestUpdateLeaf/update_remove_empty_set``` tests that when all fields are removed from existing owner (either by an Update or forced Apply), their entry is removed from managers
- ```TestUpdateSet/apply_update_apply_reorder``` is an existing test, that now also tests that when updating an object, if the caller would own nothing after the own update, their entry is removed from managers
- ```TestUpdateLeaf/apply_remove_empty_set``` tests that when applying to an object, if the caller provides an empty PSO, their entry is removed from managers

/assign @apelisse 
/assign @kwiesmueller

### DIFF
--- a/internal/fixture/state.go
+++ b/internal/fixture/state.go
@@ -259,5 +259,12 @@ func (tc TestCase) Test(parser *typed.ParseableType) error {
 		}
 	}
 
+	// Fail if any empty sets are present in the managers
+	for manager, set := range state.Managers {
+		if set.Empty() {
+			return fmt.Errorf("expected Managers to have no empty sets, but found one managed by %v", manager)
+		}
+	}
+
 	return nil
 }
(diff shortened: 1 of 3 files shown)


internal/fixture/state.go
```diff
@@ -259,5 +259,12 @@ func (tc TestCase) Test(parser *typed.ParseableType) error {
 		}
 	}
 
+	// Fail if any empty sets are present in the managers
+	for manager, set := range state.Managers {
+		if set.Empty() {
+			return fmt.Errorf("expexted Managers to have no empty sets, but found one managed by %v", manager)
```
apelisse: `expexted` :-)

apelisse:
Looks good otherwise

PR #56, opened 2019-01-17
generate schema from value function (for crds)

jennybuckley:
@apelisse @kwiesmueller

### DIFF
--- a/schema/fromvalue_test.go
+++ b/schema/fromvalue_test.go
@@ -1,0 +1,100 @@
+/*
+Copyright 2019 The Kubernetes Authors.
+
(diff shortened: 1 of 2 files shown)


schema/fromvalue_test.go
```diff
@@ -0,0 +1,88 @@
...
+	"sigs.k8s.io/structured-merge-diff/value"
+)
+
+func TestTypeRefFromValue(t *testing.T) {
+
+	table := []struct {
+		objYAML string
+		typeRef string
+	}{
+		{`a: a`, `
```
apelisse: Could you split these on multiple lines? It's very hard to parse.

PR #58, opened 2019-01-17
Deduced types

apelisse:


apelisse:
I'm writing tests now.

typed/parser.go
```diff
@@ -105,10 +107,41 @@ func (p *ParseableType) FromYAML(object YAMLObject) (TypedValue, error) {
...
 	v, err := value.FromUnstructured(in)
 	if err != nil {
 		return TypedValue{}, err
 	}
 	return AsTyped(v, &p.parser.Schema, p.typename)
 }
+
+// DeducedParseableType is a ParseableType that deduces the type from
+// the content of the object.
+type DeducedParseableType struct{}
```
jennybuckley: Why is DeducedParseableType exported, but parsableType isn't?

typed/parser.go
```diff
@@ -68,34 +68,36 @@ func (p *Parser) TypeNames() (names []string) {
...
 
 // ParseableType allows for easy production of typed objects.
-type ParseableType struct {
+type ParseableType interface {
+	IsValid() bool
+	FromYAML(YAMLObject) (TypedValue, error)
+	FromUnstructured(interface{}) (TypedValue, error)
+}
+
+type parseableType struct {
```
jennybuckley: Maybe this could be named, schemaParseableType, explicitParseableType or something else (neither of those sound too good to me), to indicate that this one comes from a schema?

typed/parser.go
jennybuckley: Is it because this type is just a struct{} alias?

typed/parser.go
apelisse: Yeah, it's private though so it's less critical?

typed/parser.go
apelisse: Because the other one needs to be created in a "fancy way", so it's build through a method that returns the interface?
I could do
```
func NewDeducedParseableType() ParseableType {
    return deducedParseableType{}
}
```
but I don't think it's very useful

apelisse: in fewer words, yes =)

PR #59, opened 2019-01-22
Merge multiple types

apelisse:
Fixes everything for merging CRDs (schema-free types).

### DIFF
--- a/typed/deduced.go
+++ b/typed/deduced.go
@@ -1,0 +1,172 @@
+/*
+Copyright 2018 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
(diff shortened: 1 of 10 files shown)


apelisse:
cc @kwiesmueller

typed/deduced.go
```diff
@@ -0,0 +1,172 @@
...
+		// From leaf to map, add leaf fields of map.
+		fieldsetDeduced(rhs, path, set)
+	} else if lhs.MapValue != nil  && rhs.MapValue == nil {
+		// Went from map to field, add field.
+		set.Insert(path)
+	} else {
+		// Both are maps.
+		for i := range rhs.MapValue.Items {
+			child := rhs.MapValue.Items[i]
+			np := path.Copy()
```
jennybuckley: I don't think all the copies are necessary, since we could just add the last part of the path and then remove it after, but it would probably be an early optimization.

PR #63, opened 2019-02-11
Drop managed fields for obsolete versions

apelisse:
This is going to be quite important for CRDs (since versions are much more volatile than for in-tree types). And also is going to be useful anyway because all versions eventually get deprecated.

/assign @kwiesmueller @lavalamp @jennybuckley

### DIFF
--- a/merge/update.go
+++ b/merge/update.go
@@ -24,6 +24,7 @@ import (
 // needs to be able to convert objects from one version to another.
 type Converter interface {
 	Convert(object typed.TypedValue, version fieldpath.APIVersion) (typed.TypedValue, error)
+	IsMissingVersionError(error) bool
 }
 
 // Updater is the object used to compute updated FieldSets and also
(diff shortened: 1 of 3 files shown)


apelisse:
Also, note that this is going to break when we integrate this in kubernetes because of the interface change.

merge/update.go
```diff
@@ -57,10 +58,18 @@ func (s *Updater) update(oldObject, newObject typed.TypedValue, version fieldpat
...
 			if err != nil {
+				if s.Converter.IsMissingVersionError(err) {
+					delete(managers, manager)
+					continue
+				}
 				return nil, fmt.Errorf("failed to convert old object: %v", err)
 			}
 			versioned.newObject, err = s.Converter.Convert(newObject, managerSet.APIVersion)
 			if err != nil {
+				if s.Converter.IsMissingVersionError(err) {
```
apelisse: I don't know how useful this is. Most of the time, if we managed to convert the oldObject, we shouldn't fail to convert the new object right after because the same destination error doesn't exist. BUT, with CRDs, I wonder if there could be a race condition.

jennybuckley: I think we need something like this here also https://github.com/kubernetes-sigs/structured-merge-diff/blob/eeef977c8ccbbb11c78b84595b547d47822dc663/merge/update.go#L164-L171

apelisse:
Good point!

apelisse:
Updated, PTAL @jennybuckley

PR #65, opened 2019-02-13
Fix bug with removing non leaf items

jennybuckley:
Since ToFieldSet creates a set which only includes leaf nodes, diffing the last applied fieldset and the fieldset generated from the merged object isn't enough to determine which list items to remove, if those items are not scalars (which is why the test we have didn't catch it).

### DIFF
--- a/merge/key_test.go
+++ b/merge/key_test.go
@@ -1,0 +1,105 @@
+/*
+Copyright 2019 The Kubernetes Authors.
+ 
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+ 
+    http://www.apache.org/licenses/LICENSE-2.0
+ 
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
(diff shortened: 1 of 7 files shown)


merge/key_test.go
```diff
@@ -0,0 +1,96 @@
...
+
+var b value.String = "b"
+ 
+func TestUpdateAssociativeLists(t *testing.T) {
+	tests := map[string]TestCase{
+		"removing_obsolete_applied_structs": {
+			Ops: []Operation{
+				Apply{
+					Manager:    "default",
+					Object:     `{"list":[{"name":"a", "value": 1}]}`,
```
apelisse: Let's make it into yaml, as soon as yaml works

jennybuckley: It works!

merge/key_test.go
```diff
@@ -0,0 +1,105 @@
...
+func TestUpdateAssociativeLists(t *testing.T) {
+	tests := map[string]TestCase{
+		"removing_obsolete_applied_structs": {
+			Ops: []Operation{
+				Apply{
+					Manager: "default",
+					Object: `
+						list:
+						- name: a
+							value: 1
```
apelisse: NONSENSE

apelisse:
I think it works and that's great, but I still think we should talk about it and what we want for multiple appliers.

jennybuckley:
I added some more tests to this, to make sure nested types work with removal

PR #66, opened 2019-02-14
FixTabsOrDie doesn't work well with YAML structs

apelisse:
Especially when the struct is in a list because:
```yaml
	- a: 1
	  b: 2
```
panics because of mixed spaces and tabs. A working alternative is to
indent with tabs, but I don't think it's very pleasant:
```yaml
	- a: 1
		b: 2
```

Update the code so that prefixed tabs from the first line are removed
from all consecutive lines, and keep it that simple.

/assign @lavalamp @jennybuckley @kwiesmueller

### DIFF
--- a/internal/fixture/state.go
+++ b/internal/fixture/state.go
@@ -35,52 +35,37 @@ type State struct {
 	Updater  *merge.Updater
 }
 
-// FixTabsOrDie counts the number of tab characters preceding the first line in
-// the given yaml object. It removes that many tabs from every line. It then
-// converts remaining tabs to spaces (two spaces per tab). It panics (it's a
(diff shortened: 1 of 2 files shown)


lavalamp:
I don't care that much whether we use tabs or spaces, but doesn't mixing them feel pretty bad? Specifically it means you could use tabs on one line and spaces on the next, and even though that would (might!) look visually very different, it'd mean the same thing? We don't have gofmt helping us out here.

(I don't care enough to write more than this about it though :) )

jennybuckley:
I think this has value in making the tests more readable. I don't think it would allow you to use tabs on one line and spaces on the next, it just trims the same number of tabs off the beginning of each line, and yaml will expect all spaces after that, right?

PR #67, opened 2019-02-14
Fix bogus conflicts comparison and reporting

apelisse:
A few of things that are broken with conflicts:
- We're comparing strings with DeepEqual, which does care about the
order (and we're more considering it like a set, so it's sort-of
random).
- When we print the error, we don't distinguish between errors and
conflict errors, leading to a confusing message.
- When there is a difference we don't report specifically what it is, so
you have to match the sets yourself, which is unhelpful.

Address all of these by having a comparison function that helps us
detect what has changed, so that we can compare and report properly.

/assign @jennybuckley @kwiesmueller

### DIFF
--- a/internal/fixture/state.go
+++ b/internal/fixture/state.go
@@ -152,6 +152,25 @@ type Operation interface {
 	run(*State) error
 }
 
+func hasConflict(conflicts merge.Conflicts, conflict merge.Conflict) bool {
+	for i := range conflicts {
+		if reflect.DeepEqual(conflict, conflicts[i]) {
+			return true
+		}
+	}
+	return false
+}
+
+func addedConflicts(one, other merge.Conflicts) merge.Conflicts {
+	added := merge.Conflicts{}
+	for _, conflict := range other {
+		if !hasConflict(one, conflict) {
+			added = append(added, conflict)
+		}
+	}
+	return added
+}
+
 // Apply is a type of operation. It is a non-forced apply run by a
 // manager with a given object. Since non-forced apply operation can
 // conflict, the user can specify the expected conflicts. If conflicts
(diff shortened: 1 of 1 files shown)


apelisse:
Current tests are flaky because of that (or lack of that).

internal/fixture/state.go
```diff
@@ -182,8 +201,22 @@ var _ Operation = &Apply{}
 
 func (a Apply) run(state *State) error {
 	err := state.Apply(a.Object, a.APIVersion, a.Manager, false)
-	if (err != nil || a.Conflicts != nil) && !reflect.DeepEqual(err, a.Conflicts) {
-		return fmt.Errorf("expected conflicts: %v, got %v", a.Conflicts, err)
+	if err != nil && err.(merge.Conflicts) == nil {
```
kwiesmueller: Did you really want to check for the cast to be nil?
[This will panic when the error is not a conflict](https://golang.org/ref/spec#Type_assertions) instead of returning err.

Instead the ok return should be used: 
```suggestion
	if _, ok := err.(merge.Conflicts); err != nil && !ok {
```

This would as far as I know cast on every pass which might not be wanted, so probably better wrap this?
```suggestion
	if err != nil {
	  if _, ok := err.(merge.Conflicts); !ok {
```

internal/fixture/state.go
```diff
@@ -182,8 +201,22 @@ var _ Operation = &Apply{}
...
+		return err
+	}
+	if a.Conflicts != nil {
+		conflicts := merge.Conflicts{}
+		if err != nil {
+			conflicts = err.(merge.Conflicts)
+		}
+		if len(addedConflicts(a.Conflicts, conflicts)) != 0 || len(addedConflicts(a.Conflicts, conflicts)) != 0 {
```
kwiesmueller: x || x
both sides are the same

maybe:
```suggestion
		if len(addedConflicts(a.Conflicts, conflicts)) != 0 || len(addedConflicts(conflicts, a.Conflicts)) != 0 {
```

apelisse:
Fixed both comments, clearly what I meant ;-)

Thanks @kwiesmueller

PR #69, opened 2019-02-15
Add broken tests for multiple appliers

jennybuckley:
only have one right now, this is to track what we need to fix before multiple appliers works

### DIFF
--- a/merge/multiple_appliers_test.go
+++ b/merge/multiple_appliers_test.go
@@ -1,0 +1,89 @@
+/*
+Copyright 2019 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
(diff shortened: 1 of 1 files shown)


jennybuckley:
@apelisse

merge/multiple_appliers_test.go
```diff
@@ -0,0 +1,233 @@
...
+			},
+			Object: `
+				list:
+				- name: a
+				- name: c
+			`,
+			Managed: fieldpath.ManagedFields{
+				"apply-one": &fieldpath.VersionedSet{
+					Set: _NS(
+						_P("list", _KBF("name", _SV("a")), "name"),
```
apelisse: And so we're considering also having `_P("list", _KBF("name", _SV("a")))` here, right?

jennybuckley: Yes. When and if we make that change, we will have to change all of our tests.

apelisse:
Cool, we probably need the same set of tests for "sets" rather than associative lists.

apelisse:
and maps? I guess it also applies to maps.

PR #70, opened 2019-02-22
 Implement list item and map key removal for multiple appliers

jennybuckley:
The goal of this PR is to define and implement the behavior of apply when it comes to removing list and map items. The strategy implemented here is most simply defined as:
```
Remove an item on apply, iff someone applied with it last time, without it this time,
and no other applier claims to manage it.
```

This solution works without converting any PSOs or fieldsets, and with linear number of conversions wrt the number of versions (so the number of conversions is only getting increased by a constant factor). The downside is that it requires exposing the concept of whether or not a manager is an updater or an applier.

I have tried to include tests for every possible corner case, like multiple appliers and updaters all of different versions, and fields which are dangling without any owner, but shouldn't be removed.

cc @apelisse @kwiesmueller @lavalamp

### DIFF
--- a/fieldpath/managers.go
+++ b/fieldpath/managers.go
@@ -20,6 +20,7 @@ type APIVersion string
 type VersionedSet struct {
 	*Set
 	APIVersion APIVersion
+	Applied    bool
 }
 
 // ManagedFields is a map from manager to VersionedSet (what they own in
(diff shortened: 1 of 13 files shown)


jennybuckley:
Update: Talked offline with @apelisse about this and there is a way to avoid the number of conversions by changing the rule to
```
Remove an item on apply, iff someone applied with it last time, without it this time,
and no other applier claims to manage it.
```
I'll update this PR soon with that change and more tests

PR #71, opened 2019-03-18
Make deduced type use type lookup instead of duplicating logic

jennybuckley:
cc @apelisse @kwiesmueller @lavalamp

This should allow us to re-use the deduced type behavior for raw extensions, instead of treating them as atomic.

I also changed the way deduced types handle map values in this PR, from always treating them as struct fields to treating them as map items, so fields can be removed from custom resources.


schema/elements.go
```diff
@@ -180,15 +184,31 @@ type Map struct {
...
 	// leave this unset to get the default behavior.
 	ElementRelationship ElementRelationship `yaml:"elementRelationship,omitempty"`
+
+	// lookupFunc should only be set if elementRelationship is `lookup`.
+	lookupFunc LookupFunc
+}
+
+type LookupFunc func(*value.Value) TypeRef
+
+func NewUntypedWithLookup(lookupFunc LookupFunc) *Untyped {
```
apelisse: Beside `deduceType`, what other `lookupFunc` are you expecting?

schema/elements.go
```diff
@@ -77,6 +79,8 @@ const (
 	// Separable means the items of the container type have no particular
 	// relationship (default behavior for maps and structs).
 	Separable = ElementRelationship("separable")
+	// Lookup only applies to untyped (see the documentation there).
```
apelisse: This semantic is a little sloppy, you can't look at the schema and now what it's going to do

schema/elements.go
```diff
@@ -180,15 +184,31 @@ type Map struct {
...
 	//   separate actors to set the elements.
 	// TODO: support "guess" (guesses at associative list keys)
-	// TODO: support "lookup" (calls a lookup function to figure out the
-	//       schema based on the data)
 	// The default behavior for untyped data is `atomic`; it's permitted to
 	// leave this unset to get the default behavior.
 	ElementRelationship ElementRelationship `yaml:"elementRelationship,omitempty"`
+
+	// lookupFunc should only be set if elementRelationship is `lookup`.
+	lookupFunc LookupFunc
```
apelisse: This is not serialized and/or persisted, which means that the behavior is sort of unpredictable?

schema/elements.go
apelisse: Since `deduceType`  is the only lookup method you have, could this be named `Deduce` instead?

typed/deduced_test.go
```diff
@@ -218,7 +218,7 @@ func TestMergeDeduced(t *testing.T) {
 func TestToSetDeduced(t *testing.T) {
 	tests := []objSetPair{
 		{`{"key":"foo","value":1}`, _NS(_P("key"), _P("value"))},
-		{`{"key":"foo","value":{"a": "b"}}`, _NS(_P("key"), _P("value", "a"))},
+		{`{"key":"foo","value":{"a": "b"}}`, _NS(_P("key"), _P("value"), _P("value", "a"))},
```
apelisse: Changes in this file clearly make sense ...

schema/elements.go
jennybuckley: None but I wanted to put the logic for deciding which schema to use in typed/deduced.go instead of here

typed/deduced_test.go
jennybuckley: Yes the changes here were all what I expected to see when changing the struct fields to map values by default

schema/elements.go
```diff
@@ -180,17 +184,45 @@ type Map struct {
 type Untyped struct {
 	// ElementRelationship states the relationship between the items, if
 	// container-typed data happens to be present here.
+	// * `deduced` implies that the behavior is based on the type of data.
```
apelisse: Can we get a slightly better explanation of what that means? Sounds like "This does the right thing in the right scenario" :-))

typed/deduced.go
```diff
@@ -17,162 +17,28 @@ limitations under the License.
...
-			modified(v.Value, child.Value, np, set)
-		}
-	}
-}
-
-// RemoveItems does nothing because all lists in a deducedTypedValue are considered atomic,
-// and there are no maps because it is indistinguishable from a struct.
-func (dv deducedTypedValue) RemoveItems(_ *fieldpath.Set) TypedValue {
-	return dv
+	return AsTypedUnvalidated(v, &deducedSchema, deducedName)
```
apelisse: Since we're getting rid of all of this, I'm curious if we should make TypedValue back into a struct rather than an interface.

schema/elements.go
jennybuckley: I put it below https://github.com/kubernetes-sigs/structured-merge-diff/pull/71/files/914b72e3c5ebdf4f5a63c4fef4235a87f1b74fa6#diff-60731b77d91302b54bc9fb3b89f6521aR197

Should I just move that here?

typed/deduced.go
jennybuckley: Will that still be compatible with what we did in k8s? I think yes but I'm not sure

apelisse: I know that Daniel wants to start keeping the thing compatible. It's not done yet, let's use that chance ;-)

lavalamp:
I didn't look carefully, but what I did see looks great :)

I also like that there's more lines removed than added.

merge/update.go
```diff
@@ -146,7 +146,7 @@ func (s *Updater) Update(liveObject, newObject typed.TypedValue, version fieldpa
 // Apply should be called when Apply is run, given the current object as
 // well as the configuration that is applied. This will merge the object
 // and return it.
-func (s *Updater) Apply(liveObject, configObject typed.TypedValue, version fieldpath.APIVersion, managers fieldpath.ManagedFields, manager string, force bool) (typed.TypedValue, fieldpath.ManagedFields, error) {
+func (s *Updater) Apply(liveObject, configObject *typed.TypedValue, version fieldpath.APIVersion, managers fieldpath.ManagedFields, manager string, force bool) (*typed.TypedValue, fieldpath.ManagedFields, error) {
```
apelisse: Should we check for nil? Since this is at the API layer

merge/update.go
```diff
@@ -119,7 +119,7 @@ func (s *Updater) update(oldObject, newObject typed.TypedValue, version fieldpat
 // that you intend to persist (after applying the patch if this is for a
 // PATCH call), and liveObject must be the original object (empty if
 // this is a CREATE call).
-func (s *Updater) Update(liveObject, newObject typed.TypedValue, version fieldpath.APIVersion, managers fieldpath.ManagedFields, manager string) (fieldpath.ManagedFields, error) {
+func (s *Updater) Update(liveObject, newObject *typed.TypedValue, version fieldpath.APIVersion, managers fieldpath.ManagedFields, manager string) (fieldpath.ManagedFields, error) {
```
apelisse: Same as below, since this is API layer, should we check for nil?

typed/typed.go
```diff
@@ -76,36 +44,38 @@ func AsTyped(v value.Value, s *schema.Schema, typeRef schema.TypeRef) (TypedValu
...
 	value   value.Value
 	typeRef schema.TypeRef
 	schema  *schema.Schema
 }
 
-var _ TypedValue = typedValue{}
-
-func (tv typedValue) AsValue() *value.Value {
+// AsValue removes the type from the TypedValue and only keeps the value.
+func (tv *TypedValue) AsValue() *value.Value {
```
apelisse: I don't think it matters, but this is now always returning the same value (while it wasn't before, because it was returning a pointer in the copy)?

typed/parser.go
```diff
@@ -69,79 +69,52 @@ func (p *Parser) TypeNames() (names []string) {
...
 	}
 }
 
 // ParseableType allows for easy production of typed objects.
-type ParseableType interface {
-	IsValid() bool
-	FromYAML(YAMLObject) (TypedValue, error)
-	FromUnstructured(interface{}) (TypedValue, error)
+type ParseableType struct {
+	schema.TypeRef
```
apelisse: I'm not a huge fan of the embedding ... (I think this is composition)

merge/update.go
jennybuckley: I don't mind adding that check, but we didn't check it before and the interface could have also been nil

apelisse: Yeah, we can skip that for now

schema/elements_test.go
```diff
@@ -70,10 +70,13 @@ func TestResolve(t *testing.T) {
 		tt := tt
 		t.Run(tt.testName, func(t *testing.T) {
 			t.Parallel()
-			s := Schema{
-				Types: tt.schemaTypeDefs,
+			s := TypeSpecifier{
+				Type: tt.typeRef,
```
apelisse: should `tt` have a typespecifier?

typed/merge.go
```diff
@@ -407,12 +406,12 @@ func (w *mergingWalker) doUntyped(t schema.Untyped) (errs ValidationErrors) {
 		// Merge using the lhs's type on a shallow copy of the merge walker,
 		// to get rid of the merged output 
 		w2 := *w
-		w2.typeRef = schema.DeduceType(w.lhs)
+		w2.typeSpecifier.Type = schema.DeduceType(w.lhs)
```
apelisse: It feels a little "leaky" that we're always doing that. Can we return a `TypeSpecifier` from that method for example?

typed/remove.go
```diff
@@ -124,7 +127,7 @@ func (w *removingWalker) doMap(t schema.Map) ValidationErrors {
 
 func (w *removingWalker) doUntyped(t schema.Untyped) ValidationErrors {
 	if t.ElementRelationship == schema.Deduced {
-		resolveSchema(w.schema, schema.DeduceType(w.value), w)
+		removeItemsWithSchema(w.value, w.toRemove, w.schema, schema.DeduceType(w.value))
```
apelisse: Here it'be nice to have the TypeSpecifier directly

typed/parser.go
jennybuckley: Fixed

PR #72, opened 2019-03-20
Update schema to support oneOf, and normalize objects

apelisse:
This does mostly two things:
- Defines the schema for OneOf: multiple oneof can be at the same level (in the same structure),
- Normalizes an object based on oneofs definition: uses the discriminator if present and has changed, uses the difference of fields if they changed, updates the discriminator if present.

And adds tests. This is still not plugged anyway though, so it's not going to do anything.

### DIFF
--- a/schema/elements.go
+++ b/schema/elements.go
@@ -83,18 +83,15 @@ const (
 
 // Struct represents a type which is composed of a number of different fields.
 // Each field has a name and a type.
-//
-// TODO: in the future, we will add one-of groups (sometimes called unions).
 type Struct struct {
 	// Each struct field appears exactly once in this list. The order in
 	// this list defines the canonical field ordering.
 	Fields []StructField `yaml:"fields,omitempty"`
 
-	// TODO: Implement unions, either this way or by inlining.
-	// Unions are groupings of fields with special rules. They may refer to
+	// Union is a grouping of fields with special rules. It may refer to
 	// one or more fields in the above list. A given field from the above
 	// list may be referenced in exactly 0 or 1 places in the below list.
-	// Unions []Union `yaml:"unions,omitempty"`
+	Union *Union `yaml:"union,omitempty"`
 
 	// ElementRelationship states the relationship between the struct's items.
 	// * `separable` (or unset) implies that each element is 100% independent.
@@ -106,6 +103,45 @@ type Struct struct {
 	// The default behavior for structs is `separable`; it's permitted to
 	// leave this unset to get the default behavior.
 	ElementRelationship ElementRelationship `yaml:"elementRelationship,omitempty"`
+}
+
+// UnionFields are mapping between the fields that are part of the union and
+// their discriminated value. The discriminated value has to be set, and
+// should not conflict with other discriminated value in the list.
+type UnionField struct {
+	// FieldName is the name of the field that is part of the union. This
+	// is the serialized form of the field.
+	FieldName string `yaml:"fieldName"`
+	// DiscriminatedBy is the value of the discriminator to select that
+	// field.
+	DiscriminatedBy string `yaml:"DiscriminatedBy"`
+}
+
+// Union, or oneof, means that only one of multiple fields of a structure can be
+// set at a time. For backward compatibility reasons, and to help "dumb clients"
+// which are not aware of the union (or can't be aware of it because they
+// don't know what fields are part of the union), the code tolerates multiple
+// fields to be set but will try to detect which fields must be cleared (there
+// should never be more than two though):
+// - If there is a discriminator and its value has changed, clear all fields
+// but the one specified by the discriminator
+// - If there is no discriminator, or it hasn't changed, if new has two of the
+// fields set, remove the one that was set in old.
+// - If there is a discriminator, set it to the value we've kept (if it changed)
+type Union struct {
+	// Discriminator, if present, is the name of the field that
+	// discriminates fields in the union. The mapping between the value of
+	// the discriminator and the field is done by using the Fields list
+	// below.
+	Discriminator *string `yaml:"discriminator,omitempty"`
+
+	// This is the list of fields that belong to this union. This fields are
+	// required to not be part of another union, or be the discriminator for
+	// another union. All the fields present in here have to be part of the
+	// parent structure. Discriminator (if oneOf has one), is NOT included
+	// in this list. The value for field is how we map the name of the field
+	// to actual value for discriminator.
+	Fields []UnionField `yaml:"fields,omitempty"`
 }
 
 // StructField pairs a field name with a field type.
--- a/schema/elements.go
+++ b/schema/elements.go
@@ -113,8 +113,9 @@ type UnionField struct {
 	// is the serialized form of the field.
 	FieldName string `yaml:"fieldName"`
 	// DiscriminatedBy is the value of the discriminator to select that
-	// field.
-	DiscriminatedBy string `yaml:"DiscriminatedBy"`
+	// field. If the union doesn't have a discriminator, this field is
+	// ignored.
+	DiscriminatedBy string `yaml:"discriminatedBy"`
 }
 
 // Union, or oneof, means that only one of multiple fields of a structure can be
@@ -135,11 +136,10 @@ type Union struct {
 	// below.
 	Discriminator *string `yaml:"discriminator,omitempty"`
 
-	// This is the list of fields that belong to this union. This fields are
-	// required to not be part of another union, or be the discriminator for
-	// another union. All the fields present in here have to be part of the
-	// parent structure. Discriminator (if oneOf has one), is NOT included
-	// in this list. The value for field is how we map the name of the field
+	// This is the list of fields that belong to this union. All the
+	// fields present in here have to be part of the parent
+	// structure. Discriminator (if oneOf has one), is NOT included in
+	// this list. The value for field is how we map the name of the field
 	// to actual value for discriminator.
 	Fields []UnionField `yaml:"fields,omitempty"`
 }
--- a/typed/union.go
+++ b/typed/union.go
@@ -1,0 +1,215 @@
+/*
+Copyright 2019 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package typed
+
+import (
+	"fmt"
+
+	"sigs.k8s.io/structured-merge-diff/schema"
+	"sigs.k8s.io/structured-merge-diff/value"
+)
+
+func normalizeUnion(w *mergingWalker) error {
+	atom, found := w.schema.Resolve(w.typeRef)
+	if !found {
+		panic(fmt.Sprintf("Unable to resolve schema in normalize union: %v/%v", w.schema, w.typeRef))
+	}
+	// Unions can only be in structures, and the struct must not have been removed
+	if atom.Struct == nil || atom.Struct.Union == nil || w.out == nil {
+		return nil
+	}
+
+	old := &value.Map{}
+	if w.lhs != nil {
+		old = w.lhs.MapValue
+	}
+	return newUnion(atom.Struct.Union).Normalize(old, w.rhs.MapValue, w.out.MapValue)
+}
+
+type discriminated string
+type field string
+
+type discriminatedNames struct {
+	f2d map[field]discriminated
+	d2f map[discriminated]field
+}
+
+func newDiscriminatedName(f2d map[field]discriminated) discriminatedNames {
+	d2f := map[discriminated]field{}
+	for key, value := range f2d {
+		d2f[value] = key
+	}
+	return discriminatedNames{
+		f2d: f2d,
+		d2f: d2f,
+	}
+}
+
+func (dn discriminatedNames) toField(d discriminated) field {
+	if f, ok := dn.d2f[d]; ok {
+		return f
+	}
+	return field(d)
+}
+
+func (dn discriminatedNames) toDiscriminated(f field) discriminated {
+	if d, ok := dn.f2d[f]; ok {
+		return d
+	}
+	return discriminated(f)
+}
+
+type discriminator struct {
+	name string
+}
+
+func (d *discriminator) Set(m *value.Map, v discriminated) {
+	if d == nil {
+		return
+	}
+	m.Set(d.name, value.StringValue(string(v)))
+}
+
+func (d *discriminator) Get(m *value.Map) discriminated {
+	if d == nil || m == nil {
+		return ""
+	}
+	f, ok := m.Get(d.name)
+	if !ok {
+		return ""
+	}
+	if f.Value.StringValue == nil {
+		return ""
+	}
+	return discriminated(*f.Value.StringValue)
+}
+
+type fieldsSet map[field]struct{}
+
+// newFieldsSet returns a map of the fields that are part of the union and are set
+// in the given map.
+func newFieldsSet(m *value.Map, fields []field) fieldsSet {
+	if m == nil {
+		return nil
+	}
+	set := fieldsSet{}
+	for _, f := range fields {
+		if _, ok := m.Get(string(f)); ok {
+			set.Add(f)
+		}
+	}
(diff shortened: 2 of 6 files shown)


typed/union.go
```diff
@@ -0,0 +1,130 @@
...
+	}
+	for _, field := range fields {
+		if _, ok := v.MapValue.Get(field); ok {
+			set[field] = struct{}{}
+		}
+	}
+	return set
+}
+
+// Set the discriminator to a specific value
```
jennybuckley: get

typed/union.go
```diff
@@ -0,0 +1,130 @@
...
+		}
+	})
+)
+
+func normalizeUnion(union schema.Union, old, new, out *value.Value) {
+	df := ""
+	if union.Discriminator != nil {
+		df = uncapitalize(*union.Discriminator)
+		odv := getDiscriminatorValue(old, df)
+		ndv := getDiscriminatorValue(new, df)
```
jennybuckley: Do we also need to uncapitalize the odv and ndv? It seems a bit kubernetes specific.

typed/union_test.go
```diff
@@ -0,0 +1,224 @@
...
+    - name: two
+      type:
+        scalar: numeric
+    - name: three
+      type:
+        scalar: numeric
+    unions:
+      - discriminator: discriminator
+        fields: [a, b, c]
+      - fields: [one, two, three]`)
```
jennybuckley: Test looks good, maybe we should also try some nested structs

typed/union.go
```diff
@@ -0,0 +1,130 @@
...
+
+	// set discriminator value to only item in nfs
+	dv := ""
+	for v := range nfs {
+		dv = v
+	}
+
+	for field := range ofs {
+		if _, ok := nfs[field]; !ok {
+			out.MapValue.Delete(field)
```
jennybuckley: Maybe we could use the same
```
            for _, field := range union.Fields {
				if field != ndv {
					out.MapValue.Delete(field)
				}
			}
```
loop from before (or a function that does that) and not worry about ofs or nfs anymore?

typed/union.go
```diff
@@ -0,0 +1,130 @@
...
+		}
+	}
+
+	if df != "" {
+		out.MapValue.Set(df, value.StringValue(dv))
```
jennybuckley: `value.StringValue(capitalize(dv))`?

schema/elements.go
```diff
@@ -106,6 +103,31 @@ type Struct struct {
...
 }
 
+// Union, or oneof, means that only one of multiple fields of a structure can be
+// set at all times. For backward compatibility reasons, and to help "dumb
+// clients" that are not aware of union (or can't be aware of this because they
+// don't know what fields are part of the union), the code tolerates multiple
+// fields to be set but will try to detect which fields must be cleaned (there
+// should never be more than two though):
+// - If the Discriminator field exist, the value of the field will be used to
+// decide which field to clean and which field to keep
```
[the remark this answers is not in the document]
apelisse: Yes, only if it changed, good!

typed/deduced.go
```diff
@@ -124,6 +124,10 @@ func (dv deducedTypedValue) Compare(rhs TypedValue) (c *Comparison, err error) {
 	return c, nil
 }
 
+func (dv deducedTypedValue) NormalizeUnions(old TypedValue) (TypedValue, error) {
```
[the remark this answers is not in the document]
apelisse: yes, I'm working on what I have. I'll work on something else when we have something else :-)

typed/union.go
```diff
@@ -0,0 +1,130 @@
...
+	}
+
+	ofs := unionFieldsSet(old, union.Fields)
+	nfs := unionFieldsSet(new, union.Fields)
+
+	// Remove fieldset in ofs from nfs
+	for field := range ofs {
+		delete(nfs, field)
+	}
+	// If we have nothing left, then nothing changed,
```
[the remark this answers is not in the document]
apelisse: Good point, I'm not sure I thought about that

typed/union.go
```diff
@@ -0,0 +1,130 @@
...
+	// if we have more than 1, it means that more than 1 field changed
+	if len(nfs) != 1 {
+		return
+	}
+
+	// set discriminator value to only item in nfs
+	dv := ""
+	for v := range nfs {
```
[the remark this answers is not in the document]
apelisse: Yes, if only

jennybuckley: What about this:
`reflect.ValueOf(nfs).MapKeys()[0].Interface().(string)`

apelisse: Yeah, that's beautiful, I wonder why I haven't done that before :rofl:

schema/elements.go
```diff
@@ -81,18 +81,15 @@ const (
...
 	// Each struct field appears exactly once in this list. The order in
 	// this list defines the canonical field ordering.
 	Fields []StructField `yaml:"fields,omitempty"`
 
-	// TODO: Implement unions, either this way or by inlining.
 	// Unions are groupings of fields with special rules. They may refer to
 	// one or more fields in the above list. A given field from the above
 	// list may be referenced in exactly 0 or 1 places in the below list.
-	// Unions []Union `yaml:"unions,omitempty"`
+	Unions []Union `yaml:"unions,omitempty"`
```
lavalamp: One of the reasons this was commented out is that I wasn't sure if I wanted to do it the way you've added it, or if it would be better to use a struct per union and support embedding a list of structs. The other way would make it impossible to try and put the same field into multiple unions. I'm not sure where, but it'd be nice to see a discussion of why this format is chosen. (Maybe you have it and I haven't seen it yet.)

lavalamp: Also consider a different name, since you have to explain "Union". Maybe Exclusivities or OneOfs?

typed/union_test.go
```diff
@@ -0,0 +1,224 @@
...
+			name: "proper union update from not-set, no discriminator",
+			old:  `{}`,
+			new:  `{"two": 1}`,
+			out:  `{"two": 1}`,
+		},
+		{
+			name: "remove union, with discriminator",
+			old:  `{"a": 1}`,
+			new:  `{}`,
+			out:  `{}`,
```
lavalamp: Does the discriminator need an explicit "not set" value?

typed/union_test.go
```diff
@@ -0,0 +1,224 @@
...
+			name: "remove union and discriminator",
+			old:  `{"a": 1, "discriminator": "a"}`,
+			new:  `{}`,
+			out:  `{}`,
+		},
+		{
+			name: "remove union, not discriminator",
+			old:  `{"a": 1, "discriminator": "a"}`,
+			new:  `{"discriminator": "a"}`,
+			out:  `{"discriminator": "a"}`,
```
lavalamp: Should the discriminator have been cleared?

typed/union_test.go
```diff
@@ -0,0 +1,224 @@
...
+		{
+			name: "dumb client update, no discriminator",
+			old:  `{"one": 1}`,
+			new:  `{"one": 2, "two": 1}`,
+			out:  `{"two": 1}`,
+		},
+		{
+			name: "dumb client update, sets discriminator",
+			old:  `{"a": 1}`,
+			new:  `{"a": 2, "b": 1}`,
```
lavalamp: I think this should have triggered an error; a dumb client needn't clear a, but it still shouldn't change a and also add b. Thoughts?

typed/union_test.go
```diff
@@ -0,0 +1,224 @@
...
+		{
+			name: "dumb client doesn't update discriminator",
+			old:  `{"a": 1, "discriminator": "a"}`,
+			new:  `{"a": 2, "b": 1, "discriminator": "a"}`,
+			out:  `{"b": 1, "discriminator": "b"}`,
+		},
+		{
+			name: "client sends multiple with discriminator",
+			old:  `{}`,
+			new:  `{"a": 1, "b": 1, "discriminator": "a"}`,
```
lavalamp: I also think this should cause an error.

typed/union_test.go
```diff
@@ -0,0 +1,224 @@
...
+			name: "multi-discriminator at the same time",
+			old:  `{"one": 1, "a": 1}`,
+			new:  `{"one": 1, "two": 1, "a": 1, "b": 1}`,
+			out:  `{"two": 1, "b": 1, "discriminator": "b"}`,
+		},
+		{
+			name: "change discriminator, nothing else",
+			old:  `{"discriminator": "a"}`,
+			new:  `{"discriminator": "random"}`,
+			out:  `{"discriminator": "random"}`,
```
lavalamp: random discriminator values should cause an error, don't you think?

typed/union_test.go
```diff
@@ -0,0 +1,224 @@
...
+			name: "change discriminator, nothing else, it drops other field",
+			old:  `{"discriminator": "a", "a": 1}`,
+			new:  `{"discriminator": "random", "a": 1}`,
+			out:  `{"discriminator": "random"}`,
+		},
+		{
+			name: "remove discriminator, nothing else",
+			old:  `{"discriminator": "a", "a": 1}`,
+			new:  `{"a": 1}`,
+			out:  `{"a": 1}`,
```
lavalamp: I guess I would have expected a to be cleared also--not 100% sure.

typed/union_test.go
```diff
@@ -0,0 +1,224 @@
...
+		{
+			name: "old object has two of same union set",
+			old:  `{"a": 1, "b": 2}`,
+			new:  `{"a": 2, "b": 1}`,
+			out:  `{"a": 2, "b": 1}`,
+		},
+		{
+			name: "new object has three of same union set",
+			old:  `{"a": 1}`,
+			new:  `{"a": 2, "b": 1, "c": 3}`,
```
lavalamp: This should cause an error, I think.

typed/union_test.go
```diff
@@ -0,0 +1,224 @@
...
+			name: "old object has two of same union, but we add third",
+			old:  `{"discriminator": "a", "a": 1, "b": 1}`,
+			new:  `{"discriminator": "a", "a": 1, "b": 1, "c": 1}`,
+			out:  `{"discriminator": "c", "c": 1}`,
+		},
+		// Following use-cases are not doing anything because they are "broken"
+		{
+			name: "old object has two of same union set",
+			old:  `{"a": 1, "b": 2}`,
+			new:  `{"a": 2, "b": 1}`,
```
lavalamp: I'd like a test case exactly like this except only one of the values changes.

typed/union_test.go
jennybuckley: I think this is similar to the one below, which is to make sure changing a DeploymentStrategyType from RollingUpdate to Recreate clears the RollingUpdate field.
Recreate isn't a name of a field in the Union but it is an allowed value of DeploymentStrategyType. If it's not allowed as a DeploymentStrategyType, I think validation outside smd could handle that kind of thing.

typed/union_test.go
jennybuckley: Also we should test that creating a new object should still set the discriminator correctly

typed/union_test.go
jennybuckley: I think this would cause an error outside smd as well

typed/union_test.go
apelisse: Actually, the use-case in mind here is the following:
- I'm using kustomize
- my base deployment uses an emptyDir
- my overlay uses a persistent disk
- kustomize typically merges this by having both (you have to explicitely set `emptyDir` to `null`)

In that case, you could use the discriminator to clarify (which is much better than setting discriminator to null)

lavalamp: It sounds like kustomize should use this library to do the merges. It'd be pretty nifty if customize produced output which had each overlay as a separate field manager.

typed/union_test.go
apelisse: I really really didn't want to be in the business of comparing deeply nested sub-structures, where fields can have been dropped because of Golang. My goal was very clear: compare the list of fields set, not their content :-/

lavalamp: I'm on the fence for update, but for apply--consider what happens the next time they apply if `{"a":2,"b":1}` is what is in their local file. I think we should attempt to prevent them from having a local file that sets multiple fields.

I guess think update should prevent this, too, except in cases where that is adding an additional validation constraint--for those we need validation latching.

typed/union_test.go
jennybuckley: So this would mean someone could send 
```
deploymentStrategyType: Recreate
rollingUpdate:
  maxUnavailable: 1
```
and it would be accepted and treated as this now?
```
deploymentStrategyType: Recreate
rollingUpdate: {}
```

typed/union_test.go
apelisse: I thought about this, but we can't detect that once we have merged the object, so I was planning on adding another validation step just for that before we merge.

schema/elements.go
apelisse: The way it's going, we're only going to accept one union per structure in the OpenAPI, which would prevent any sort of overlap between fields

schema/elements.go
```diff
@@ -110,6 +107,33 @@ type Struct struct {
...
+// - If there is a discriminator and its value has changed, clear all fields
+// but the one specified by the discriminator
+// - If there is no discriminator, or it hasn't changed, if new has two of the
+// fields set, remove the one that was set in old.
+// - If there is a discriminator, set it to the value we've kept (if it changed)
+type Union struct {
+	// Discriminator, if present, is the name of the field that
+	// discriminates fields in the union. The mapping between the value of
+	// the discriminator and fields works as by uncapitalizing the first
+	// letter (or leave untouched if not capitalized).
```
jennybuckley: this comment is outdated

schema/elements.go
```diff
@@ -110,6 +107,33 @@ type Struct struct {
...
+	Discriminator *string `yaml:"discriminator,omitempty"`
+
+	// This is the list of fields that belong to this union. This fields are
+	// required to not be part of another union, or be the discriminator for
+	// another union. All the fields present in here have to be part of the
+	// parent structure. Discriminator (if oneOf has one), is NOT included
+	// in this list. The value for field is how we map the name of the field
+	// to actual value for discriminator.
+	Fields map[string]string `yaml:"fields,omitempty"`
```
jennybuckley: can a value be mapped to from multiple keys?

jennybuckley: Also maybe this should be a list, since we use lists in all the other schema pieces

apelisse: > Also maybe this should be a list, since we use lists in all the other schema pieces

Good point

> can a value be mapped to from multiple keys?

No, it shouldn't. The part that converts the openapi into this schema is going to verify that this is not possible. If it still happens that we have to mappings going to the same key, it will probably use the last one in the list.

apelisse:
Fixed

typed/union.go
```diff
@@ -0,0 +1,215 @@
...
+
+	if u.d.Get(old) != u.d.Get(new) && u.d.Get(new) != "" {
+		if len(diff) == 1 && u.d.Get(new) != u.dn.toDiscriminated(discriminator) {
+			return fmt.Errorf("discriminator and field changed: %v/%v", discriminator, u.d.Get(new))
+		}
+		u.clear(out, u.dn.toField(u.d.Get(new)))
+		return nil
+	}
+
+	if discriminator != "" {
```
jennybuckley: is it even possible for discriminator to be "" at this point? it is hard to tell

typed/union_test.go
```diff
@@ -0,0 +1,288 @@
...
+    - name: b
+      type:
+        scalar: numeric
+    - name: c
+      type:
+        scalar: numeric
+    union:
+      fields:
+      - fieldName: a
+        discriminatedBy: A
```
jennybuckley: Seems strange to have "discriminatedBy" set in "nondiscriminated", no?

schema/elements.go
```diff
@@ -110,6 +107,45 @@ type Struct struct {
...
+// don't know what fields are part of the union), the code tolerates multiple
+// fields to be set but will try to detect which fields must be cleared (there
+// should never be more than two though):
+// - If there is a discriminator and its value has changed, clear all fields
+// but the one specified by the discriminator
+// - If there is no discriminator, or it hasn't changed, if new has two of the
+// fields set, remove the one that was set in old.
+// - If there is a discriminator, set it to the value we've kept (if it changed)
+type Union struct {
+	// Discriminator, if present, is the name of the field that
```
jennybuckley: I almost think that we should always have a discriminator, or maybe a virtual one when the object doesn't actually store it. That way we could be guaranteed to detect conflicts in the union. I think we talked about this before, either adding a discriminator everywhere, (ex, in "volumeSource" we were going to add "volumeSourceType", or we could infer a name from in this repo if it isn't present)

Adding the type everywhere should be safe as long as we update it for the user automatically

apelisse:
I forgot to cc @sttts, I'll happily update code based on your comments Stefan!

PR #73, opened 2019-03-27
Support atoms with multiple types

jennybuckley:
Get rid of schema.DeduceType and instead allow atoms with multiple types defined and use the type of data to eliminate incompatible types.

Also changes Untyped Deduced to this:
```
types:
- name: __untyped_deduced_
  scalar: untyped
  list:
    elementType:
      namedType: __untyped_atomic_
    elementRelationship: atomic
  map:
    elementType:
      namedType: __untyped_deduced_
    elementRelationship: separable
```

This is necessary if we want to support things like JSONPropsOrArray in the future


typed/merge.go
```diff
@@ -374,8 +392,8 @@ func (w *mergingWalker) visitMapItems(t schema.Map, lhs, rhs *value.Map) (errs V
 
 func (w *mergingWalker) doMap(t schema.Map) (errs ValidationErrors) {
 	var lhs, rhs *value.Map
-	errs = append(errs, w.derefMapOrStruct("lhs: ", "map", w.lhs, &lhs)...)
-	errs = append(errs, w.derefMapOrStruct("rhs: ", "map", w.rhs, &rhs)...)
+	w.derefMapOrStruct("lhs: ", "map", w.lhs, &lhs)
+	w.derefMapOrStruct("rhs: ", "map", w.rhs, &rhs)
```
jennybuckley: We could actually remove these errors from all the cases, but I wanted to keep the diff smaller

apelisse: If we don't think that they should ever happen, can we panic?

jennybuckley: the error case can happen, but what we want it to do in that case is treat the value as nil

typed/merge.go
```diff
@@ -61,12 +63,26 @@ var (
...
-	errs := resolveSchema(w.schema, w.typeRef, w)
+	a, ok := w.schema.Resolve(w.typeRef)
+	if !ok {
+		return w.errorf("schema error: no type found matching: %v", *w.typeRef.NamedType)
+	}
+
+	alhs := deduceAtom(a, w.lhs)
+	arhs := deduceAtom(a, w.rhs)
+	if reflect.DeepEqual(alhs, arhs) {
+		errs =append(errs, handleAtom(arhs, w.typeRef, w)...)
```
apelisse: very important: missing space :-)

apelisse:
Do you want to squash and I lgtm or do you want to move the type to schemaconv?

apelisse:
I think we agree that it was better before that last commit?

apelisse:
We need to add a little bit more testing for this feature, let's follow-up after.
Also we need to fix the gofmt thing, I'll probably force merge.

PR #79, opened 2019-04-20
Default unspecified values of unspecified key fields in associative lists

jennybuckley:
If any key fields in a multi-key associative list is defaulted, currently we require all keys to be specified or we fail. This means, for example, that an applier needs to specify both port and protocol to edit a container port.

This PR adds support for that use case by running a defaulter on the applied config, and only keeping the defaulted values of the missing key fields.

@apelisse @kwiesmueller @lavalamp


typed/multikeys.go
```diff
@@ -0,0 +1,89 @@
...
+package typed
+
+import (
+	"fmt"
+	"reflect"
+
+	"sigs.k8s.io/structured-merge-diff/schema"
+)
+
+func completeKeys(w *mergingWalker) error {
```
jennybuckley: Going to unit test this next

jennybuckley:
I think we decided against this approach but I'll leave this here because we probably still want at least some of the test cases

PR #80, opened 2019-04-26
Oneof lists

apelisse:
Implement support for multiple oneof's in the same structure.

/assign @jennybuckley 
/cc @lavalamp


jennybuckley:
This looks good, do we need anything validating that there is no overlap within the unions? or will that be in a different layer

apelisse:
Yeah, I was thinking about that, I'll add this here.

apelisse:
Actually it should happen in `openapi -> smd`.

PR #81, opened 2019-05-14
Merge map and struct, allowing for unknown struct fields

jennybuckley:
Ehen deserializing a core type we throw away unknown fields because they don't fit into the go struct, but with CRs we don't throw them away because we are always deserializing into Unstructured. Right now we have no concept in structured-merge-diff of additional properties on a struct, and any fields of an object being set, other than the ones in the schema, make it fail validation.

This could be solved by pruning CRDs, which is being worked on, but still will never always be done because it's backwards incompatible (current CRD validation fields might not always be fully specified).

A more complete solution is to merge struct and map in our schema, and allow additional properties in structs (if a defaultElementType field is set). This addresses issues with CRD unknown fields not always being pruned. It also makes some sense to merge these things because map and struct have one representation in value, but two in schema, only one is allowed to be set in an Atom, and they have a lot of duplicated code. By merging these we can actually remove map from the schema altogether, and replace it with a struct with no fields defined and just a defaultElementType (I did this and all our existing tests pass)

### DIFF
--- a/merge/nested_test.go
+++ b/merge/nested_test.go
@@ -27,7 +27,7 @@ import (
 var nestedTypeParser = func() typed.ParseableType {
 	parser, err := typed.NewParser(`types:
 - name: type
-  struct:
+  map:
     fields:
       - name: listOfLists
         type:
@@ -47,7 +47,7 @@ var nestedTypeParser = func() typed.ParseableType {
 - name: listOfLists
   list:
     elementType:
-      struct:
+      map:
         fields:
         - name: name
           type:
@@ -66,7 +66,7 @@ var nestedTypeParser = func() typed.ParseableType {
 - name: listOfMaps
   list:
     elementType:
-      struct:
+      map:
         fields:
         - name: name
           type:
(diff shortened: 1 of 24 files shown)


merge/nested_test.go
```diff
@@ -78,23 +78,23 @@ var nestedTypeParser = func() typed.ParseableType {
     keys:
     - name
 - name: map
-  map:
-    elementType:
+  struct:
```
apelisse: should we find a better name than struct now?

schema/elements.go
```diff
@@ -95,6 +92,9 @@ type Struct struct {
 	// overlap between unions.
 	Unions []Union `yaml:"unions,omitempty"`
 
+	// DefaultElementType is the type of the structs's unknown fields.
```
apelisse: can it be unset? do we always allow unknown fields?

merge/nested_test.go
jennybuckley: map might be the better name

schema/elements.go
jennybuckley: Yes, if this is unset (set to an empty `schema.TypeRef{}`) then type resolution fails on unknown fields and so the validation still fails.

jennybuckley:
Beyond this, we just need to add
```
elementType:
  namedType: __untyped_deduced_
```
to maps(structs) generated by schemaconv when the openapi has x-kubernetes-preserve-unknown=true.

PR #82, opened 2019-05-21
Change union to not "guess" based on difference

apelisse:
We're still using the difference, but only to see if something wrong was sent.
    
The new logic is to only trust the disicriminator, which means that the
behavior for non-discriminated union is now back to before. It also
means that we can't use that process anymore to normalize applied
unions. Applied unions are normalized by trusting the applied fields
exclusively: applying two fields doesn't work, applying a discriminator
and a different value doesn't work.

Also rename `discriminatedBy` into `discriminatorValue`

### DIFF
--- a/schema/elements.go
+++ b/schema/elements.go
@@ -133,22 +133,27 @@ type UnionField struct {
 }
 
 // Union, or oneof, means that only one of multiple fields of a structure can be
-// set at a time. For backward compatibility reasons, and to help "dumb clients"
-// which are not aware of the union (or can't be aware of it because they
-// don't know what fields are part of the union), the code tolerates multiple
-// fields to be set but will try to detect which fields must be cleared (there
-// should never be more than two though):
-// - If there is a discriminator and its value has changed, clear all fields
-// but the one specified by the discriminator
-// - If there is no discriminator, or it hasn't changed, if new has two of the
-// fields set, remove the one that was set in old.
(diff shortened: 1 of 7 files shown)


apelisse:
/assign @liggitt 

Sending an update to the KEP now.

schema/elements.go
```diff
@@ -133,23 +133,28 @@ type UnionField struct {
...
 	// the discriminator and the field is done by using the Fields list
 	// below.
 	Discriminator *string `yaml:"discriminator,omitempty"`
 
+	// DeduceInvalidDiscriminator indicates if the discriminator
+	// should be updated automatically based on the fields set. This
+	// typically defaults to false since we don't want to deduce by
+	// default (the behavior exists to maintain compatibility on
+	// existing types and shouldn't be used for new types).
+	DeduceInvalidDiscriminator bool `yaml:"deduceInvalidDiscriminator,omitempty"`
```
jennybuckley: Is this just for where we can't add a discriminator? Are there any places like that?

jennybuckley:
/lgtm
Just a question about how we are planning to use this

schema/elements.go
apelisse: We only want to deduce the value of the discriminator for unions that already existed and didn't have a discriminator. New unions should have a discriminator from the beginning, it should be required, and there shouldn't be a need to deduce the discriminator at all.

PR #83, opened 2019-05-29
Treat unspecified keys as their default value if provided

jennybuckley:
Replacement for #79 
Relies on default key values being part of the schema so we can treat null as that value when constructing field paths, instead of actually doing defaulting.

Probably won't allow this except for some existing cases like containerPorts

cc @apelisse

### DIFF
--- a/merge/default_keys_test.go
+++ b/merge/default_keys_test.go
@@ -1,0 +1,338 @@
+/*
+Copyright 2019 The Kubernetes Authors.
+ Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+     http://www.apache.org/licenses/LICENSE-2.0
+ Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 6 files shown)


apelisse:
How do you plan on generating this?

jennybuckley:
Talked offline, but I'll write it here too.

The plan is to get this from the openapi, and that will come from the default markers in types.go that are going to eventually be defined. Like:
``` golang
// types.go
type Foo struct {
  // +default={"bar": "abc"}
  Foo *Bar
}
```
``` json
# openapi
{"properties": {"name": "foo", ... "default": {"bar": "abc"}}}
```
``` yaml
# schema.List
defaultedKeys:
- fieldName: foo
  defaultValue: "{\"bar\": \"abc\"}"
```

apelisse:
Right, I like that idea.

Are we going to generate "defaultedKeys" only for keys or for all values?

jennybuckley:
rebased

jennybuckley:
https://github.com/kubernetes-sigs/structured-merge-diff/pull/83#issuecomment-498448991

I think only for defaulted keys, but that can be decided once we are integrating this into kube-openapi

PR #87, opened 2019-07-09
Add benchmarks for merge package

apelisse:
This is testing the high-level part of SMD, we would probably need
lower-level benchmarks in this repo, but this at least gives us a
general idea of the direction.

/assign @lavalamp @jennybuckley

### DIFF
--- a/merge/deduced_test.go
+++ b/merge/deduced_test.go
@@ -534,3 +534,280 @@ func TestDeduced(t *testing.T) {
 		})
 	}
 }
+
+func BenchmarkDeducedSimple(b *testing.B) {
+	test := TestCase{
+		Ops: []Operation{
+			Apply{
+				Manager:    "default",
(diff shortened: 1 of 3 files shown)


merge/deduced_test.go
```diff
@@ -534,3 +534,280 @@ func TestDeduced(t *testing.T) {
...
+						    f:
+						    - name: n
+						      value: 1
+					`,
+			},
+		},
+		Object: `
+				a: 2
+				b:
+				  c:
```
lavalamp: mixed tabs and spaces?

apelisse: Actually, gofmt moved `Object` one level down when I moved TestCase outside the list, but obviously it doesn't indent yaml in strings...

apelisse: OK I fixed this.

PR #105, opened 2019-09-19
Add new cli command to generate fieldset

apelisse:
Creates a new command-line mode to generate fieldsets (as we work on new
formats, it can be useful to generate some fieldsets to compare them).
Also adds a test and some golden data for what the fieldset is supposed
to look like.


internal/cli/options.go
```diff
@@ -63,6 +64,7 @@ func (o *Options) AddFlags(fs *flag.FlagSet) {
 	fs.StringVar(&o.validatePath, "validate", "", "Path to a file to perform a validation operation on.")
 	fs.BoolVar(&o.merge, "merge", false, "Perform a merge operation between --lhs and --rhs")
 	fs.BoolVar(&o.compare, "compare", false, "Perform a compare operation between --lhs and --rhs")
+	fs.StringVar(&o.fieldset, "fieldset", "", "Build a field set for the given file.")
```
lavalamp: `"Path to a file for which we should build a fieldset"`

(otherwise people will have no idea how to pass "the given file")

lavalamp:
looks good, I think it's important to fix the help text.

A nice followup would be adding an output-format flag that supports e.g. pretty-printed json or yaml. (also somehow will eventually need to let people do the v2 format)

apelisse:
Yeah, pretty printing can easily be done with separate tool, and we never print the field set as yaml. When we have a second format I"ll make an option of course.

apelisse:
PTAL Updated the comment

PR #110, opened 2019-10-02
Minor tests improvements

apelisse:
Use Fatalf rather than Errorf so that tests stop rather than panic later.
Also fix a small bug in a test.


apelisse:
Fixed, PTAL

PR #111, opened 2019-10-02
Remove unused 'FromValue' code

apelisse:
The code is not used at all.


fieldpath/fromvalue.go
```diff
@@ -1,123 +0,0 @@
...
-*/
-
-package fieldpath
-
-import (
-	"sigs.k8s.io/structured-merge-diff/value"
-)
-
-// SetFromValue creates a set containing every leaf field mentioned in v.
-func SetFromValue(v value.Value) *Set {
```
lavalamp: Why do you want to delete this? It's a public function on this library?

fieldpath/fromvalue.go
```diff
@@ -1,123 +0,0 @@
...
-		return
-	}
-
-	// Leaf fields get added to the set.
-	if len(w.path) > 0 {
-		w.do(w.path)
-	}
-}
-
-// AssociativeListCandidateFieldNames lists the field names which are
```
lavalamp: Is this implemented in another way? I still feel pretty strongly that we should be doing this guessing for schemaless CRDs?

apelisse: We don't have this feature right now. I agree that we might want to do that eventually, but we' re not doing it right now and it is extra work to update this code anytime I want to experiment something, which is frustrating. I think we should delete it until we explicitly decide to use it but I don' t think it's worth the cost of constantly updating it right now.

apelisse: it's also confusing because you never know if the code is actually used or not ...

apelisse:
Don't have time to argue ;-)

PR #112, opened 2019-10-02
Improve Value and PathElement String()

apelisse:
PathElements are frequently turned into strings to find out if a path has already been evaluated. Values are part of PathElements so they are also turned into strings. That's a non-negligible improvement.

```
benchmark                                   old ns/op     new ns/op     delta
BenchmarkOperations/Pod/Update-12           384166        329574        -14.21%
BenchmarkOperations/Node/Update-12          576277        498512        -13.49%
BenchmarkOperations/Pod/Apply-12            543377        482016        -11.29%
BenchmarkOperations/Node/Apply-12           764054        684379        -10.43%
BenchmarkOperations/Endpoints/Apply-12      2065245       1997567       -3.28%
BenchmarkOperations/Endpoints/Update-12     55470         53943         -2.75%

benchmark                                  old allocs     new allocs     delta
BenchmarkOperations/Node/Update-12         1155           967            -16.28%
BenchmarkOperations/Node/Apply-12          1876           1581           -15.72%
BenchmarkOperations/Pod/Update-12          843            727            -13.76%
BenchmarkOperations/Pod/Apply-12           1179           1024           -13.15%
BenchmarkOperations/Endpoints/Apply-12     3142           3043           -3.15%
```

### DIFF
--- a/fieldpath/element.go
+++ b/fieldpath/element.go
@@ -17,8 +17,8 @@ limitations under the License.
 package fieldpath
 
 import (
-	"fmt"
 	"sort"
+	"strconv"
 	"strings"
 
 	"sigs.k8s.io/structured-merge-diff/value"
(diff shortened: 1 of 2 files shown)


lavalamp:
> PathElements are frequently turned into strings to find out if a path has already been evaluated

Where? This shouldn't be done?

lavalamp:
We should just store the path or path element and use Less (and/or use or implement Equals if not already).

apelisse:
Will replace with something else.

PR #113, opened 2019-10-02
Cache type information in Schema

apelisse:
When the schema is big, a lot of time is spent going through the list of each possible type to find the one it's looking for. No memory improvement, pure CPU/time:
```
benchmark                                   old ns/op     new ns/op     delta
BenchmarkOperations/Node/Update-12          576277        434979        -24.52%
BenchmarkOperations/Endpoints/Apply-12      2065245       1664846       -19.39%
BenchmarkOperations/Pod/Update-12           384166        312411        -18.68%
BenchmarkOperations/Node/Apply-12           764054        622976        -18.46%
BenchmarkOperations/Endpoints/Update-12     55470         46779         -15.67%
BenchmarkOperations/Pod/Apply-12            543377        478227        -11.99%
```

### DIFF
--- a/schema/elements.go
+++ b/schema/elements.go
@@ -16,9 +16,17 @@ limitations under the License.
 
 package schema
 
+import "sync"
+
 // Schema is a list of named types.
+//
+// Schema types are indexed in a map before the first search so this type
+// should be considered immutable.
 type Schema struct {
 	Types []TypeDef `yaml:"types,omitempty"`
+
+	once sync.Once
+	m    map[string]TypeDef
 }
 
 // A TypeSpecifier references a particular type in a schema.
@@ -92,6 +100,9 @@ const (
 //
 // Maps may also represent a type which is composed of a number of different fields.
 // Each field has a name and a type.
+//
+// Fields are indexed in a map before the first search so this type
+// should be considered immutable.
 type Map struct {
 	// Each struct field appears exactly once in this list. The order in
 	// this list defines the canonical field ordering.
@@ -117,6 +128,22 @@ type Map struct {
 	// The default behavior for maps is `separable`; it's permitted to
 	// leave this unset to get the default behavior.
 	ElementRelationship ElementRelationship `yaml:"elementRelationship,omitempty"`
+
+	once sync.Once
+	m    map[string]StructField
+}
+
+// FindField is a convenience function that returns the referenced StructField,
+// if it exists, or (nil, false) if it doesn't.
+func (m *Map) FindField(name string) (StructField, bool) {
+	m.once.Do(func() {
+		m.m = make(map[string]StructField, len(m.Fields))
+		for _, field := range m.Fields {
+			m.m[field.Name] = field
+		}
+	})
+	sf, ok := m.m[name]
+	return sf, ok
 }
 
 // UnionFields are mapping between the fields that are part of the union and
@@ -204,13 +231,15 @@ type List struct {
 
 // FindNamedType is a convenience function that returns the referenced TypeDef,
 // if it exists, or (nil, false) if it doesn't.
-func (s Schema) FindNamedType(name string) (TypeDef, bool) {
-	for _, t := range s.Types {
-		if t.Name == name {
-			return t, true
+func (s *Schema) FindNamedType(name string) (TypeDef, bool) {
+	s.once.Do(func() {
+		s.m = make(map[string]TypeDef, len(s.Types))
+		for _, t := range s.Types {
+			s.m[t.Name] = t
 		}
-	}
-	return TypeDef{}, false
+	})
+	t, ok := s.m[name]
+	return t, ok
 }
 
 // Resolve is a convenience function which returns the atom referenced, whether
--- a/schema/elements_test.go
+++ b/schema/elements_test.go
@@ -40,8 +40,45 @@ func TestFindNamedType(t *testing.T) {
 				Types: tt.defs,
 			}
 			td, exist := s.FindNamedType(tt.namedType)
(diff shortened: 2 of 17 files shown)


apelisse:
I forgot to call `FindType` ...

apelisse:
New benchmark data:
```
BenchmarkOperations/Pod/Update-12                      384166        252547        -34.26%
BenchmarkOperations/Pod/Apply-12                       543377        390523        -28.13%
BenchmarkOperations/Node/Update-12                     576277        395116        -31.44%
BenchmarkOperations/Node/Apply-12                      764054        593929        -22.27%
BenchmarkOperations/Endpoints/Update-12                55470         31772         -42.72%
BenchmarkOperations/Endpoints/Apply-12                 2065245       1321254       -36.02%
BenchmarkFromUnstructured/Pod-12                       195262        125611        -35.67%
BenchmarkFromUnstructured/Node-12                      320953        272154        -15.20%
BenchmarkFromUnstructured/Endpoints-12                 8145833       6478752       -20.47%
```

schema/elements.go
```diff
@@ -117,6 +122,20 @@ type Map struct {
 	// The default behavior for maps is `separable`; it's permitted to
 	// leave this unset to get the default behavior.
 	ElementRelationship ElementRelationship `yaml:"elementRelationship,omitempty"`
+
+	once sync.Once
+	m    map[string]TypeRef
+}
+
+func (m *Map) FindType(name string) (TypeRef, bool) {
```
lavalamp: Maps have fields, not types?

apelisse: the name is probably not the best. We're looking for the type for a specific field

schema/elements.go
```diff
@@ -16,9 +16,14 @@ limitations under the License.
...
 package schema
 
+import "sync"
+
 // Schema is a list of named types.
 type Schema struct {
 	Types []TypeDef `yaml:"types,omitempty"`
+
+	once sync.Once
+	m    map[string]TypeDef
```
lavalamp: I'm not super excited about this, it means it's no longer OK to modify a schema if ?some? function has been called on it.

schema/elements.go
apelisse: `FindFieldType`?

schema/elements.go
apelisse: All of this would be const/final if we could, there is no reason to change the schema while processing things?

typed/helpers.go
```diff
@@ -99,7 +99,7 @@ func (ef errorFormatter) prefixError(prefix string, err error) ValidationErrors
 type atomHandler interface {
 	doScalar(schema.Scalar) ValidationErrors
 	doList(schema.List) ValidationErrors
-	doMap(schema.Map) ValidationErrors
+	doMap(*schema.Map) ValidationErrors
```
lavalamp: Please change everything not just map if you do this.

schema/elements.go
lavalamp: At a minimum, you need to make that clear in the documentation on these types.

schema/elements.go
lavalamp: No, FindField--callers can get the type from the result.

apelisse:
Updated, PTAL. Still not sure how to fix the test though :-/

jennybuckley:
@apelisse its failing because the test uses quick.Check which does a reflect.DeepEqual?

jennybuckley:
or because of testing/quick.Check generating a map

apelisse:
`quick.Check` doesn't want to deal with types that have unexported fields :-/. I guess I have to randomly generate values? I can probably use gofuzz instead

schema/elements.go
```diff
@@ -117,6 +128,22 @@ type Map struct {
 	// The default behavior for maps is `separable`; it's permitted to
 	// leave this unset to get the default behavior.
 	ElementRelationship ElementRelationship `yaml:"elementRelationship,omitempty"`
+
+	once sync.Once
+	m    map[string]StructField
+}
+
+// FindNamedType is a convenience function that returns the referenced StructField,
```
lavalamp: Comment doesn't match new name.

schema/elements.go
```diff
@@ -117,6 +128,22 @@ type Map struct {
...
 	// leave this unset to get the default behavior.
 	ElementRelationship ElementRelationship `yaml:"elementRelationship,omitempty"`
+
+	once sync.Once
+	m    map[string]StructField
+}
+
+// FindField is a convenience function that returns the referenced StructField,
+// if it exists, or (nil, false) if it doesn't.
+func (m *Map) FindField(name string) (StructField, bool) {
```
lavalamp: will any test fail if someone breaks this (or the next) function?

schema/elements_test.go
```diff
@@ -50,6 +50,43 @@ func TestFindNamedType(t *testing.T) {
...
+	}
+	for _, tt := range tests {
+		tt := tt
+		t.Run(tt.testName, func(t *testing.T) {
+			t.Parallel()
+			s := Map{
+				Fields: tt.defs,
+			}
+			sf, exist := s.FindField(tt.fieldName)
+			if !reflect.DeepEqual(sf, tt.expectStructField) {
```
lavalamp: Side note: this usage of DeepEqual illustrates why this "embedded private cache field" isn't a good pattern...

Not asking for a change just noting.

apelisse: Wait, that's not the right code

apelisse: Yeah, we do have an equal function, I could have used that.

apelisse: Done!

PR #114, opened 2019-10-03
Use PathElementSet and similar to index PathElement

apelisse:
Currently, we index PathElement by turning them into strings, which is neither fast nor accurate.
We can simply use the `PathElementSet` when we need a set, and we can build a map of PathElement to Value when needed using very similar code.

Improvement:
```
benchmark                                              old ns/op     new ns/op     delta
BenchmarkOperations/Pod/Update-12                      187994        162602        -13.51%
BenchmarkOperations/Pod/Apply-12                       525375        452963        -13.78%
BenchmarkOperations/Node/Update-12                     273261        220959        -19.14%
BenchmarkOperations/Node/Apply-12                      760287        667991        -12.14%
BenchmarkOperations/Endpoints/Update-12                32649         32794         +0.44%
BenchmarkOperations/Endpoints/Apply-12                 2072047       1862396       -10.12%
BenchmarkFromUnstructured/Pod-12                       202893        182379        -10.11%
BenchmarkFromUnstructured/Node-12                      331193        308072        -6.98%
BenchmarkFromUnstructured/Endpoints-12                 8057365       7754632       -3.76%

benchmark                                              old allocs     new allocs     delta
BenchmarkOperations/Pod/Update-12                      445            284            -36.18%
BenchmarkOperations/Pod/Apply-12                       1179           771            -34.61%
BenchmarkOperations/Node/Update-12                     601            336            -44.09%
BenchmarkOperations/Node/Apply-12                      1876           1138           -39.34%
BenchmarkOperations/Endpoints/Update-12                74             74             +0.00%
BenchmarkOperations/Endpoints/Apply-12                 3142           1120           -64.35%
BenchmarkFromUnstructured/Pod-12                       477            392            -17.82%
BenchmarkFromUnstructured/Node-12                      1107           900            -18.70%
BenchmarkFromUnstructured/Endpoints-12                 21115          19089          -9.60%
```

### DIFF
--- a/typed/merge.go
+++ b/typed/merge.go
@@ -168,8 +168,9 @@ func (w *mergingWalker) visitListItems(t *schema.List, lhs, rhs *value.List) (er
 	rhsOrder := []fieldpath.PathElement{}
 
 	// First, collect all RHS children.
-	observedRHS := map[string]value.Value{}
+	var observedRHS fieldpath.PathElementValueMap
 	if rhs != nil {
+		observedRHS = fieldpath.MakePathElementValueMap(len(rhs.Items))
 		for i, child := range rhs.Items {
 			pe, err := listItemToPathElement(t, i, child)
 			if err != nil {
@@ -179,18 +180,18 @@ func (w *mergingWalker) visitListItems(t *schema.List, lhs, rhs *value.List) (er
 				// this element.
 				continue
 			}
-			keyStr := pe.String()
-			if _, found := observedRHS[keyStr]; found {
-				errs = append(errs, w.errorf("rhs: duplicate entries for key %v", keyStr)...)
+			if _, ok := observedRHS.Get(pe); ok {
+				errs = append(errs, w.errorf("rhs: duplicate entries for key %v", pe.String())...)
 			}
-			observedRHS[keyStr] = child
+			observedRHS.Insert(pe, child)
 			rhsOrder = append(rhsOrder, pe)
 		}
 	}
 
 	// Then merge with LHS children.
-	observedLHS := map[string]struct{}{}
+	var observedLHS fieldpath.PathElementSet
 	if lhs != nil {
+		observedLHS = fieldpath.MakePathElementSet(len(lhs.Items))
 		for i, child := range lhs.Items {
 			pe, err := listItemToPathElement(t, i, child)
 			if err != nil {
@@ -200,15 +201,14 @@ func (w *mergingWalker) visitListItems(t *schema.List, lhs, rhs *value.List) (er
 				// this element.
 				continue
 			}
-			keyStr := pe.String()
-			if _, found := observedLHS[keyStr]; found {
-				errs = append(errs, w.errorf("lhs: duplicate entries for key %v", keyStr)...)
+			if observedLHS.Has(pe) {
+				errs = append(errs, w.errorf("lhs: duplicate entries for key %v", pe.String())...)
 				continue
 			}
-			observedLHS[keyStr] = struct{}{}
+			observedLHS.Insert(pe)
 			w2 := w.prepareDescent(pe, t.ElementType)
 			w2.lhs = &child
-			if rchild, ok := observedRHS[keyStr]; ok {
+			if rchild, ok := observedRHS.Get(pe); ok {
 				w2.rhs = &rchild
 			}
 			if newErrs := w2.merge(); len(newErrs) > 0 {
@@ -217,22 +217,22 @@ func (w *mergingWalker) visitListItems(t *schema.List, lhs, rhs *value.List) (er
 				out.Items = append(out.Items, *w2.out)
 			}
 			w.finishDescent(w2)
-			// Keep track of children that have been handled
-			delete(observedRHS, keyStr)
 		}
 	}
 
-	for _, rhsToCheck := range rhsOrder {
-		if unmergedChild, ok := observedRHS[rhsToCheck.String()]; ok {
-			w2 := w.prepareDescent(rhsToCheck, t.ElementType)
-			w2.rhs = &unmergedChild
-			if newErrs := w2.merge(); len(newErrs) > 0 {
-				errs = append(errs, newErrs...)
-			} else if w2.out != nil {
-				out.Items = append(out.Items, *w2.out)
-			}
-			w.finishDescent(w2)
+	for _, pe := range rhsOrder {
+		if observedLHS.Has(pe) {
+			continue
 		}
+		value, _ := observedRHS.Get(pe)
+		w2 := w.prepareDescent(pe, t.ElementType)
+		w2.rhs = &value
+		if newErrs := w2.merge(); len(newErrs) > 0 {
+			errs = append(errs, newErrs...)
+		} else if w2.out != nil {
+			out.Items = append(out.Items, *w2.out)
+		}
+		w.finishDescent(w2)
 	}
 
 	if len(out.Items) > 0 {
--- a/typed/validate.go
+++ b/typed/validate.go
@@ -142,7 +142,7 @@ func (v *validatingObjectWalker) doScalar(t *schema.Scalar) ValidationErrors {
 }
 
 func (v *validatingObjectWalker) visitListItems(t *schema.List, list *value.List) (errs ValidationErrors) {
-	observedKeys := map[string]struct{}{}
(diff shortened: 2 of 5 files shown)


apelisse:
More or less replaces #112

typed/validate.go
```diff
@@ -152,11 +152,10 @@ func (v *validatingObjectWalker) visitListItems(t schema.List, list *value.List)
 			// this element.
 			continue
 		}
-		keyStr := pe.String()
-		if _, found := observedKeys[keyStr]; found {
-			errs = append(errs, v.errorf("duplicate entries for key %v", keyStr)...)
+		if observedKeys.Has(pe) {
```
jennybuckley: Is this really faster? i know the benchmarks are better but it makes the constant time operation with something that has to sort a slice and do a binary search

https://github.com/kubernetes-sigs/structured-merge-diff/blob/master/fieldpath/element.go#L243-L255

apelisse: cc @lavalamp 

It is slightly faster but most importantly it's more correct. Using the string as the index is a little wrong since it's not 100% canonicalized and is not meant to be used for "production use".

apelisse: Allocating a buffer for the string, converting the pathelement to a string, hashing the string, comparing the strings, is quite expensive too

lavalamp: String() was actually written to make this work, so this should be equally correct.

The set impl is O(log n) but with a low enough constant factor that it was faster than the O(1) thing it replaced. Now we don't have a O(1) thing to compare it with, I assume at some size it would stop being faster.

typed/validate.go
```diff
@@ -142,7 +142,7 @@ func (v *validatingObjectWalker) doScalar(t schema.Scalar) ValidationErrors {
 }
 
 func (v *validatingObjectWalker) visitListItems(t schema.List, list *value.List) (errs ValidationErrors) {
-	observedKeys := map[string]struct{}{}
+	observedKeys := fieldpath.PathElementSet{}
```
lavalamp: Pre-allocating the array to be the right size might save a few allocs. (followup)

apelisse: Clearly, thanks!

typed/merge.go
```diff
@@ -168,7 +219,7 @@ func (w *mergingWalker) visitListItems(t schema.List, lhs, rhs *value.List) (err
 	rhsOrder := []fieldpath.PathElement{}
 
 	// First, collect all RHS children.
-	observedRHS := map[string]value.Value{}
```
lavalamp: I'm surprised I missed these uses of strings when I was working on this last quarter...

apelisse: 1. We had much bigger things to improve,
2. Maybe because of the test data?!

Anyways, /shrug

apelisse: This PR is a fairly small improvement compared to some of the things that are coming.

typed/merge.go
```diff
@@ -17,11 +17,62 @@ limitations under the License.
...
+
+func (s *pathElementValueMap) Get(pe fieldpath.PathElement) *value.Value {
+	loc := sort.Search(len(s.members), func(i int) bool {
+		return !s.members[i].PathElement.Less(pe)
+	})
+	if loc == len(s.members) {
+		return nil
+	}
+	if s.members[loc].PathElement.Equals(pe) {
+		return &s.members[loc].Value
```
lavalamp: This is very risky, a subsequent insert will make the pointer point to the wrong thing.

apelisse: yeah, we could make it return a copy, bool to be safer?

typed/merge.go
```diff
@@ -17,11 +17,62 @@ limitations under the License.
...
 	"sigs.k8s.io/structured-merge-diff/value"
 )
 
+type pathElementValueMap struct {
+	members sortedPathElementValues
+}
+
+type pathElementValue struct {
+	PathElement fieldpath.PathElement
+	Value       value.Value
```
lavalamp: I was going to say, it might be worth testing whether *value.Value is better for perf (less data to copy around), but below I found an even better reason.

lavalamp: An alternative is to never move objects in the slice but instead keep a parallel []int containing the sorted order.

apelisse: Let's not over-optimize. Values are going to change significantly anyway.

typed/merge.go
lavalamp: Yes, that'd be better.

typed/merge.go
```diff
@@ -17,11 +17,62 @@ limitations under the License.
...
 
 import (
+	"sort"
+
 	"sigs.k8s.io/structured-merge-diff/fieldpath"
 	"sigs.k8s.io/structured-merge-diff/schema"
 	"sigs.k8s.io/structured-merge-diff/value"
 )
 
+type pathElementValueMap struct {
```
lavalamp: When I wrote the set version of this, getting the sort call and less function to do the right thing was tricky. I think unit tests of this particular thing are in order to keep it working :/

Alternatively consider moving this into the same package as the set type where it'd be easier to share code and tests?

apelisse: Yeah, we're doing very similar things everywhere, but I'd also like to avoid having `PathElementMap` to `interface{}` (which we could use in all three places)...

lavalamp: Why? I think we need either tests or shared implementation. I don't think interface{} is going to make it slow if that's what you're worried about?

apelisse: Yeah, I suspect it has to escape to heap when you cast to `interface{}`, but I might be wrong, it's worth trying.

apelisse:
I think we should probably re-use the code in other places rather than add another implementation of the set, but I don't have a lot of time for that right now so I've decided to add a test instead.
I can create an issue to track that if needed. And I'll add a comment.

typed/validate.go
apelisse: I finally remembered to do this one, another 20% improvement ;-)

apelisse:
OK, I've added the pre-allocation of the right size for the arrays, here are the new numbers, pretty good:
```
benchmark                                              old ns/op     new ns/op     delta
BenchmarkDeducedSimple-12                              113151        114199        +0.93%
BenchmarkDeducedNested-12                              319775        296626        -7.24%
BenchmarkDeducedNestedAcrossVersion-12                 329968        328015        -0.59%
BenchmarkLeafConflictAcrossVersion-12                  114734        107390        -6.40%
BenchmarkMultipleApplierRecursiveRealConversion-12     1565560       1588022       +1.43%
BenchmarkOperations/Pod/Update-12                      187994        115494        -38.57%
BenchmarkOperations/Pod/Apply-12                       525375        294227        -44.00%
BenchmarkOperations/Node/Update-12                     273261        172191        -36.99%
BenchmarkOperations/Node/Apply-12                      760287        502819        -33.86%
BenchmarkOperations/Endpoints/Update-12                32649         22770         -30.26%
BenchmarkOperations/Endpoints/Apply-12                 2072047       1003676       -51.56%
BenchmarkFromUnstructured/Pod-12                       202893        115718        -42.97%
BenchmarkFromUnstructured/Node-12                      331193        234963        -29.06%
BenchmarkFromUnstructured/Endpoints-12                 8057365       6069791       -24.67%

benchmark                                              old allocs     new allocs     delta
BenchmarkDeducedSimple-12                              552            552            +0.00%
BenchmarkDeducedNested-12                              1534           1498           -2.35%
BenchmarkDeducedNestedAcrossVersion-12                 1602           1566           -2.25%
BenchmarkLeafConflictAcrossVersion-12                  532            532            +0.00%
BenchmarkMultipleApplierRecursiveRealConversion-12     5229           5232           +0.06%
BenchmarkOperations/Pod/Update-12                      445            282            -36.63%
BenchmarkOperations/Pod/Apply-12                       1179           744            -36.90%
BenchmarkOperations/Node/Update-12                     601            342            -43.09%
BenchmarkOperations/Node/Apply-12                      1876           1103           -41.20%
BenchmarkOperations/Endpoints/Update-12                74             72             -2.70%
BenchmarkOperations/Endpoints/Apply-12                 3142           1103           -64.89%
BenchmarkFromUnstructured/Pod-12                       477            370            -22.43%
BenchmarkFromUnstructured/Node-12                      1107           854            -22.85%
BenchmarkFromUnstructured/Endpoints-12                 21115          19076          -9.66%
```

lavalamp:
/lgtm
/approve

Sorry, didn't realize this was waiting on me.

PR #117, opened 2019-10-09
Change "value" package altogether to avoid copying everything

apelisse:
This is a WIP for a few reasons:
- Some things are much better, look at perf below,
- Some things are much worse, though it's still better overall,
- It moved some of the allocations from one place to another, I have ways to remove some altogether,
- I need to add more tests to value package
- This is based on the pathelement-set PR (#114)

Perf so far:
```
benchmark                                              old ns/op     new ns/op     delta
BenchmarkDeducedSimple-12                              109043        91375         -16.20%
BenchmarkDeducedNested-12                              360530        336906        -6.55%
BenchmarkDeducedNestedAcrossVersion-12                 397451        370451        -6.79%
BenchmarkLeafConflictAcrossVersion-12                  94595         96292         +1.79%
BenchmarkMultipleApplierRecursiveRealConversion-12     1565003       1679940       +7.34%
BenchmarkOperations/Pod/Update-12                      100904        105945        +5.00%
BenchmarkOperations/Pod/Apply-12                       285154        270291        -5.21%
BenchmarkOperations/Node/Update-12                     158819        162062        +2.04%
BenchmarkOperations/Node/Apply-12                      494102        449763        -8.97%
BenchmarkOperations/Endpoints/Update-12                12728         15428         +21.21%
BenchmarkOperations/Endpoints/Apply-12                 1002018       1717987       +71.45%
BenchmarkFromUnstructured/Pod-12                       115741        34212         -70.44%
BenchmarkFromUnstructured/Node-12                      238225        69035         -71.02%
BenchmarkFromUnstructured/Endpoints-12                 6223548       1721100       -72.35%

benchmark                                              old allocs     new allocs     delta
BenchmarkDeducedSimple-12                              570            554            -2.81%
BenchmarkDeducedNested-12                              1910           1635           -14.40%
BenchmarkDeducedNestedAcrossVersion-12                 1978           1813           -8.34%
BenchmarkLeafConflictAcrossVersion-12                  505            540            +6.93%
BenchmarkMultipleApplierRecursiveRealConversion-12     5183           5045           -2.66%
BenchmarkOperations/Pod/Update-12                      255            399            +56.47%
BenchmarkOperations/Pod/Apply-12                       717            1032           +43.93%
BenchmarkOperations/Node/Update-12                     315            559            +77.46%
BenchmarkOperations/Node/Apply-12                      1076           1574           +46.28%
BenchmarkOperations/Endpoints/Update-12                45             69             +53.33%
BenchmarkOperations/Endpoints/Apply-12                 1076           7138           +563.38%
BenchmarkFromUnstructured/Pod-12                       370            161            -56.49%
BenchmarkFromUnstructured/Node-12                      854            312            -63.47%
BenchmarkFromUnstructured/Endpoints-12                 19076          7024           -63.18%
```


fieldpath/serialize-pe.go
```diff
@@ -83,14 +84,21 @@ func DeserializePathElement(s string) (PathElement, error) {
...
+		iter.ReadObjectCB(func(iter *jsoniter.Iterator, key string) bool {
+			v, err := value.ReadJSONIter(iter)
+			if err != nil {
+				iter.Error = err
+				return false
+			}
+			kv = append(kv, KeyValue{Key: key, Value: v})
+			return true
+		})
+		sort.SliceStable(kv, func(i, j int) bool {
```
apelisse: I should use `KeyValueSort`

fieldpath/path.go
```diff
@@ -88,15 +89,18 @@ func MakePath(parts ...interface{}) (Path, error) {
...
 		case string:
 			fp = append(fp, PathElement{FieldName: &t})
-		case []value.Field:
+		case []KeyValue:
 			if len(t) == 0 {
 				return nil, fmt.Errorf("associative list key type path elements must have at least one key (got zero)")
 			}
-			fp = append(fp, PathElement{Key: &value.Map{Items: t}})
-		case value.Value:
+			sort.SliceStable(t, func(i, j int) bool {
```
apelisse: I should use `KeyValueSort`

fieldpath/set_test.go
```diff
@@ -46,20 +44,23 @@ var randomPathMaker = randomPathAlphabet(MakePathOrDie(
...
+	_V(2),
+	_V(3),
+	_V("aa"),
+	_V("ab"),
+	_V(true),
 	1, 2, 3, 4,
 ))
 
 func BenchmarkFieldSet(b *testing.B) {
+	// Before: 32, 48, 64
```
apelisse: Not needed ...

apelisse:
I've added a couple of commits, the last one is definitely WIP since it's not tracking the pathelement (only used for errors, but we can refactor error handling to only build the path on errors).

These two commits significantly improve the situation though (all tests are still passing):
```
benchmark                                              old ns/op     new ns/op     delta
BenchmarkDeducedSimple-12                              117303        91303         -22.16%
BenchmarkDeducedNested-12                              318361        337676        +6.07%
BenchmarkDeducedNestedAcrossVersion-12                 317505        352831        +11.13%
BenchmarkLeafConflictAcrossVersion-12                  115084        93126         -19.08%
BenchmarkMultipleApplierRecursiveRealConversion-12     1649080       1588821       -3.65%
BenchmarkOperations/Pod/Update-12                      351283        108252        -69.18%
BenchmarkOperations/Pod/Apply-12                       516962        253059        -51.05%
BenchmarkOperations/Node/Update-12                     569627        166258        -70.81%
BenchmarkOperations/Node/Apply-12                      755372        415201        -45.03%
BenchmarkOperations/Endpoints/Update-12                56002         15827         -71.74%
BenchmarkOperations/Endpoints/Apply-12                 2081751       30775         -98.52%
BenchmarkFromUnstructured/Pod-12                       202841        26335         -87.02%
BenchmarkFromUnstructured/Node-12                      327686        54673         -83.32%
BenchmarkFromUnstructured/Endpoints-12                 7909537       1097087       -86.13%

benchmark                                              old allocs     new allocs     delta
BenchmarkDeducedSimple-12                              572            543            -5.07%
BenchmarkDeducedNested-12                              1562           1588           +1.66%
BenchmarkDeducedNestedAcrossVersion-12                 1630           1766           +8.34%
BenchmarkLeafConflictAcrossVersion-12                  544            529            -2.76%
BenchmarkMultipleApplierRecursiveRealConversion-12     5268           4923           -6.55%
BenchmarkOperations/Pod/Update-12                      843            399            -52.67%
BenchmarkOperations/Pod/Apply-12                       1179           988            -16.20%
BenchmarkOperations/Node/Update-12                     1155           559            -51.60%
BenchmarkOperations/Node/Apply-12                      1876           1453           -22.55%
BenchmarkOperations/Endpoints/Update-12                100            69             -31.00%
BenchmarkOperations/Endpoints/Apply-12                 3143           127            -95.96%
BenchmarkFromUnstructured/Pod-12                       477            49             -89.73%
BenchmarkFromUnstructured/Node-12                      1107           109            -90.15%
BenchmarkFromUnstructured/Endpoints-12                 21114          7              -99.97%
```

fieldpath/element.go
```diff
@@ -24,6 +24,58 @@ import (
 	"sigs.k8s.io/structured-merge-diff/value"
 )
 
+type KeyValue struct {
```
lavalamp: Is this the same as value.Field? https://github.com/kubernetes-sigs/structured-merge-diff/blob/master/value/value.go#L127 Can we reuse that?

fieldpath/element.go
```diff
@@ -24,6 +24,58 @@ import (
...
+	Key   string
+	Value value.Value
+}
+
+func SortKeyValues(kv []KeyValue) {
```
lavalamp: Godoc comments on public things in this file?

fieldpath/element.go
```diff
@@ -24,6 +24,58 @@ import (
...
+			kv[0], kv[1] = kv[1], kv[0]
+		}
+		return
+	}
+	sort.SliceStable(kv, func(i, j int) bool {
+		return kv[i].Key < kv[j].Key
+	})
+}
+
+func KeyValueLess(lhs, rhs []KeyValue) bool {
```
lavalamp: This is actually comparing lists of keyvalue pairs, "KeyValueListLess"?

fieldpath/element.go
apelisse: I'm curious if I should make a type for `[]KeyValue` and make that a method of the type.

fieldpath/element.go
apelisse: > I'm curious if I should make a type for `[]KeyValue` and make that a method of the type.

And then I would make this a Less method of the `[]KeyValue` type.

fieldpath/element.go
```diff
@@ -123,17 +172,16 @@ func (e PathElement) String() string {
...
-	out := []value.Field{}
+	out := []KeyValue{}
 	for i := 0; i < len(nameValues)-1; i += 2 {
-		out = append(out, value.Field{
-			Name:  nameValues[i].(string),
-			Value: nameValues[i+1].(value.Value),
-		})
+		out = append(out, KeyValue{Key: nameValues[i].(string), Value: nameValues[i+1]})
 	}
+	SortKeyValues(out)
```
lavalamp: I don't claim it's better without looking at all call sites, but it's conceivable that we could panic here if they provided entries in the wrong order, since it may be misleading (for the humans reading failing tests) to specify things in one order and have them come out in a different order.

fieldpath/element.go
lavalamp: Maybe even do a `type KeyValueList []KeyValue`? And consider making this call work like `lhs.Less(rhs)`?

fieldpath/element.go
lavalamp: Alternative interface if you make a type: `func (kvl KeyValueList) Sort() { ... }`

apelisse: Yeah, that's what I had in mind

fieldpath/element_test.go
```diff
@@ -91,7 +91,7 @@ func TestPathElementLess(t *testing.T) {
 		}, {
 			name: "FieldName-3",
 			a:    PathElement{FieldName: strptr("capybara")},
-			b:    PathElement{Key: &value.Map{Items: []value.Field{{Name: "dog", Value: value.IntValue(3)}}}},
+			b:    PathElement{Key: &[]KeyValue{{"dog", 3}}},
```
lavalamp: hm, it'd be nice to use the KeyByFields function. Maybe declare a local:

```
func kvlptr(args... interface{}) *KeyValueList {
  kvl := KeyByFields(args...)
  return &kvl
}
```

or something?

value/value.go
```diff
@@ -266,96 +280,179 @@ func (m *Map) Less(rhs *Map) bool {
...
+		vk, ok := key.(string)
+		if !ok {
+			continue
 		}
-		return "[" + strings.Join(strs, ",") + "]"
-	case v.MapValue != nil:
-		strs := []string{}
-		for _, i := range v.MapValue.Items {
-			strs = append(strs, fmt.Sprintf("%v=%v", i.Name, i.Value))
+		if !iter(vk, m[key]) {
```
lavalamp: why not `for key, val := range m {` to avoid the lookup?

value/value.go
```diff
@@ -266,96 +280,179 @@ func (m *Map) Less(rhs *Map) bool {
...
-		return fmt.Sprintf("%v", *v.BooleanValue)
-	case v.ListValue != nil:
-		strs := []string{}
-		for _, item := range v.ListValue.Items {
-			strs = append(strs, item.String())
+func (m MapInterface) Iterate(iter func(string, Value) bool) {
+	for key := range m {
+		vk, ok := key.(string)
+		if !ok {
+			continue
```
lavalamp: wouldn't this be a major correctness problem if it happened?

value/value.go
apelisse: Good point

value/value.go
apelisse: Yes, when I started that I thought maybe we would sometimes need to terminate the loop early, but it's not needed so I clearly don't have a strong opinion about that.

lavalamp:
so, overall, we currently are doing 2 conversions, typed struct -> unstructured -> value.Value.

This change removes the latter conversion, at some cost to this library's interface & functionality (?).

But to get rid of both conversions, the only choice is to make some sort of interface and some way of letting our existing typed structs implement it. In person Antoine said that iterating a map is expensive as the closure causes the variables to escape and be placed on the heap, making a lot of allocations. But without this I don't see how we can get rid of the 2nd conversion.

So, my current recommendation is:
* preserve the existing value type / package
* make an interface with the features of the current value.Value (which the existing value type will implement)
* make a wrapper that exposes map[string]interface{} in terms of this interface (intermediate checkpoint)
* make a wrapper that uses reflection to to expose arbitrary go objects in terms of this interface (2nd checkpoint - usually needed as a fallback for generated code)
* make a wrapper (or generate code, ugh) functions that expose our built-in resources in terms of this interface - this has to go in the main repo...

I admit to being slightly biased to like the existing value.Value because I think it is just a better version of an unstructured k8s object. An alternative to this PR is to remove the unstructured step and go directly from struct to value.Value in k8s code - basically swap out the internals of the unstructured object there.

apelisse:
> - make an interface with the features of the current value.Value (which the existing value type will implement)
> - make a wrapper that exposes map[string]interface{} in terms of this interface (intermediate checkpoint)

I've started working on these points, but so far the results are unsatisfying. I'll quickly check for something obvious, but I won't have a lot of time for that. Unless we actually manage to make your suggested changes work, I would recommend we move forward with this PR.

```
benchmark                                  old ns/op     new ns/op     delta
BenchmarkFromUnstructured/Pod-12           23063         43953         +90.58%
BenchmarkFromUnstructured/Node-12          47925         102539        +113.96%
BenchmarkFromUnstructured/Endpoints-12     865869        1764150       +103.74%

benchmark                                  old allocs     new allocs     delta
BenchmarkFromUnstructured/Pod-12           49             245            +400.00%
BenchmarkFromUnstructured/Node-12          109            625            +473.39%
BenchmarkFromUnstructured/Endpoints-12     7              9038           +129014.29%
```

apelisse:
I just looked at pprof and confirmed with the escape analysis (`-gcflags '-m'`) that the extra wrapper is the thing that escapes to the heap.

I want to note that while I agree with your point in principle, and definitely believe that this is a possible further improvement, I think we should move forward with this existing PR code since it provides a significant improvement that we shouldn't give-up (and also given that there is no concern regarding correctness of this code).

apelisse:
Replaced by #129

PR #118, opened 2019-10-23
Simplify PathElement Keys type

apelisse:
As discussed in #117

### DIFF
--- a/fieldpath/element.go
+++ b/fieldpath/element.go
@@ -35,7 +35,7 @@ type PathElement struct {
 	// Key selects the list element which has fields matching those given.
 	// The containing object must be an associative list with map typed
 	// elements.
-	Key *value.Map
+	Key *value.FieldList
 
 	// Value selects the list element with the given value. The containing
 	// object must be an associative list with a primitive typed element
@@ -62,7 +62,7 @@ func (e PathElement) Less(rhs PathElement) bool {
 		if rhs.Key == nil {
 			return true
 		}
-		return e.Key.Less(rhs.Key)
+		return e.Key.Less(*rhs.Key)
 	} else if rhs.Key != nil {
 		return false
 	}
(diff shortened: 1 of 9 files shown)


apelisse:
@lavalamp

apelisse:
This is shorter :-)

apelisse:
> I don't claim it's better without looking at all call sites, but it's conceivable that we could panic here if they provided entries in the wrong order, since it may be misleading (for the humans reading failing tests) to specify things in one order and have them come out in a different order.

I kept it this way because there are many call sites

fieldpath/path.go
```diff
@@ -88,11 +88,11 @@ func MakePath(parts ...interface{}) (Path, error) {
 			fp = append(fp, PathElement{Index: &t})
 		case string:
 			fp = append(fp, PathElement{FieldName: &t})
-		case []value.Field:
-			if len(t) == 0 {
+		case *value.FieldList:
```
lavalamp: hrm, can we avoid an allocation if we just store []value.FieldList?

apelisse: I'm not super concerned about allocations on this path since this is only used in test AFAICT. Also, we would need to take the pointer to this anyway?

value/value.go
```diff
@@ -129,6 +129,55 @@ type Field struct {
 	Value Value
 }
 
+type FieldList []Field
```
lavalamp: Godoc comments?

fieldpath/path.go
lavalamp: Oh, this is MakePath, sorry. I don't care too much about this, let me find a better place to make a comment.

fieldpath/element.go
```diff
@@ -35,7 +35,7 @@ type PathElement struct {
 	// Key selects the list element which has fields matching those given.
 	// The containing object must be an associative list with map typed
 	// elements.
-	Key *value.Map
+	Key *value.FieldList
```
lavalamp: Now that it's not a map, I *think* we can just make this a FieldList? A zero-length field list is an invalid key, right? A slice is already a pointer, no need to double-point?

lavalamp: Antoine mentioned IRL that the size of this object might be a concern, so let's merge this and we can always change this later.

PR #120, opened 2019-10-28
Fix wg-apply mailing list link

apelisse:
I'm sure we can also fix the slack link, not sure how to refer to a specific channel.


jpbetz:
That link works.

/lgtm

apelisse:
@lavalamp Looks like we have to add Joe somewhere to a list.

PR #121, opened 2019-10-29
WIP: Experimental new value interface

apelisse:
This is not meant to work yet, this is work-in-progress to test having an interface for value so potentially we wouldn't have to convert anything in anything.


value/value.go
```diff
@@ -182,32 +127,325 @@ func (f FieldList) Less(rhs FieldList) bool {
...
+	Float() float64
+	String() string
+
+	Interface() interface{}
 }
 
-// Less compares two lists lexically.
-func (l *List) Less(rhs *List) bool {
+type List interface {
+	Interface() []interface{}
```
lavalamp: I wouldn't have expected this method?

apelisse: I don't know if I use it?

value/value.go
```diff
@@ -182,32 +127,325 @@ func (f FieldList) Less(rhs FieldList) bool {
...
+		return true
+	}
+	return false
+}
+
+func (v ValueInterface) Map() Map {
+	if v.Value == nil {
+		return MapString(nil)
+	}
+	switch t := v.Value.(type) {
```
lavalamp: I would not expect this?

value/value.go
```diff
@@ -182,32 +127,325 @@ func (f FieldList) Less(rhs FieldList) bool {
...
-// Less compares two lists lexically.
-func (l *List) Less(rhs *List) bool {
+type List interface {
+	Interface() []interface{}
+	Length() int
+	Iterate(func(int, Value))
+	At(int) Value
+}
+
+type ValueInterface struct {
```
lavalamp: This is an implementation, not an interface?

value/value.go
jennybuckley: Didn't this already exist?

lavalamp: ah, ok, I think I understand now.

jennybuckley: Oh I was only looking at the last commit

value/value.go
lavalamp: Specifically, this is the k8s unstructured implementation. I think this implementation should go in a different package and get a different name :)

value/value.go
```diff
@@ -319,96 +502,68 @@ func (m *Map) Less(rhs *Map) bool {
...
+		vk, ok := key.(string)
+		if !ok {
+			continue
 		}
-		return "[" + strings.Join(strs, ",") + "]"
-	case v.MapValue != nil:
-		strs := []string{}
-		for _, i := range v.MapValue.Items {
-			strs = append(strs, fmt.Sprintf("%v=%v", i.Name, i.Value))
+		if !iter(vk, ValueInterface{Value: value}) {
```
lavalamp: This is the thing that escapes?

apelisse: yes

apelisse:
```
value/value.go:565:30: ValueInterface literal escapes to heap
value/value.go:565:30:  from iter(vk, ValueInterface literal) (parameter to indirect call) at value/value.go:565:11
```

lavalamp: Here's an idea, probably worth what you paid for it:

```
type mapValue struct {
  m map[string]interface{}
  k string
}

...

if !iter(vk, mapValue{m, vk}) {
...
```

apelisse: but the mapValue is still going to escape?

lavalamp: mapValue is not a pointer though...

apelisse: `ValueInterface{}` is not a pointer either?

lavalamp: oh wait, it gets moved to the heap when stuffed in an interface? :(

I guess that makes sense.

apelisse: Yeah ...

lavalamp: If that's the problem, we can allocate a single one and use it for everything?

apelisse: I would have to think of the consequences of doing so? Are we sure nothing is going to keep a copy of it somewhere?

lavalamp: Hopefully that would break a test...

apelisse: I'm pretty sure it would, we can try this out

apelisse:
Replaced by #129

PR #123, opened 2019-11-04
WIP - Add string table generation command

jpbetz:
This is a simple script that consumes an OpenAPI spec and produces a string table for use with server side apply fieldset compression.

I needed to locally patch kubernetes with https://github.com/kubernetes/kube-openapi/pull/176 to add the enums to the OpenAPI spec.

This includes the OpenAPI spec the string table was generated from.


jpbetz:
cc @jennybuckley @apelisse

stringtable/genstringtable.go
```diff
@@ -0,0 +1,207 @@
+/*
```
apelisse: Commenting on the name of the file:

We have an `internal/cli` directory where this might make a lot of sense.

stringtable/genstringtable.go
```diff
@@ -0,0 +1,207 @@
...
+	jsoniter "github.com/json-iterator/go"
+	flag "github.com/spf13/pflag"
+	"gopkg.in/yaml.v2"
+
+	openapi_v2 "github.com/googleapis/gnostic/OpenAPIv2"
+	"k8s.io/klog"
+)
+
+var (
+	file = flag.String("file", "", "The address of the Kubernetes API server. Overrides any value in kubeconfig. Only required if out-of-cluster.")
```
apelisse: You're doing `os.Open` on file, so I suspect it's not the "address of a kubernetes API server". Opening a file is the way to go though.

PR #124, opened 2019-11-04
Perf/add crd example

apelisse:
I tried to run something on the cluster and this specific CRD just timed-out and is VERY slow to even do an update to. So I think it's a good idea to have it in the benchmark. It's pretty slow to do an update and to parse, but we can also see that the size of the set is pretty big (15k) so that's an interesting use-case to investigate.


internal/testdata/podset.json
```diff
@@ -1 +1 @@
...
....":{}},".":{}},"f:terminationMessagePath":{},"f:terminationMessagePolicy":{},"f:volumeMounts":{"k:{\"mountPath\":\"/var/run/secrets/kubernetes.io/serviceaccount\"}":{"f:mountPath":{},"f:name":{},"f:readOnly":{},".":{}},".":{}},".":{}},".":{}},"f:dnsPolicy":{},"f:nodeName":{},"f:priority":{},"f:restartPolicy":{},"f:schedulerName":{},"f:securityContext":{},"f:serviceAccount":{},"f:serviceAccountName":{},"f:terminationGracePeriodSeconds":{},"f:tolerations":{},"f:volumes":{"k:{\"name\":\"default-token-hu5jz\"}":{"f:name":{},"f:secret":{"f:defaultMode":{},"f:secretName":{},".":{}},".":{}},".":{}},".":{}},"f:status":{"f:conditions":{"k:{\"type\":\"ContainersReady\"}":{"f:lastProbeTime":{},"f:lastTransitionTime":{},"f:status":{},"f:type":{},".":{}},"k:{\"type\":\"Initialized\"}":{"f:lastProbeTime":{},"f:lastTransitionTime":{},"f:status":{},"f:type":{},".":{}},"k:{\"type\":\"PodScheduled\"}":{"f:lastProbeTime":{},"f:lastTransitionTime":{},"f:status":{},"f:type":{},".":{}},"k:{\"type\":\"Ready\"}":{"f:lastProbeTime":{},"f:lastTransitionTime":{},"f:status":{},"f:type":{},".":{}},".":{}},"f:containerStatuses":{},"f:hostIP":{},"f:phase":{},"f:podIP":{},"f:qosClass":{},"f:startTime":{},".":{}}}
```
apelisse: For people who don't want to hurt their eyes, the difference is because we've updated the k8s-schema above (so that we have access to CRD types ...), and the `Pod.Spec.Volumes` went from atomic to associative on the name, hence the difference in the set.

internal/testdata/prometheus-crdset.json
```diff
@@ -0,0 +1 @@
...{},".":{}},"f:endpoint":{"f:description":{},"f:type":{},".":{}},"f:insecure":{"f:description":{},"f:type":{},".":{}},"f:secretKey":{"f:description":{},"f:properties":{"f:key":{"f:description":{},"f:type":{},".":{}},"f:name":{"f:description":{},"f:type":{},".":{}},"f:optional":{"f:description":{},"f:type":{},".":{}},".":{}},"f:required":{},".":{}},"f:signatureVersion2":{"f:description":{},"f:type":{},".":{}},".":{}},".":{}},"f:sha":{"f:description":{},"f:type":{},".":{}},"f:tag":{"f:description":{},"f:type":{},".":{}},"f:version":{"f:description":{},"f:type":{},".":{}},".":{}},".":{}},"f:tolerations":{"f:description":{},"f:items":{},"f:type":{},".":{}},"f:version":{"f:description":{},"f:type":{},".":{}},".":{}},".":{}},"f:status":{"f:description":{},"f:properties":{"f:availableReplicas":{"f:description":{},"f:format":{},"f:type":{},".":{}},"f:paused":{"f:description":{},"f:type":{},".":{}},"f:replicas":{"f:description":{},"f:format":{},"f:type":{},".":{}},"f:unavailableReplicas":{"f:description":{},"f:format":{},"f:type":{},".":{}},"f:updatedReplicas":{"f:description":{},"f:format":{},"f:type":{},".":{}},".":{}},"f:required":{},".":{}},".":{}},".":{}},".":{}},"f:version":{},".":{}}}
```
apelisse: This is the large thing, probably most of the problem.

PR #127, opened 2019-11-08
typed: Round-trip to/from unstructured in benchmark

apelisse:
We're only benchmarking "FromUnstructured" but there is no reason that "ToUnstructured" would have the exact same results.

### DIFF
--- a/typed/parser_test.go
+++ b/typed/parser_test.go
@@ -42,7 +42,7 @@ func lastPart(s string) string {
 	return s[strings.LastIndex(s, ".")+1:]
 }
 
-func BenchmarkFromUnstructured(b *testing.B) {
+func BenchmarkConvertUnstructured(b *testing.B) {
 	tests := []struct {
 		typename string
 		obj      []byte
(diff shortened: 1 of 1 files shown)


typed/parser_test.go
```diff
@@ -86,8 +86,10 @@ func BenchmarkFromUnstructured(b *testing.B) {
 			b.ReportAllocs()
 			b.ResetTimer()
 			for n := 0; n < b.N; n++ {
-				if _, err := pt.FromUnstructured(obj); err != nil {
+				if u, err := pt.FromUnstructured(obj); err != nil {
 					b.Fatal(err)
+				} else {
+					_ = u.AsValue().ToUnstructured(false)
```
lavalamp: Why not add a second b.Run(...) instead of making the existing one less clear?

Also, should check the error and b.Fatal if non-nil?

apelisse: `ToUnstructure` can't fail. Yeah, I'll make another test ...

apelisse: Updated.

PR #129, opened 2019-11-22
Perf/new value rebased

apelisse:
Rebased and somehow cleaned-up some of the general interface PR. Still a lot to clean-up I think.

Benchmark results:
```
name                                                  old time/op    new time/op    delta
pkg:sigs.k8s.io/structured-merge-diff/merge goos:linux goarch:amd64
DeducedSimple-12                                         117µs ± 3%     121µs ± 3%      +4.13%  (p=0.016 n=5+5)
DeducedNested-12                                         408µs ± 2%     390µs ± 4%      -4.37%  (p=0.008 n=5+5)
DeducedNestedAcrossVersion-12                            435µs ± 2%     439µs ± 2%        ~     (p=0.421 n=5+5)
LeafConflictAcrossVersion-12                             105µs ± 4%     115µs ± 2%     +10.15%  (p=0.008 n=5+5)
MultipleApplierRecursiveRealConversion-12               1.69ms ± 1%    1.89ms ± 2%     +11.70%  (p=0.008 n=5+5)
Operations/Pod/Create-12                                 110µs ± 4%     159µs ± 3%     +44.25%  (p=0.008 n=5+5)
Operations/Pod/Apply-12                                  306µs ± 4%     392µs ± 2%     +27.90%  (p=0.008 n=5+5)
Operations/Pod/Update-12                                 189µs ± 2%     337µs ± 3%     +78.32%  (p=0.008 n=5+5)
Operations/Pod/UpdateVersion-12                          262µs ± 3%     510µs ± 3%     +94.66%  (p=0.008 n=5+5)
Operations/Node/Create-12                                160µs ± 1%     229µs ± 2%     +43.51%  (p=0.008 n=5+5)
Operations/Node/Apply-12                                 513µs ± 4%     572µs ± 2%     +11.47%  (p=0.008 n=5+5)
Operations/Node/Update-12                                285µs ± 1%     531µs ± 1%     +86.46%  (p=0.008 n=5+5)
Operations/Node/UpdateVersion-12                         389µs ± 4%     826µs ± 3%    +112.40%  (p=0.008 n=5+5)
Operations/Endpoints/Create-12                          12.2µs ± 3%    18.9µs ± 2%     +54.47%  (p=0.008 n=5+5)
Operations/Endpoints/Apply-12                            907µs ± 1%      35µs ± 3%     -96.16%  (p=0.008 n=5+5)
Operations/Endpoints/Update-12                           290µs ± 2%    1915µs ± 1%    +560.98%  (p=0.008 n=5+5)
Operations/Endpoints/UpdateVersion-12                    556µs ± 1%    3634µs ± 1%    +553.34%  (p=0.008 n=5+5)
Operations/CustomResourceDefinition/Create-12           1.36ms ± 1%    1.73ms ± 2%     +27.47%  (p=0.008 n=5+5)
Operations/CustomResourceDefinition/Apply-12            3.42ms ± 3%    4.26ms ± 2%     +24.69%  (p=0.008 n=5+5)
Operations/CustomResourceDefinition/Update-12           1.81ms ± 2%    3.31ms ± 3%     +82.74%  (p=0.008 n=5+5)
Operations/CustomResourceDefinition/UpdateVersion-12    2.30ms ± 1%    4.76ms ± 2%    +106.83%  (p=0.008 n=5+5)
pkg:sigs.k8s.io/structured-merge-diff/typed goos:linux goarch:amd64
ConvertUnstructured/Pod/From-12                          125µs ± 3%      50µs ± 2%     -60.04%  (p=0.008 n=5+5)
ConvertUnstructured/Pod/To-12                           31.3µs ± 4%     0.0µs ± 9%    -100.00%  (p=0.008 n=5+5)
ConvertUnstructured/Node/From-12                         260µs ± 2%     107µs ± 1%     -58.87%  (p=0.008 n=5+5)
ConvertUnstructured/Node/To-12                          68.5µs ± 3%     0.0µs ± 2%    -100.00%  (p=0.008 n=5+5)
ConvertUnstructured/Endpoints/From-12                   5.58ms ± 3%    2.25ms ± 3%     -59.58%  (p=0.008 n=5+5)
ConvertUnstructured/Endpoints/To-12                     1.46ms ± 1%    0.00ms ± 6%    -100.00%  (p=0.008 n=5+5)
ConvertUnstructured/CustomResourceDefinition/From-12    1.97ms ± 3%    0.78ms ± 2%     -60.36%  (p=0.008 n=5+5)
ConvertUnstructured/CustomResourceDefinition/To-12       613µs ± 2%       0µs ± 2%    -100.00%  (p=0.008 n=5+5)

name                                                  old alloc/op   new alloc/op   delta
pkg:sigs.k8s.io/structured-merge-diff/merge goos:linux goarch:amd64
DeducedSimple-12                                        33.8kB ± 0%    28.2kB ± 0%     -16.52%  (p=0.008 n=5+5)
DeducedNested-12                                         103kB ± 0%      87kB ± 0%     -15.42%  (p=0.008 n=5+5)
DeducedNestedAcrossVersion-12                            107kB ± 0%      96kB ± 0%     -10.30%  (p=0.008 n=5+5)
LeafConflictAcrossVersion-12                            31.2kB ± 0%    28.2kB ± 0%      -9.53%  (p=0.016 n=5+4)
MultipleApplierRecursiveRealConversion-12                509kB ± 0%     513kB ± 0%      +0.89%  (p=0.008 n=5+5)
Operations/Pod/Create-12                                17.6kB ± 0%    22.6kB ± 0%     +28.35%  (p=0.008 n=5+5)
Operations/Pod/Apply-12                                 58.2kB ± 0%    57.8kB ± 0%      -0.62%  (p=0.000 n=4+5)
Operations/Pod/Update-12                                26.5kB ± 0%    46.4kB ± 0%     +75.07%  (p=0.008 n=5+5)
Operations/Pod/UpdateVersion-12                         34.4kB ± 0%    69.2kB ± 0%    +101.18%  (p=0.008 n=5+5)
Operations/Node/Create-12                               26.4kB ± 0%    33.2kB ± 0%     +25.88%  (p=0.008 n=5+5)
Operations/Node/Apply-12                                98.1kB ± 0%    83.0kB ± 0%     -15.42%  (p=0.008 n=5+5)
Operations/Node/Update-12                               37.7kB ± 0%    71.0kB ± 0%     +88.18%  (p=0.008 n=5+5)
Operations/Node/UpdateVersion-12                        48.1kB ± 0%   107.9kB ± 0%    +124.16%  (p=0.008 n=5+5)
Operations/Endpoints/Create-12                          2.86kB ± 0%    3.63kB ± 0%     +26.94%  (p=0.016 n=5+4)
Operations/Endpoints/Apply-12                           46.9kB ± 0%     6.9kB ± 0%     -85.21%  (p=0.016 n=5+4)
Operations/Endpoints/Update-12                          4.47kB ± 0%  263.77kB ± 0%   +5806.21%  (p=0.008 n=5+5)
Operations/Endpoints/UpdateVersion-12                   5.07kB ± 0%  522.83kB ± 0%  +10221.50%  (p=0.008 n=5+5)
Operations/CustomResourceDefinition/Create-12            118kB ± 0%     183kB ± 0%     +54.96%  (p=0.008 n=5+5)
Operations/CustomResourceDefinition/Apply-12             409kB ± 0%     520kB ± 0%     +27.03%  (p=0.008 n=5+5)
Operations/CustomResourceDefinition/Update-12            139kB ± 0%     402kB ± 0%    +189.47%  (p=0.008 n=5+5)
Operations/CustomResourceDefinition/UpdateVersion-12     159kB ± 0%     621kB ± 0%    +290.37%  (p=0.008 n=5+5)
pkg:sigs.k8s.io/structured-merge-diff/typed goos:linux goarch:amd64
ConvertUnstructured/Pod/From-12                         35.6kB ± 0%     6.4kB ± 0%     -81.99%  (p=0.008 n=5+5)
ConvertUnstructured/Pod/To-12                           14.4kB ± 0%     0.0kB         -100.00%  (p=0.008 n=5+5)
ConvertUnstructured/Node/From-12                        75.8kB ± 0%    14.7kB ± 0%     -80.60%  (p=0.008 n=5+5)
ConvertUnstructured/Node/To-12                          28.8kB ± 0%     0.0kB             ~     (p=0.079 n=4+5)
ConvertUnstructured/Endpoints/From-12                   2.22MB ± 0%    0.29MB ± 0%     -86.97%  (p=0.008 n=5+5)
ConvertUnstructured/Endpoints/To-12                      803kB ± 0%       0kB         -100.00%  (p=0.008 n=5+5)
ConvertUnstructured/CustomResourceDefinition/From-12     596kB ± 0%      93kB ± 0%     -84.40%  (p=0.016 n=5+4)
ConvertUnstructured/CustomResourceDefinition/To-12       307kB ± 0%       0kB         -100.00%  (p=0.008 n=5+5)
```

This is noticeably slower on some operations, faster on some others:
Still, if we look at a specifically bad case like Endpoint/Update, Conversion + Update gives:
5.58ms + 290µs + 1.46ms = 7.33ms vs 2.25ms + 1915µs + 0ms = 4.16ms. 
So that's an improvement, and we can still remove one level of conversion.


value/map.go
```diff
@@ -0,0 +1,174 @@
...
+	for k, v := range m {
+		ks, ok := k.(string)
+		if !ok {
+			return false
+		}
+		vo, ok := other.Get(ks)
+		if !ok {
+			return false
+		}
+		if !Equals(ValueInterface{Value: v}, vo) {
```
apelisse: About 41% of the allocs on "Endpoints/Update" here.

> `value/map.go:125:28: composite literal escapes to heap`

value/map.go
```diff
@@ -0,0 +1,174 @@
...
+
+func (m MapInterface) Set(key string, val Value) {
+	m[key] = val.Interface()
+}
+
+func (m MapInterface) Get(key string) (Value, bool) {
+	if v, ok := m[key]; !ok {
+		return nil, false
+	} else {
+		return ValueInterface{Value: v}, true
```
apelisse: About 41% of the allocations on `Endpoints/Update`:
> `value/map.go:90:24: composite literal escapes to heap`

typed/validate.go
```diff
@@ -185,35 +156,25 @@ func (v *validatingObjectWalker) doList(t *schema.List) (errs ValidationErrors)
...
-			errs = append(errs, v2.validate()...)
-			v.finishDescent(v2)
-		} else {
-			v2 := v.prepareDescent(pe, t.ElementType)
-			v2.value = item.Value
-			errs = append(errs, v2.validate()...)
-			v2.doNode()
-			v.finishDescent(v2)
+func (v *validatingObjectWalker) visitMapItems(t *schema.Map, m value.Map) (errs ValidationErrors) {
+	m.Iterate(func(key string, val value.Value) bool {
```
apelisse: We have a lot of allocations because of the lambda escaping:
> `typed/validate.go:160:12: func literal escapes to heap`
Maybe about 36% of 'ConvertUnstructured/Endpoints/From`

value/map.go
```diff
@@ -0,0 +1,174 @@
...
+func (m MapInterface) Delete(key string) {
+	delete(m, key)
+}
+
+func (m MapInterface) Iterate(fn func(key string, value Value) bool) bool {
+	for k, v := range m {
+		if ks, ok := k.(string); !ok {
+			continue
+		} else {
+			if !fn(ks, ValueInterface{Value: v}) {
```
apelisse: 56% of allocations for `ConvertUnstructured/Endpoints/From` are because of the escaping literal.

jennybuckley:
The improvement from this PR would mostly be in the k8s benchmarks, right? (since we would be able to stop allocating a Value struct?)

value/map.go
```diff
@@ -0,0 +1,174 @@
...
+		return nil, false
+	} else {
+		return ValueInterface{Value: v}, true
+	}
+}
+
+func (m MapInterface) Delete(key string) {
+	delete(m, key)
+}
+
```
lavalamp: Idea:

```
var viPool sync.Pool

func rentVI() *ValueInterface {
  return viPool.Get().(*ValueInterface)
}

func returnV(v Value) {
  if vi, ok := v.(ValueInterface); ok {
    viPool.Put(&vi)
  }
}
```

value/map.go
lavalamp: Idea:

```
  pvi := rentVI()
  pvi.Value = v
  vv := Value(*pvi)
  if !fn(ks, vv) { ... }
  returnV(vv)
```

value/map.go
```diff
@@ -0,0 +1,174 @@
...
+	}
+}
+
+type MapInterface map[interface{}]interface{}
+
+func (m MapInterface) Set(key string, val Value) {
+	m[key] = val.Interface()
+}
+
+func (m MapInterface) Get(key string) (Value, bool) {
```
lavalamp: Idea:

```
func (m MapInterface) GetInto(key string, destination Value) (Value, bool) {
  destVI, _ := destination.(ValueInterface)
  if v, ok := m[key]; !ok {
    return nil, false
  } else {
    destVI.Value = v
    return destVI, true
  }
}
```

value/map.go
lavalamp: Same idea as previous one.

value/map.go
lavalamp: (If this works out it can be further improved by threading a context through everything so that there's no lock contention)

typed/validate.go
lavalamp: ah I missed this one. I think you can do this without making a closure:

```
func (v *validatingObjectWalker) visitMapItems(t *schema.Map, m value.Map) (errs ValidationErrors) {
  v.mapiterParent = t
  v.mapiterParent = m
  v.mapiterErrs = v.mapiterErrs[:0]
  m.Iterate(v.mapiter)
  return v.mapiterErrs
}

func (v *validatingObjectWalker) mapiter(key string, val value.Value) {
  // body of iteration function here
}
```

typed/helpers.go
```diff
@@ -116,12 +97,12 @@ func resolveSchema(s *schema.Schema, tr schema.TypeRef, v *value.Value, ah atomH
 
 func deduceAtom(a schema.Atom, v *value.Value) schema.Atom {
```
lavalamp: If value.Value is an interface (implying it can be nil) we probably should not pass a pointer to it, that's unnecessary and unusual.

apelisse: I think in this case, `nil` is different from not present

fieldpath/pathelementmap_test.go
```diff
@@ -29,22 +29,22 @@ func TestPathElementValueMap(t *testing.T) {
 		t.Fatal("Unexpected path-element found in empty map")
 	}
 
-	m.Insert(PathElement{FieldName: strptr("carrot")}, value.StringValue("knife"))
-	m.Insert(PathElement{FieldName: strptr("chive")}, value.IntValue(2))
+	m.Insert(PathElement{FieldName: strptr("carrot")}, value.ValueInterface{Value: "knife"})
```
lavalamp: Can we get back the "StringValue()" family of functions? This change is unnecessary.

fieldpath/pathelementmap_test.go
```diff
@@ -29,22 +29,22 @@ func TestPathElementValueMap(t *testing.T) {
...
+	m.Insert(PathElement{FieldName: strptr("chive")}, value.ValueInterface{Value: 2})
 
 	if _, ok := m.Get(PathElement{FieldName: strptr("onion")}); ok {
 		t.Fatal("Unexpected path-element in map")
 	}
 
 	if val, ok := m.Get(PathElement{FieldName: strptr("carrot")}); !ok {
 		t.Fatal("Missing path-element in map")
-	} else if !val.Equals(value.StringValue("knife")) {
+	} else if !value.Equals(val, value.ValueInterface{Value: "knife"}) {
```
lavalamp: hm, I expected the interface to let you also not change this line.

apelisse: Yeah, I'm not sure yet ...

apelisse:
Updated the interface and a bit of clean-up and added documentation. I'd love to merge that since it's insanely hard to rebase.

apelisse:
New benchmarks stats:
```
name                                                  old time/op    new time/op    delta
pkg:sigs.k8s.io/structured-merge-diff/typed goos:linux goarch:amd64
ConvertUnstructured/Pod/From-12                          125µs ± 2%      51µs ± 3%   -58.89%  (p=0.000 n=9+10)
ConvertUnstructured/Pod/To-12                           31.3µs ± 4%     0.0µs ± 5%  -100.00%  (p=0.000 n=10+10)
ConvertUnstructured/Node/From-12                         257µs ± 4%     118µs ± 3%   -54.19%  (p=0.000 n=9+9)
ConvertUnstructured/Node/To-12                          69.3µs ± 4%     0.0µs ± 3%  -100.00%  (p=0.000 n=10+7)
ConvertUnstructured/Endpoints/From-12                   5.46ms ± 2%    2.03ms ± 3%   -62.89%  (p=0.000 n=10+9)
ConvertUnstructured/Endpoints/To-12                     1.43ms ± 4%    0.00ms ± 7%  -100.00%  (p=0.000 n=10+10)
ConvertUnstructured/CustomResourceDefinition/From-12    1.89ms ± 5%    0.70ms ± 2%   -62.76%  (p=0.000 n=9+8)
ConvertUnstructured/CustomResourceDefinition/To-12       598µs ± 3%       0µs ± 5%  -100.00%  (p=0.000 n=10+10)
pkg:sigs.k8s.io/structured-merge-diff/merge goos:linux goarch:amd64
DeducedSimple-12                                         115µs ± 6%     121µs ± 3%    +5.09%  (p=0.000 n=10+10)
DeducedNested-12                                         408µs ± 4%     384µs ± 5%    -5.97%  (p=0.000 n=10+9)
DeducedNestedAcrossVersion-12                            429µs ± 4%     424µs ± 3%      ~     (p=0.315 n=9+10)
LeafConflictAcrossVersion-12                             103µs ± 3%     114µs ± 4%   +10.00%  (p=0.000 n=10+10)
MultipleApplierRecursiveRealConversion-12               1.71ms ± 3%    1.89ms ± 5%   +10.28%  (p=0.000 n=10+10)
Operations/Pod/Create-12                                 110µs ± 3%     157µs ± 5%   +42.33%  (p=0.000 n=10+10)
Operations/Pod/Apply-12                                  313µs ± 7%     387µs ± 5%   +23.83%  (p=0.000 n=10+10)
Operations/Pod/Update-12                                 190µs ± 2%     326µs ± 4%   +71.97%  (p=0.000 n=10+10)
Operations/Pod/UpdateVersion-12                          260µs ± 3%     489µs ± 3%   +88.04%  (p=0.000 n=10+10)
Operations/Node/Create-12                                159µs ± 2%     228µs ± 4%   +43.60%  (p=0.000 n=9+10)
Operations/Node/Apply-12                                 510µs ± 3%     565µs ± 2%   +10.76%  (p=0.000 n=8+10)
Operations/Node/Update-12                                280µs ± 4%     500µs ± 2%   +78.32%  (p=0.000 n=10+10)
Operations/Node/UpdateVersion-12                         388µs ± 4%     757µs ± 2%   +94.95%  (p=0.000 n=10+10)
Operations/Endpoints/Create-12                          12.2µs ± 3%    18.5µs ± 4%   +52.39%  (p=0.000 n=10+10)
Operations/Endpoints/Apply-12                            908µs ± 3%      34µs ± 4%   -96.22%  (p=0.000 n=10+10)
Operations/Endpoints/Update-12                           288µs ± 2%    1338µs ± 2%  +364.24%  (p=0.000 n=10+9)
Operations/Endpoints/UpdateVersion-12                    559µs ± 2%    2665µs ± 2%  +376.57%  (p=0.000 n=10+10)
Operations/CustomResourceDefinition/Create-12           1.37ms ± 2%    1.68ms ± 2%   +22.68%  (p=0.000 n=9+9)
Operations/CustomResourceDefinition/Apply-12            3.44ms ± 4%    4.12ms ± 3%   +19.67%  (p=0.000 n=10+9)
Operations/CustomResourceDefinition/Update-12           1.82ms ± 3%    3.09ms ± 2%   +70.08%  (p=0.000 n=10+9)
Operations/CustomResourceDefinition/UpdateVersion-12    2.30ms ± 2%    4.39ms ± 2%   +90.42%  (p=0.000 n=9+10)

name                                                  old alloc/op   new alloc/op   delta
pkg:sigs.k8s.io/structured-merge-diff/typed goos:linux goarch:amd64
ConvertUnstructured/Pod/From-12                         35.6kB ± 0%     4.6kB ± 0%   -87.01%  (p=0.000 n=10+10)
ConvertUnstructured/Pod/To-12                           14.4kB ± 0%     0.0kB       -100.00%  (p=0.000 n=7+10)
ConvertUnstructured/Node/From-12                        75.8kB ± 0%    11.5kB ± 0%   -84.88%  (p=0.000 n=10+10)
ConvertUnstructured/Node/To-12                          28.8kB ± 0%     0.0kB       -100.00%  (p=0.000 n=10+10)
ConvertUnstructured/Endpoints/From-12                   2.22MB ± 0%    0.18MB ± 0%   -92.02%  (p=0.000 n=10+10)
ConvertUnstructured/Endpoints/To-12                      803kB ± 0%       0kB       -100.00%  (p=0.000 n=10+10)
ConvertUnstructured/CustomResourceDefinition/From-12     596kB ± 0%      61kB ± 0%   -89.81%  (p=0.000 n=10+10)
ConvertUnstructured/CustomResourceDefinition/To-12       307kB ± 0%       0kB       -100.00%  (p=0.000 n=10+10)
pkg:sigs.k8s.io/structured-merge-diff/merge goos:linux goarch:amd64
DeducedSimple-12                                        33.8kB ± 0%    26.5kB ± 0%   -21.46%  (p=0.000 n=10+10)
DeducedNested-12                                         103kB ± 0%      83kB ± 0%   -19.91%  (p=0.000 n=10+10)
DeducedNestedAcrossVersion-12                            107kB ± 0%      89kB ± 0%   -16.10%  (p=0.000 n=10+9)
LeafConflictAcrossVersion-12                            31.2kB ± 0%    26.0kB ± 0%   -16.48%  (p=0.000 n=10+10)
MultipleApplierRecursiveRealConversion-12                509kB ± 0%     507kB ± 0%    -0.35%  (p=0.000 n=10+10)
Operations/Pod/Create-12                                17.6kB ± 0%    21.2kB ± 0%   +20.44%  (p=0.000 n=10+7)
Operations/Pod/Apply-12                                 58.2kB ± 0%    53.6kB ± 0%    -7.83%  (p=0.000 n=10+10)
Operations/Pod/Update-12                                26.5kB ± 0%    38.3kB ± 0%   +44.28%  (p=0.000 n=10+10)
Operations/Pod/UpdateVersion-12                         34.4kB ± 0%    54.3kB ± 0%   +57.77%  (p=0.000 n=10+10)
Operations/Node/Create-12                               26.4kB ± 0%    30.9kB ± 0%   +17.40%  (p=0.000 n=10+10)
Operations/Node/Apply-12                                98.1kB ± 0%    76.3kB ± 0%   -22.27%  (p=0.000 n=9+10)
Operations/Node/Update-12                               37.7kB ± 0%    55.0kB ± 0%   +45.75%  (p=0.000 n=10+10)
Operations/Node/UpdateVersion-12                        48.1kB ± 0%    78.1kB ± 0%   +62.23%  (p=0.000 n=10+10)
Operations/Endpoints/Create-12                          2.86kB ± 0%    3.44kB ± 0%   +20.29%  (p=0.000 n=10+8)
Operations/Endpoints/Apply-12                           46.9kB ± 0%     6.4kB ± 0%   -86.44%  (p=0.000 n=10+9)
Operations/Endpoints/Update-12                          4.47kB ± 0%    6.44kB ± 0%   +44.26%  (p=0.000 n=10+10)
Operations/Endpoints/UpdateVersion-12                   5.07kB ± 0%    8.44kB ± 0%   +66.56%  (p=0.000 n=10+10)
Operations/CustomResourceDefinition/Create-12            118kB ± 0%     170kB ± 0%   +43.89%  (p=0.000 n=10+10)
Operations/CustomResourceDefinition/Apply-12             409kB ± 0%     480kB ± 0%   +17.42%  (p=0.000 n=9+10)
Operations/CustomResourceDefinition/Update-12            139kB ± 0%     294kB ± 0%  +111.79%  (p=0.000 n=10+10)
Operations/CustomResourceDefinition/UpdateVersion-12     159kB ± 0%     418kB ± 0%  +162.71%  (p=0.000 n=6+10)
```

apelisse:
Recomputing the statement I made before:
Still, if we look at a specifically bad case like Endpoint/Update, Conversion + Update gives:
5.46ms + 288µs + 1.43ms = 7.17ms vs 2.03ms + 1338µs + 0ms = 3.36ms (-0.8ms compared to initial test).

apelisse:
For the custom resource we get:
1.89ms + 1.82ms + 598µs = 4.30ms vs 0.70ms + 3.09ms + 0µs = 3.79ms

lavalamp:
Recall that we think we have a garbage problem, I wouldn't even look at the time numbers, they basically don't correspond to real life at all until we get our garbage under control.

apelisse:
Let me recompute these numbers for number of allocs. They are somewhat correlated so I would expect the same order of magnitude but you're right.

fieldpath/element.go
```diff
@@ -71,7 +71,7 @@ func (e PathElement) Less(rhs PathElement) bool {
 		if rhs.Value == nil {
 			return true
 		}
-		return e.Value.Less(*rhs.Value)
+		return value.Less(*e.Value, *rhs.Value)
```
lavalamp: Can we change the interface such that this line doesn't change?

fieldpath/element.go
```diff
@@ -103,12 +103,12 @@ func (e PathElement) String() string {
 	case e.Key != nil:
 		strs := make([]string, len(*e.Key))
 		for i, k := range *e.Key {
-			strs[i] = fmt.Sprintf("%v=%v", k.Name, k.Value)
+			strs[i] = fmt.Sprintf("%v=%v", k.Name, value.ToString(k.Value))
```
lavalamp: Can we change the interface such that this line doesn't change?

fieldpath/fromvalue.go
```diff
@@ -43,37 +43,36 @@ type objectWalker struct {
 
 func (w *objectWalker) walk() {
 	switch {
-	case w.value.Null:
```
lavalamp: Did this (Null) get lost?

apelisse:
For endpoints, we get: 19,067+10,038+77=29,182 vs 5,023+174=5,197
For CR we get: 596,198+307,127+139,045=1,042,370 vs 29,4481+60,748=355,229

fieldpath/fromvalue.go
```diff
@@ -43,37 +43,36 @@ type objectWalker struct {
...
 		// statement).
 
 	// Descend
-	case w.value.ListValue != nil:
+	case w.value.IsList():
 		// If the list were atomic, we'd break here, but we don't have
 		// a schema, so we can't tell.
 
-		for i, child := range w.value.ListValue.Items {
+		for i := 0; i < w.value.List().Length(); i++ {
```
lavalamp: optional nit: for readability I would probably do `l := w.value.List()`

fieldpath/path.go
```diff
@@ -93,10 +93,10 @@ func MakePath(parts ...interface{}) (Path, error) {
 				return nil, fmt.Errorf("associative list key type path elements must have at least one key (got zero)")
 			}
 			fp = append(fp, PathElement{Key: t})
-		case value.Value:
+		case *value.Value:
```
lavalamp: It doesn't seem good to want a pointer to an interface here. Why do we want that?

fieldpath/path_test.go
```diff
@@ -22,6 +22,13 @@ import (
 	"sigs.k8s.io/structured-merge-diff/value"
 )
 
+var (
+	_V = func(i interface{}) *value.Value {
```
lavalamp: Why is this a variable and not just a function declaration?

fieldpath/path_test.go
```diff
@@ -31,16 +38,16 @@ func TestPathString(t *testing.T) {
 		{"basic1", MakePathOrDie("foo", 1), ".foo[1]"},
 		{"basic2", MakePathOrDie("foo", "bar", 1, "baz"), ".foo.bar[1].baz"},
 		{"associative-list-ref", MakePathOrDie("foo", KeyByFields(
-			"a", value.StringValue("b"),
-			"c", value.IntValue(1),
-			"d", value.FloatValue(1.5),
-			"e", value.BooleanValue(true),
+			"a", "b",
+			"c", 1,
+			"d", 1.5,
```
lavalamp: This reads better now, but it's no longer obvious that we're testing all the types. Maybe a comment would help?

fieldpath/pathelementmap_test.go
```diff
@@ -29,22 +29,22 @@ func TestPathElementValueMap(t *testing.T) {
...
+	m.Insert(PathElement{FieldName: strptr("chive")}, value.NewValueInterface(2))
 
 	if _, ok := m.Get(PathElement{FieldName: strptr("onion")}); ok {
 		t.Fatal("Unexpected path-element in map")
 	}
 
 	if val, ok := m.Get(PathElement{FieldName: strptr("carrot")}); !ok {
 		t.Fatal("Missing path-element in map")
-	} else if !val.Equals(value.StringValue("knife")) {
+	} else if !value.Equals(val, value.NewValueInterface("knife")) {
```
lavalamp: `val.Equals` - no need to change the interface?

fieldpath/serialize-pe.go
```diff
@@ -134,19 +135,20 @@ func serializePathElementToWriter(w io.Writer, pe PathElement) error {
...
 		}
 		stream.WriteObjectStart()
+
 		for i, field := range *pe.Key {
 			if i > 0 {
 				stream.WriteMore()
 			}
 			stream.WriteObjectField(field.Name)
-			field.Value.WriteJSONStream(stream)
+			value.WriteJSONStream(field.Value, stream)
```
lavalamp: Did this call site have to change?

go.mod
```diff
@@ -9,3 +9,5 @@ require (
 	github.com/modern-go/reflect2 v1.0.1 // indirect
 	github.com/stretchr/testify v1.3.0 // indirect
 )
+
+go 1.13
```
lavalamp: Unexpected-- if we're making this the start of v2 shouldn't the top line in this file change?

internal/cli/operation.go
```diff
@@ -113,7 +114,7 @@ func (m merge) Execute(w io.Writer) error {
 		return err
 	}
 
-	yaml, err := out.AsValue().ToYAML()
+	yaml, err := yaml.Marshal(out.AsValue().Interface())
```
lavalamp: Hm, did this line have to change? (sorry for all these comments, but keeping the diff down is also nice to any users...)

merge/conflict_test.go
```diff
@@ -29,24 +29,26 @@ var (
 	_NS  = fieldpath.NewSet
 	_P   = fieldpath.MakePathOrDie
 	_KBF = fieldpath.KeyByFields
-	_SV  = value.StringValue
-	_IV  = value.IntValue
+	_V   = func(i interface{}) *value.Value {
```
lavalamp: ah, the other definition copied this code. I still think it is weird to need a `*value.Value`; if you just needed a `value.Value` then this would just be `_V = value.NewValueInterface`

typed/deduced_test.go
```diff
@@ -201,12 +201,9 @@ func TestMergeDeduced(t *testing.T) {
 			if err != nil {
 				t.Errorf("got validation errors: %v", err)
 			} else {
-				t.Logf("got:\v%v", got.AsValue())
-				gotUS := got.AsValue().ToUnstructured(true)
-				expectUS := out.AsValue().ToUnstructured(true)
-				if !reflect.DeepEqual(gotUS, expectUS) {
+				if !value.Equals(got.AsValue(), out.AsValue()) {
```
lavalamp: Hopefully we can get `if !got.Equals(out) {`

Was the existing reflect.DeepEqual an oversight or deliberate?

typed/helpers.go
```diff
@@ -28,7 +28,7 @@ import (
 
 // ValidationError reports an error about a particular field
 type ValidationError struct {
-	Path         fieldpath.Path
+	Path         string
```
lavalamp: Why? Is this because errors outlive other things? Maybe it's OK to explicitly make a copy for the errors?

typed/helpers.go
lavalamp: Are you using `nil` to mean the same thing as the previous `Null bool`? That explains things. Why not have an explicit IsNull() function? Then you don't need this double indirection?

typed/helpers.go
```diff
@@ -116,12 +97,12 @@ func resolveSchema(s *schema.Schema, tr schema.TypeRef, v *value.Value, ah atomH
 
 func deduceAtom(a schema.Atom, v *value.Value) schema.Atom {
 	switch {
-	case v == nil:
-	case v.FloatValue != nil, v.IntValue != nil, v.StringValue != nil, v.BooleanValue != nil:
+	case v == nil, *v == nil:
```
lavalamp: This function looks to not have handled `v.Null == true` correctly previously. Is dropping down to the `return a` line the right thing to do?

typed/helpers.go
lavalamp: Can you add a comment explaining what this function is doing? It looks like it usually completely ignores the input schema, how can that be correct?

typed/helpers.go
```diff
@@ -56,58 +56,39 @@ func (errs ValidationErrors) Error() string {
...
+	for i := range errs {
+		errs[i].Path = prefix + errs[i].Path
 	}
+	return errs
 }
 
-func (ef errorFormatter) errorf(format string, args ...interface{}) ValidationErrors {
+func errorf(format string, args ...interface{}) ValidationErrors {
 	return ValidationErrors{{
-		Path:         append(fieldpath.Path{}, ef.path...),
```
lavalamp: I don't understand this change, it looks like errors no longer carry the exact path information with them?

typed/helpers.go
```diff
@@ -142,80 +123,50 @@ func handleAtom(a schema.Atom, tr schema.TypeRef, ah atomHandler) ValidationErro
...
-	return nil
+	return errorf("schema error: invalid atom: %v", name)
 }
 
 // Returns the list, or an error. Reminder: nil is a valid list and might be returned.
-func listValue(val value.Value) (*value.List, error) {
-	switch {
-	case val.Null:
+func listValue(val value.Value) (value.List, error) {
+	if val.IsNull() {
```
lavalamp: Wait, you already have an explicit IsNull()? I really don't understand your comment on deduceAtom now?

typed/helpers.go
```diff
@@ -142,80 +123,50 @@ func handleAtom(a schema.Atom, tr schema.TypeRef, ah atomHandler) ValidationErro
...
 	}
+	return val.List(), nil
 }
 
 // Returns the map, or an error. Reminder: nil is a valid map and might be returned.
-func mapValue(val value.Value) (*value.Map, error) {
-	switch {
-	case val.Null:
+func mapValue(val value.Value) (value.Map, error) {
+	if val == nil || val.IsNull() {
```
lavalamp: I think val == nil should be an error caught earlier. We should make the rule that Null objects are represented with `val.IsNull() == true`.

typed/helpers.go
```diff
@@ -142,80 +123,50 @@ func handleAtom(a schema.Atom, tr schema.TypeRef, ah atomHandler) ValidationErro
...
+	if child.IsNull() {
 		// For now, the keys are required which means that null entries
 		// are illegal.
 		return pe, errors.New("associative list with keys may not have a null element")
 	}
-	if child.MapValue == nil {
+	if !child.IsMap() {
 		return pe, errors.New("associative list with keys may not have non-map elements")
 	}
 	keyMap := value.FieldList{}
```
lavalamp: pre-sizing this slice might save an alloc or two. Also if we do any of the things we were talking about to value.Value, we could consider caching this once it's computed? (no action needed, just ideas for followup)

fieldpath/path_test.go
apelisse: Yeah, these are residues from when I've made the value as a strict `interface{}`

go.mod
apelisse: Go keeps inserting this in the file ... and so I keep committing it by accident, lovely ...

typed/helpers.go
apelisse: Yes, the problem is that: 
1. Ideally we don't have a lot of errors, technically we're not supposed to have any when we're doing Updates
2. This forces the escapes of numbers, strings, Values, everything on every single field of objects, for ... a non-nominal use-case.

apelisse: And eventually we just convert everything to a string anyway, so we don't really need the PE

internal/cli/operation.go
```diff
@@ -113,7 +114,7 @@ func (m merge) Execute(w io.Writer) error {
 		return err
 	}
 
-	yaml, err := out.AsValue().ToYAML()
+	yaml, err := yaml.Marshal(out.AsValue().Unstructured())
```
lavalamp: minor preference for changing to: value.ToYAML(out.AsValue())

typed/deduced_test.go
lavalamp: did you add Equals() to the interface?

typed/helpers.go
lavalamp: But we shouldn't actually construct one of these unless there actually is an error? Why would this cause stuff to escape all the time? It shouldn't be a problem if it escapes only when there's an error?

I guess I'd have expected a []string for the replacement?

typed/helpers.go
lavalamp: I still don't think I understand this

typed/helpers.go
lavalamp: is the == nil check still needed? I think the equivalent condition previously returned an error?

typed/helpers.go
jennybuckley: For `v.Null == true`, we intentionally dropped down into the `return a` line, since lists and maps can both be null, and if it's an associative list for example, then we didn't want someone applying null to delete all the entries in the list.

typed/merge.go
```diff
@@ -67,11 +71,11 @@ var (
 func (w *mergingWalker) merge() (errs ValidationErrors) {
 	if w.lhs == nil && w.rhs == nil {
 		// check this condidition here instead of everywhere below.
-		return w.errorf("at least one of lhs and rhs must be provided")
+		return errorf("at least one of lhs and rhs must be provided")
```
lavalamp: How can a caller know the path that this error came from?

jennybuckley: I think this error would only happen at root anyway, since I can't think of a way that we would create a mergingWalker with no lhs or rhs sides from descending.

typed/merge.go
```diff
@@ -140,81 +144,89 @@ func (w *mergingWalker) prepareDescent(pe fieldpath.PathElement, tr schema.TypeR
...
 			w2 := w.prepareDescent(pe, t.ElementType)
-			w2.lhs = &child
+			w2.lhs = value.Value(child)
 			if rchild, ok := observedRHS.Get(pe); ok {
-				w2.rhs = &rchild
+				w2.rhs = rchild
 			}
 			if newErrs := w2.merge(); len(newErrs) > 0 {
-				errs = append(errs, newErrs...)
+				errs = append(errs, newErrs.WithPrefix(pe.String())...)
```
lavalamp: Why is a WithPrefix merited here but not immediately above on line 217?

typed/merge.go
```diff
@@ -140,81 +144,89 @@ func (w *mergingWalker) prepareDescent(pe fieldpath.PathElement, tr schema.TypeR
...
 	// taking dest as input so that it can be called as a one-liner with
 	// append.
 	if v == nil {
 		return nil
 	}
-	m, err := mapValue(*v)
+	m, err := mapValue(v)
 	if err != nil {
-		return w.prefixError(prefix, err)
+		return errorf("%v: %v", prefix, err)
```
lavalamp: This substantially changes the way the path is reported in the error, no? Is this way of reporting the path consistent with what WithPrefix does?

typed/helpers.go
jennybuckley: Whether that's the desired behavior is maybe up for debate. For a concrete example, the way v.Null is treated now (in master) means that these two operations are allowed without conflicts, and if we changed v.Null == true to act like a scalar, then the second operation here fails with a conflict on ".a":
https://github.com/kubernetes-sigs/structured-merge-diff/blob/e9173f465414834d9f259ca30766989a2183a390/merge/multiple_appliers_test.go#L712-L730

typed/helpers.go
```diff
@@ -56,58 +56,39 @@ func (errs ValidationErrors) Error() string {
...
+// when unwinding the stack on errors.
+func (errs ValidationErrors) WithPrefix(prefix string) ValidationErrors {
+	for i := range errs {
+		errs[i].Path = prefix + errs[i].Path
 	}
+	return errs
 }
 
-func (ef errorFormatter) errorf(format string, args ...interface{}) ValidationErrors {
+func errorf(format string, args ...interface{}) ValidationErrors {
```
lavalamp: Previously, I was very confident the path would always be reported correctly. After this change, we rely on every call site to correctly use WithPrefix or not as needed. Can we change this to capture the path when the error is created? It's fine to convert it to string or whatever.

typed/merge.go
```diff
@@ -140,81 +144,89 @@ func (w *mergingWalker) prepareDescent(pe fieldpath.PathElement, tr schema.TypeR
...
+			child := rhs.At(i)
 			pe, err := listItemToPathElement(t, i, child)
 			if err != nil {
-				errs = append(errs, w.errorf("rhs: element %v: %v", i, err.Error())...)
+				errs = append(errs, errorf("rhs: element %v: %v", i, err.Error())...)
 				// If we can't construct the path element, we can't
 				// even report errors deeper in the schema, so bail on
 				// this element.
-				continue
+				return
```
lavalamp: Why this change?

apelisse: Yeah, this one is definitely wrong, good catch

jennybuckley: did this cause a test to fail?

typed/merge.go
```diff
@@ -140,81 +144,89 @@ func (w *mergingWalker) prepareDescent(pe fieldpath.PathElement, tr schema.TypeR
...
+			child := lhs.At(i)
 			pe, err := listItemToPathElement(t, i, child)
 			if err != nil {
-				errs = append(errs, w.errorf("lhs: element %v: %v", i, err.Error())...)
+				errs = append(errs, errorf("lhs: element %v: %v", i, err.Error())...)
 				// If we can't construct the path element, we can't
 				// even report errors deeper in the schema, so bail on
 				// this element.
-				continue
+				return
```
lavalamp: This return also seems like it should be part of another PR with a separate justification?

typed/merge.go
```diff
@@ -140,81 +144,89 @@ func (w *mergingWalker) prepareDescent(pe fieldpath.PathElement, tr schema.TypeR
...
 			}
 			if observedLHS.Has(pe) {
-				errs = append(errs, w.errorf("lhs: duplicate entries for key %v", pe.String())...)
-				continue
+				errs = append(errs, errorf("lhs: duplicate entries for key %v", pe.String())...)
+				return
```
lavalamp: Here too.

typed/merge.go
```diff
@@ -226,45 +238,46 @@ func (w *mergingWalker) visitListItems(t *schema.List, lhs, rhs *value.List) (er
...
+			out = append(out, *w2.out)
 		}
 		w.finishDescent(w2)
 	}
 
-	if len(out.Items) > 0 {
-		w.out = &value.Value{ListValue: out}
+	if len(out) > 0 {
+		i := interface{}(out)
+		w.out = &i
```
lavalamp: Why is w an `*interface{}`? Shouldn't it be a `*value.Interface`?

typed/merge.go
apelisse: No, probably because: 1. We don't test errors a lot, 2. It will just prevent the list of errors to be complete.

typed/deduced_test.go
apelisse: I did not. I think the deepequal was an oversight.

typed/helpers.go
apelisse: > Why would this cause stuff to escape all the time? It shouldn't be a problem if it escapes only when there's an error?

Good questions. Unfortunately, It did escape in my tests none the less :-/

> I guess I'd have expected a []string for the replacement?

Does it make a difference? It will be concatenated anyway.

typed/helpers.go
apelisse: As mentioned before, it didn't work out that way. I can try again.

typed/helpers.go
apelisse: yeah, the condition is still required, I can change it to return an error.

typed/merge.go
apelisse: See rest of the discussion about `WithPrefix`

typed/merge.go
```diff
@@ -280,72 +293,72 @@ func (w *mergingWalker) doList(t *schema.List) (errs ValidationErrors) {
...
+		errs = append(errs, newErrs.WithPrefix(pe.String())...)
+	} else if w2.out != nil {
+		out[key] = *w2.out
+	}
+	w.finishDescent(w2)
+	return nil
+}
+
+func (w *mergingWalker) visitMapItems(t *schema.Map, lhs, rhs value.Map) (errs ValidationErrors) {
+	out := map[string]interface{}{}
```
lavalamp: Hm, I'm really sad about the reduction in type safety.

I think for the reflect based one to work this is going to have to switch to a factory and/or we're going to have to add getters/setters to the value interface :/

apelisse: Not disagreeing. We do plan on improving that.

apelisse: I suspect this change is big and old enough already though ... and blocking a couple of people :-/

typed/merge.go
apelisse: Not sure which of these you're talking about:
1. We don't collect the path in the same way, agreed.
2. Should I change back to `errorf(prefix + err)`, which is more similar to what we had before?

typed/merge.go
apelisse: because these errors have been created at a sub-level (within that pe) compared to the error above.

typed/validate.go
apelisse: Didn't work :-/

typed/helpers.go
lavalamp: Yeah I'd prefer to keep the exact same behavior in this PR wherever possible.

typed/helpers.go
lavalamp: I guess I don't actually care how you store it, if a string is best for you that's fine.

I do care that the path always corresponds to exact place in the object that sourced the error, for that I think you need to capture whatever it is you want to store exactly where the error is generated and not build it after the fact with WithPrefix calls :/ The other way to get the guarantee is to exhaustively test every error with its path, which I don't think is easier :/

typed/helpers.go
lavalamp: By "path" it doesn't need to be a literal path, a string is fine. I care about where it is captured, not what you store.

typed/merge.go
lavalamp: I think we need a mechanism that doesn't rely on every call site doing the right thing with the prefix. I don't care if you change the format. I care a little bit that it's consistent, but that should be the case if you satisfy my first constraint anyway :)

typed/merge.go
lavalamp: not asking for a change here :'(

typed/merge.go
```diff
@@ -280,72 +291,71 @@ func (w *mergingWalker) doList(t *schema.List) (errs ValidationErrors) {
...
+					rval = item
 				}
 			}
-			if newErrs := w2.merge(); len(newErrs) > 0 {
-				errs = append(errs, newErrs...)
-			} else if w2.out != nil {
-				out.Items = append(out.Items, value.Field{litem.Name, *w2.out})
+			errs = append(errs, w.visitMapItem(t, out, key, val, rval)...)
+			if rval != nil {
+				rval.Recycle()
```
lavalamp: optional nit: can `defer` this when rval is set?

typed/merge.go
```diff
@@ -280,72 +291,71 @@ func (w *mergingWalker) doList(t *schema.List) (errs ValidationErrors) {
...
 
 	if rhs != nil {
-		for j := range rhs.Items {
-			ritem := &rhs.Items[j]
+		rhs.Iterate(func(key string, val value.Value) bool {
 			if lhs != nil {
-				if _, ok := lhs.Get(ritem.Name); ok {
-					continue
+				if v, ok := lhs.Get(key); ok {
+					v.Recycle()
```
lavalamp: is there a `Has(key)` that could be called? This seems a little weird.

typed/parser_test.go
```diff
@@ -99,7 +97,7 @@ func BenchmarkConvertUnstructured(b *testing.B) {
 				b.ResetTimer()
 
 				for n := 0; n < b.N; n++ {
-					_ = u.AsValue().ToUnstructured(false)
+					// This is a no-op now.
```
lavalamp: might as well delete the whole for loop, this comment will be super mysterious to someone not looking at the diff.

typed/remove.go
```diff
@@ -20,61 +20,60 @@ import (
 )
 
 type removingWalker struct {
```
lavalamp: Now that you have the interface, I think it'd make a lot of sense to have this actually be an implementation of the interface that takes a value and a fieldset, and then do the "removal" lazily by pretending things in the FS don't exist? (can be followup)

typed/tofieldset.go
```diff
@@ -0,0 +1,160 @@
...
+	"sigs.k8s.io/structured-merge-diff/fieldpath"
+	"sigs.k8s.io/structured-merge-diff/schema"
+	"sigs.k8s.io/structured-merge-diff/value"
+)
+
+var tPool = sync.Pool{
+	New: func() interface{} { return &tosetObjectWalker{} },
+}
+
+func (tv TypedValue) toFieldSetWalker() *tosetObjectWalker {
```
lavalamp: Why not `toFieldSetWalker() *toFieldSetWalker`?

typed/tofieldset.go
```diff
@@ -0,0 +1,160 @@
...
+
+func (v *tosetObjectWalker) visitListItems(t *schema.List, list value.List) (errs ValidationErrors) {
+	for i := 0; i < list.Length(); i++ {
+		child := list.At(i)
+		pe, _ := listItemToPathElement(t, i, child)
+		v2 := v.prepareDescent(pe, t.ElementType)
+		v2.value = child
+		errs = append(errs, v2.toFieldSet()...)
+
+		v2.set.Insert(v2.path)
```
lavalamp: A future optimization could build the set a level at a time rather than using the whole path. I think walking the whole path again (as Insert must) makes this an N log N operation instead of just N?

typed/typed.go
```diff
@@ -79,15 +79,12 @@ func (tv TypedValue) Validate() error {
...
 // validation errors, if any were encountered.
 func (tv TypedValue) ToFieldSet() (*fieldpath.Set, error) {
-	s := fieldpath.NewSet()
-	w := tv.walker()
+	w := tv.toFieldSetWalker()
 	defer w.finished()
-	w.leafFieldCallback = func(p fieldpath.Path) { s.Insert(p) }
-	w.nodeFieldCallback = func(p fieldpath.Path) { s.Insert(p) }
-	if errs := w.validate(); len(errs) != 0 {
+	if errs := w.toFieldSet(); len(errs) != 0 {
```
lavalamp: Was this function sufficiently tested before, or were we relying on the general tests for the validation walk?

typed/typed.go
```diff
@@ -162,11 +158,11 @@ func (tv TypedValue) NormalizeUnions(new *TypedValue) (*TypedValue, error) {
 	var errs ValidationErrors
 	var normalizeFn = func(w *mergingWalker) {
 		if w.rhs != nil {
-			v := *w.rhs
+			v := w.rhs.Unstructured()
```
lavalamp: I guess we'll need setters, too, to make this work with reflect based ones.

typed/typed.go
```diff
@@ -206,7 +202,7 @@ func (tv TypedValue) NormalizeUnionsApply(new *TypedValue) (*TypedValue, error)
 }
 
 func (tv TypedValue) Empty() *TypedValue {
-	tv.value = value.Value{Null: true}
+	tv.value = nil
```
lavalamp: Shouldn't this use a value that is explicitly a Null?

typed/merge.go
apelisse: done

typed/merge.go
apelisse: We don't, but I agree :-/

typed/parser_test.go
apelisse: I've removed the loop and clarified the comment. I've kept the dummy benchmark though.

typed/remove.go
apelisse: Definitely part of the plan

typed/tofieldset.go
apelisse: Fixed

typed/tofieldset.go
apelisse: Yeah that's correct. We have optimizations we can try on these paths.

typed/typed.go
apelisse: Yes, we have extensive tests for toField and tests for validate.

typed/typed.go
apelisse: Clearly something I missed

typed/validate.go
lavalamp: Really? That's very sad

typed/union.go
```diff
@@ -105,38 +108,38 @@ type discriminator struct {
 	name string
 }
 
-func (d *discriminator) Set(m *value.Map, v discriminated) {
+func (d *discriminator) Set(m value.Map, v discriminated) {
```
lavalamp: followup: why is this method public?

typed/validate.go
lavalamp: How about this disgusting thing, does it work? (followup idea)

```
v.funcPtrStorage = func(key string, val value.Value) bool { ... }
m.Iterate(v.funcPtrStorage)
```

typed/union.go
```diff
@@ -105,38 +108,38 @@ type discriminator struct {
...
 		return ""
 	}
-	f, ok := m.Get(d.name)
+	val, ok := m.Get(d.name)
 	if !ok {
 		return ""
 	}
-	if f.Value.StringValue == nil {
+	if !val.IsString() {
 		return ""
```
lavalamp: I sure hope this is checked by the schema :/

typed/union.go
```diff
@@ -212,15 +215,15 @@ func newUnion(su *schema.Union) *union {
 
 // clear removes all the fields in map that are part of the union, but
 // the one we decided to keep.
-func (u *union) clear(m *value.Map, f field) {
+func (u *union) clear(m value.Map, f field) {
 	for _, fieldName := range u.f {
 		if field(fieldName) != f {
 			m.Delete(string(fieldName))
```
lavalamp: Wait, what? If the maps can do this why is the field deleter so complex? This deserves a TODO at least?

typed/validate.go
```diff
@@ -185,39 +154,30 @@ func (v *validatingObjectWalker) doList(t *schema.List) (errs ValidationErrors)
...
+		pe := fieldpath.PathElement{FieldName: &key}
+		tr := t.ElementType
+		if sf, ok := t.FindField(key); ok {
+			tr = sf.Type
+		} else if (t.ElementType == schema.TypeRef{}) {
+			errs = append(errs, errorf("field not declared in schema").WithPrefix(pe.String())...)
```
lavalamp: The prefix should be added by the caller, not here?

typed/validate.go
```diff
@@ -185,39 +154,30 @@ func (v *validatingObjectWalker) doList(t *schema.List) (errs ValidationErrors)
...
+			return false
 		}
-	}
+		v2 := v.prepareDescent(tr)
+		v2.value = val
+		// Giving pe.String as a parameter actually increases the allocations.
+		errs = append(errs, v2.validate(func() string { return pe.String() })...)
```
lavalamp: need to add prefix to the errors.

typed/validate.go
```diff
@@ -83,84 +64,76 @@ func (v *validatingObjectWalker) prepareDescent(pe fieldpath.PathElement, tr sch
...
+			observedKeys.Insert(pe)
 		}
-		observedKeys.Insert(pe)
-		v2 := v.prepareDescent(pe, t.ElementType)
+		v2 := v.prepareDescent(t.ElementType)
 		v2.value = child
-		errs = append(errs, v2.validate()...)
-
-		v2.doNode()
+		errs = append(errs, v2.validate(pe.String)...)
```
lavalamp: prefix the errors.

typed/validate.go
lavalamp: Ignore this I was just confused by line 164.

typed/validate.go
lavalamp: Ah it's because you're *not* calling validate here that you need to prefix, otherwise you lose the field name. Worth a comment.

typed/validate.go
lavalamp: Yeah, no, I was confused by line 164. Ignore this, sorry for the noise.

value/fields.go
```diff
@@ -0,0 +1,81 @@
...
+		return f[i].Name < f[j].Name
+	})
+}
+
+// Less compares two lists lexically.
+func (f FieldList) Less(rhs FieldList) bool {
+	return f.Compare(rhs) == -1
+}
+
+// Less compares two lists lexically. The result will be 0 if f==rhs, -1
```
lavalamp: Comment doesn't match method name.

value/less_test.go
```diff
@@ -25,301 +25,228 @@ func TestValueLess(t *testing.T) {
 		name string
 		// we expect a < b and !(b < a) unless eq is true, in which
 		// case we expect less to return false in both orders.
-		a, b Value
+		a, b interface{}
 		eq   bool
 	}{
 		{
-			name: "Invalid-1",
```
lavalamp: Is it no longer possible to construct an invalid value? Is there a test that somehow verifies that of a Value's IsXXX() methods, exactly one always returns true?

value/list.go
```diff
@@ -0,0 +1,80 @@
...
+type List interface {
+	// Length returns how many items can be found in the map.
+	Length() int
+	// At returns the item at the given position in the map. It will
+	// panic if the index is out of range.
+	At(int) Value
+}
+
+// Equals compares two lists lexically.
+func ListEquals(lhs, rhs List) bool {
```
lavalamp: Idea for discussion (for followup or ignore):

```
type ListOp struct {
  LHS, RHS List
}

func (o ListOp) Equals() bool { ... }
func (o ListOp) Less() bool { ... }
func (o ListOp) Compare() int { ... }

```

Not claiming this is better.

value/list.go
```diff
@@ -0,0 +1,80 @@
...
+			rv.Recycle()
+			return false
+		}
+		lv.Recycle()
+		rv.Recycle()
+	}
+	return true
+}
+
+// Less compares two lists lexically.
```
lavalamp: If you don't like the idea, you should fix the comments to match the function names :)

value/map.go
```diff
@@ -0,0 +1,108 @@
...
+	Equals(other Map) bool
+	// Iterate runs the given function for each key/value in the
+	// map. Returning false in the closure prematurely stops the
+	// iteration.
+	Iterate(func(key string, value Value) bool) bool
+	// Length returns the number of items in the map.
+	Length() int
+}
+
+// Less compares two maps lexically.
```
lavalamp: MapLess (below functions wrong, too)

value/map.go
```diff
@@ -0,0 +1,108 @@
...
+// Map represents a Map or go structure.
+type Map interface {
+	// Set changes or set the value of the given key.
+	Set(key string, val Value)
+	// Get returns the value for the given key, if present, or (nil, false) otherwise.
+	Get(key string) (Value, bool)
+	// Delete removes the key from the map.
+	Delete(key string)
+	// Equals compares the two maps, and return true if they are the same, false otherwise.
+	Equals(other Map) bool
```
lavalamp: Shouldn't have both this and the MapEquals function below?

If this is some sort of fastpath then probably the name and comment should change? (and how can a caller tell if two Maps are compatible)?

value/mapunstructured.go
```diff
@@ -0,0 +1,135 @@
...
+
+func (m mapUnstructuredInterface) Set(key string, val Value) {
+	m[key] = val.Unstructured()
+}
+
+func (m mapUnstructuredInterface) Get(key string) (Value, bool) {
+	if v, ok := m[key]; !ok {
+		return nil, false
+	} else {
+		return NewValueInterface(v), true
```
lavalamp: Offering a `Has(key string) bool` method would save some garbage if that's all the caller needs to know.

typed/union.go
apelisse: It doesn't really matter since the type is private anyway. I've used it as a way to indicate that the method is public to other types within the package.

typed/union.go
apelisse: It's not that easy? I can't remember if we're checking for type aliases to strings. What do you suggest?

typed/union.go
apelisse: yeah, that part is not specifically that hard. It's just that we're not doing much in place right now.

value/less_test.go
apelisse: We could make a "IsInvalid" if that's useful. The content of Value can be litteraly anything, some things are invalid like ... `value.NewValueInterface(struct{})`.

value/valueunstructured.go
```diff
@@ -0,0 +1,159 @@
...
+	}
+	if _, ok := v.Value.(map[interface{}]interface{}); ok {
+		return true
+	}
+	return false
+}
+
+func (v valueUnstructured) Map() Map {
+	if v.Value == nil {
+		return mapUnstructuredString(nil)
```
lavalamp: Why this and not just `nil`? e.g. mapUnstructuredString's Get doesn't look like it's prepared to deal with this.

I think `panic`ing is an option, too--that'd be like the built-in reflect package.

value/list.go
apelisse: I really don't see the point.

value/valueunstructured.go
```diff
@@ -0,0 +1,159 @@
...
+	vi := viPool.Get().(*valueUnstructured)
+	vi.Value = v
+	return Value(vi)
+}
+
+type valueUnstructured struct {
+	Value interface{}
+}
+
+func (v valueUnstructured) IsMap() bool {
```
lavalamp: Is it possible for IsMap and IsNull to return true at the same time?

value/less_test.go
apelisse: Which means I could re-add these tests

value/mapunstructured.go
apelisse: It means it has to be implemented for every single implementation of the interface. I don't disagree with the general comment, we just have to be careful what we decide (we can't grow this list indefinitely).

value/valueunstructured.go
```diff
@@ -0,0 +1,159 @@
...
+		return false
+	} else if _, ok := v.Value.(int); ok {
+		return true
+	} else if _, ok := v.Value.(int8); ok {
+		return true
+	} else if _, ok := v.Value.(int16); ok {
+		return true
+	} else if _, ok := v.Value.(int32); ok {
+		return true
+	} else if _, ok := v.Value.(int64); ok {
```
lavalamp: readability nit: wouldn't a type switch look nicer?

value/valueunstructured.go
```diff
@@ -0,0 +1,159 @@
...
+}
+
+func (v valueUnstructured) Float() float64 {
+	if f, ok := v.Value.(float32); ok {
+		return float64(f)
+	}
+	return v.Value.(float64)
+}
+
+func (v valueUnstructured) IsInt() bool {
```
lavalamp: What are `uint`s?

value/map.go
apelisse: I can clarify in the comment. One is a general implementation that can be used to implement this. But callers should use this one.

value/valueunstructured.go
apelisse: Switch can't get inline, not sure it matters ...

value/value.go
```diff
@@ -17,76 +17,170 @@ limitations under the License.
...
+func WriteJSONStream(v Value, stream *jsoniter.Stream) {
+	stream.WriteVal(v.Unstructured())
+}
+
+// ToYAML marshals a value as YAML.
+func ToYAML(v Value) ([]byte, error) {
+	return yaml.Marshal(v.Unstructured())
+}
+
+// A Value corresponds to an 'atom' in the schema.
```
lavalamp: Move to top of file?

value/value.go
```diff
@@ -17,76 +17,170 @@ limitations under the License.
...
+	// IsInt returns true if the Value can be converted to a int64,
+	// false otherwise.
+	IsInt() bool
+	// IsFloat returns true if the Value can be converted to a
+	// float64, false otherwise.
+	IsFloat() bool
+	// IsString returns true if the Value can be converted to a
+	// string, false otherwise.
+	IsString() bool
+	// IsMap returns true if the Value is null, false otherwise.
```
lavalamp: IsNull

Make it clear in the comments somewhere that it's a requirement for implementors that exactly one of the IsXXX functions must return true at all times. The schema permits the concept of "a null map" but the value type doesn't-- a thing is either null or a map, never both at the same time.

apelisse: > Make it clear in the comments somewhere that it's a requirement for implementors that exactly one of the IsXXX functions must return true at all times. 

I don't think that's true

value/value.go
```diff
@@ -17,76 +17,170 @@ limitations under the License.
...
+}
+
+// ToYAML marshals a value as YAML.
+func ToYAML(v Value) ([]byte, error) {
+	return yaml.Marshal(v.Unstructured())
+}
+
+// A Value corresponds to an 'atom' in the schema.
+type Value interface {
+	// IsMap returns true if the Value can be converted to a Map,
```
lavalamp: I would not use the wording "converted to", since that makes it sound like an object could be multiple of these things.

value/value.go
lavalamp: Consider adding a test for this for the unstructured implementation, since you removed my tests of invalid values. :)

value/value.go
```diff
@@ -17,76 +17,170 @@ limitations under the License.
...
+	Float() float64
+	// String converts the Value into a string (or panic if the type
+	// doesn't allow it).
+	String() string
+
+	// Returns a value of this type that is no longer needed. The
+	// value shouldn't be used after this call.
+	Recycle()
+
+	// Converts the Value into an Unstructured interface{}.
```
lavalamp: We'll probably want to change this to make the reflect stuff optimal.

I think a `NewOfSameType() Value` might come in handy, as well as a `Set(valueToCopy Value)`. (followup)

value/valueunstructured.go
apelisse: I don't think so

value/value.go
lavalamp: er, now that I see your comment:

* say that that condition makes the value "invalid"
* and put the tests with invalid values back in?

value/mapunstructured.go
lavalamp: My guess is that `Has` should make the cut, but we can add it in a followup if needed.

value/map.go
lavalamp: Yeah that definitely needs a comment.

value/less_test.go
lavalamp: Yeah, funcs, channels, structs, uints...

that last one should probably be fixed...

You might need a test object to make it easy to represent arbitrary invalid objects to put these tests back. The old Value actually might make that easy ;)

apelisse: Yeah, I'm fixing `uint`s right now, I'll add back the tests too

typed/union.go
lavalamp: I don't know, but if this is not enforced by the schema (which really should be able to follow aliases), then this is going to be a VERY mysterious failure mode for someone.

typed/union.go
lavalamp: Needs a comment either way. It doesn't seem great for the deleter to implicitly do a copy + delete.

typed/union.go
apelisse: Agreed, we should improve that

typed/union.go
lavalamp: I think I tried to do something similar once and got the feedback that it was really weird to do that. Does this pass go vet?

apelisse: Yes

value/value.go
apelisse: Absolutely agree!

value/valueunstructured.go
apelisse: You're right, that doesn't make a mistake. There are a few places where we treat nil as a map so I don't think we can panic here.

lavalamp: We should fix that for real-- I'd honestly prefer to make people always check IsNull first.

Panic here and exhaustively test?

None of the other types here permit nils.

apelisse: Yeah, it crashes in union, should be fairly limited and easy to test, will fix now

apelisse:
OK I think I've addressed it all

value/valueunstructured.go
apelisse: solved

value/valueunstructured.go
lavalamp: I don't know what that means but this isn't important.

lavalamp:
OK we need to make sure master is marked as a v2 module asap if we haven't already. This definitely breaks clients.

/lgtm

PR #131, opened 2019-12-04
Fix useless error message 'invalid atom'

jennybuckley:
For a while now, any schema errors would result in a very useless error message "invalid atom", this was confusing, and also made debugging wiring the schema and value interface very tricky. This problem was originally caused when I made a schema able to represent something which could possibly be one of multiple types (was needed for CRD support) because we started hitting the "schema error: invalid atom" error before the validation code can generate a more useful error message.

Some examples of this problem (from typed/validate_test.go): 

| input | old error | new error |
|---|---|---|
| **TestSchemaValidation/simple_pair:** |
| `{"key":{"foo":true}}` | `.key: schema error: invalid atom: inlined` | `.key: expected string, got {foo=true}` |
| `{"key":[1, 2]}` | `.key: schema error: invalid atom: inlined` | `.key: expected string, got [1, 2]` |
| **TestSchemaValidation/associative_list:** |
| `{"key":true,"value":1}` | `.key: schema error: invalid atom: inlined\n.value: schema error: invalid atom: inlined` | `.key: field not declared in schema\n.value: field not declared in schema` |
| `{"list":{"key":true,"value":1}}` | `.list: schema error: invalid atom: named type: myList` | `.list: expected list, got {key=true;value=1}` |
| **TestSchemaValidation/struct_grab_bag:** |
| `{"numeric":{"a":1}}` | `.numeric: schema error: invalid atom: inlined` | `.numeric: expected numeric (int or float), got {a=1}` |
| `{"bool":["foo"]}` | `.bool: schema error: invalid atom: inlined` | `.bool: expected boolean, got ["foo"]` |

### DIFF
--- a/typed/helpers.go
+++ b/typed/helpers.go
@@ -114,17 +114,27 @@ func resolveSchema(s *schema.Schema, tr schema.TypeRef, v *value.Value, ah atomH
 	return handleAtom(a, tr, ah)
 }
 
-func deduceAtom(a schema.Atom, v *value.Value) schema.Atom {
+// deduceAtom determines which of the possible types in atom 'atom' applies to value 'val'.
+// If val is of a type allowed by atom, return a copy of atom with all other types set to nil.
+// if val is nil, or is not of a type allowed by atom, just return the original atom,
+// and validation will fail at a later stage. (with a more useful error)
+func deduceAtom(atom schema.Atom, val *value.Value) schema.Atom {
 	switch {
-	case v == nil:
-	case v.FloatValue != nil, v.IntValue != nil, v.StringValue != nil, v.BooleanValue != nil:
-		return schema.Atom{Scalar: a.Scalar}
-	case v.ListValue != nil:
-		return schema.Atom{List: a.List}
-	case v.MapValue != nil:
-		return schema.Atom{Map: a.Map}
+	case val == nil:
+	case val.FloatValue != nil, val.IntValue != nil, val.StringValue != nil, val.BooleanValue != nil:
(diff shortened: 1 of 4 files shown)


typed/helpers.go
```diff
@@ -114,17 +114,27 @@ func resolveSchema(s *schema.Schema, tr schema.TypeRef, v *value.Value, ah atomH
 	return handleAtom(a, tr, ah)
 }
 
-func deduceAtom(a schema.Atom, v *value.Value) schema.Atom {
+// deduceAtom determines which of the possible types in atom 'atom' applies to value 'val'.
```
lavalamp: I don't understand this, atoms only permit one type?

jennybuckley: We changed that to allow for a generic schema for CRDs

https://github.com/kubernetes-sigs/structured-merge-diff/pull/73

lavalamp: I don't see how that's better than untyped but I'll take your word for it for the moment.

jennybuckley: The reason we did it this way was, to be able to clearly define the behavior of different Untyped elementRelationships, like "atomic" and "separable". We wanted different a elementRelationship and elementType for lists and maps in what would have been the "untyped separable" case (atomic lists and separable maps), and we wanted to solve it in a generic way so we don't have to duplicate the logic and don't have to hide the definition of its behavior in code.

 Also, as a side effect we should also be able to eventually use this for CRD validation shemas which define a field like "map or array of maps".

apelisse:
We should try to back-port that to v1.0 branch

PR #132, opened 2019-12-05
This conversion doesn't need to check keys for uniqueness

lavalamp:


jennybuckley:
/close
This isn't possible now, with the new value interface

PR #133, opened 2019-12-09
add extended remove method for typed values

kwiesmueller:
While working on an implementation of https://github.com/kubernetes/enhancements/pull/1123
i discovered that removing fields is not behaving like I expected.

Provided an original fieldSet like:
```
.status.phase
```
and a fieldSet to be removed:
```
.status
```
The current Difference and Intersect methods for fieldSets do not match all child paths when a parent path is defined, because `.status` != `.status.phase`.

For most cases, this seems reasonable (like ownership).

But to define a set of fields that should be removed from the object, it is impractical.
To avoid a user has to define every deep path to be removed (for status wiping),
a second Remove method is added to value.TypedValue that removes a fieldpath once found.

It extends the already available RemoveItems behavior to also remove from maps.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,12 +1,13 @@
 module sigs.k8s.io/structured-merge-diff/v3
 
-require gopkg.in/yaml.v2 v2.2.1
+require gopkg.in/yaml.v2 v2.2.2
 
 require (
 	github.com/google/gofuzz v1.0.0
 	github.com/json-iterator/go v1.1.6
 	github.com/modern-go/concurrent v0.0.0-20180306012644-bacd9c7ef1dd // indirect
 	github.com/modern-go/reflect2 v1.0.1 // indirect
+	github.com/stretchr/testify v1.5.0 // indirect
 )
 
 go 1.13
--- a/internal/fixture/state.go
+++ b/internal/fixture/state.go
@@ -130,7 +130,7 @@ func (s *State) UpdateObject(tv *typed.TypedValue, version fieldpath.APIVersion,
 
 // Update the current state with the passed in object
 func (s *State) Update(obj typed.YAMLObject, version fieldpath.APIVersion, manager string) error {
-	tv, err := s.Parser.Type(string(version)).FromYAML(FixTabsOrDie(obj))
+	tv, err := s.parseConfig(version, obj)
 	if err != nil {
 		return err
 	}
@@ -159,7 +159,7 @@ func (s *State) ApplyObject(tv *typed.TypedValue, version fieldpath.APIVersion, 
 
 // Apply the passed in object to the current state
 func (s *State) Apply(obj typed.YAMLObject, version fieldpath.APIVersion, manager string, force bool) error {
-	tv, err := s.Parser.Type(string(version)).FromYAML(FixTabsOrDie(obj))
+	tv, err := s.parseConfig(version, obj)
 	if err != nil {
 		return err
 	}
@@ -182,6 +182,10 @@ func (s *State) CompareLive(obj typed.YAMLObject, version fieldpath.APIVersion) 
 		return nil, err
 	}
 	return live.Compare(tv)
+}
+
+func (s *State) parseConfig(apiVersion fieldpath.APIVersion, obj typed.YAMLObject) (*typed.TypedValue, error) {
+	return s.Parser.Type(string(apiVersion)).FromYAML(FixTabsOrDie(obj))
 }
 
 // dummyConverter doesn't convert, it just returns the same exact object, as long as a version is provided.
@@ -471,12 +475,12 @@ func (tc TestCase) TestWithConverter(parser Parser, converter merge.Converter) e
 
 	// If LastObject was specified, compare it with LiveState
 	if tc.Object != typed.YAMLObject("") {
-		comparison, err := state.CompareLive(tc.Object, tc.APIVersion)
-		if err != nil {
-			return fmt.Errorf("failed to compare live with config: %v", err)
+		expect := ExpectState{
+			APIVersion: tc.APIVersion, Object: tc.Object,
 		}
-		if !comparison.IsSame() {
-			return fmt.Errorf("expected live and config to be the same:\n%v\nConfig: %v\n", comparison, value.ToString(state.Live.AsValue()))
+
+		if err := expect.run(&state); err != nil {
+			return err
 		}
 	}
 
@@ -504,4 +508,45 @@ func (tc TestCase) TestWithConverter(parser Parser, converter merge.Converter) e
 	}
 
 	return nil
+}
+
+// PrintState is an Operation printing the current state to help with debugging tests
+type PrintState struct{}
+
+var _ Operation = PrintState{}
+
+func (op PrintState) run(s *State) error {
+	fmt.Println(value.ToString(s.Live.AsValue()))
+	return nil
+}
+
+func (op PrintState) preprocess(_ Parser) (Operation, error) {
+	return op, nil
+}
+
+// ExpectState is an Operation comparing the current state to the defined config to help with debugging tests
+type ExpectState struct {
+	APIVersion fieldpath.APIVersion
+	Object     typed.YAMLObject
+}
+
+var _ Operation = ExpectState{}
+
+func (op ExpectState) run(state *State) error {
+	comparison, err := state.CompareLive(op.Object, op.APIVersion)
+	if err != nil {
+		return fmt.Errorf("failed to compare live with config: %v", err)
+	}
+	if !comparison.IsSame() {
+		config, err := state.parseConfig(op.APIVersion, op.Object)
+		if err != nil {
+			return fmt.Errorf("failed to parse config: %w", err)
+		}
+		return fmt.Errorf("expected live and config to be the same:\n%v\nConfig: %v\nLive:   %v\n", comparison, value.ToString(config.AsValue()), value.ToString(state.Live.AsValue()))
+	}
+	return nil
+}
+
+func (op ExpectState) preprocess(parser Parser) (Operation, error) {
+	return op, nil
 }
--- a/merge/multiple_appliers_test.go
+++ b/merge/multiple_appliers_test.go
@@ -670,7 +670,7 @@ func TestMultipleAppliersNestedType(t *testing.T) {
 			},
 			Object: `
 				mapOfMapsRecursive:
-				  a:
+				  a: {}
 				  c:
 				    d:
 				      e:
@@ -788,7 +788,7 @@ func TestMultipleAppliersDeducedType(t *testing.T) {
 				},
 			},
 			Object: `
-				a:
+				a: {}
 				c:
 				  d:
 				    e:
--- a/typed/remove.go
+++ b/typed/remove.go
@@ -25,14 +25,16 @@ type removingWalker struct {
 	schema    *schema.Schema
 	toRemove  *fieldpath.Set
 	allocator value.Allocator
+	onlyItems bool
 }
 
-func removeItemsWithSchema(val value.Value, toRemove *fieldpath.Set, schema *schema.Schema, typeRef schema.TypeRef) value.Value {
+func removeWithSchema(val value.Value, toRemove *fieldpath.Set, schema *schema.Schema, typeRef schema.TypeRef, onlyItems bool) value.Value {
 	w := &removingWalker{
 		value:     val,
 		schema:    schema,
 		toRemove:  toRemove,
 		allocator: value.NewFreelistAllocator(),
+		onlyItems: onlyItems,
 	}
 	resolveSchema(schema, typeRef, val, w)
 	return value.NewValueInterface(w.out)
@@ -46,8 +48,10 @@ func (w *removingWalker) doScalar(t *schema.Scalar) ValidationErrors {
 func (w *removingWalker) doList(t *schema.List) (errs ValidationErrors) {
 	l := w.value.AsListUsing(w.allocator)
 	defer w.allocator.Free(l)
-	// If list is null, empty, or atomic just return
-	if l == nil || l.Length() == 0 || t.ElementRelationship == schema.Atomic {
+
+	// If list is null or empty just return
+	if l == nil || l.Length() == 0 {
+		w.out = w.value.Unstructured()
 		return nil
 	}
 
@@ -63,13 +67,11 @@ func (w *removingWalker) doList(t *schema.List) (errs ValidationErrors) {
 			continue
 		}
 		if subset := w.toRemove.WithPrefix(pe); !subset.Empty() {
-			item = removeItemsWithSchema(item, subset, w.schema, t.ElementType)
+			item = removeWithSchema(item, subset, w.schema, t.ElementType, w.onlyItems)
 		}
 		newItems = append(newItems, item.Unstructured())
 	}
-	if len(newItems) > 0 {
-		w.out = newItems
-	}
+	w.out = newItems
 	return nil
 }
 
@@ -78,8 +80,9 @@ func (w *removingWalker) doMap(t *schema.Map) ValidationErrors {
 	if m != nil {
 		defer w.allocator.Free(m)
 	}
-	// If map is null, empty, or atomic just return
-	if m == nil || m.Empty() || t.ElementRelationship == schema.Atomic {
+	// If map is null or empty just return
+	if m == nil || m.Empty() {
+		w.out = w.value.Unstructured()
 		return nil
 	}
 
@@ -92,6 +95,11 @@ func (w *removingWalker) doMap(t *schema.Map) ValidationErrors {
 	m.Iterate(func(k string, val value.Value) bool {
 		pe := fieldpath.PathElement{FieldName: &k}
 		path, _ := fieldpath.MakePath(pe)
+
+		if !w.onlyItems && w.toRemove.Has(path) {
+			return true
+		}
+
 		fieldType := t.ElementType
 		if ft, ok := fieldTypes[k]; ok {
 			fieldType = ft
@@ -101,13 +109,11 @@ func (w *removingWalker) doMap(t *schema.Map) ValidationErrors {
 			}
 		}
 		if subset := w.toRemove.WithPrefix(pe); !subset.Empty() {
-			val = removeItemsWithSchema(val, subset, w.schema, fieldType)
+			val = removeWithSchema(val, subset, w.schema, fieldType, w.onlyItems)
 		}
 		newMap[k] = val.Unstructured()
 		return true
 	})
-	if len(newMap) > 0 {
-		w.out = newMap
-	}
+	w.out = newMap
 	return nil
 }
--- a/typed/remove_test.go
+++ b/typed/remove_test.go
@@ -1,0 +1,492 @@
+package typed
+
+import (
+	"fmt"
+	"testing"
+
+	"sigs.k8s.io/structured-merge-diff/v3/fieldpath"
+	"sigs.k8s.io/structured-merge-diff/v3/value"
+)
+
+var (
+	// Short names for readable test cases.
+	_NS  = fieldpath.NewSet
+	_P   = fieldpath.MakePathOrDie
+	_KBF = fieldpath.KeyByFields
+	_V   = value.NewValueInterface
+)
+
+func TestRemoveDeduced(t *testing.T) {
+	var cases = []struct {
+		object YAMLObject
+		remove *fieldpath.Set
+		expect YAMLObject
+	}{
+		{
+			object: `{}`,
+			remove: _NS(_P("a")),
+			expect: `{}`,
+		},
+		{
+			object: `{"a": "value"}`,
+			remove: _NS(_P("a")),
+			expect: `{}`,
+		},
+		{
+			object: `{"a": "value", "b": "value"}`,
+			remove: _NS(_P("a")),
+			expect: `{"b": "value"}`,
+		},
+		{
+			object: `{"a": "value", "b": {}}`,
+			remove: _NS(_P("a")),
+			expect: `{"b": {}}`,
+		},
+		{
+			object: `{"a": "value", "b": {"c":"value"}}`,
+			remove: _NS(_P("b")),
+			expect: `{"a": "value"}`,
+		},
+		{
+			object: `{"a": "value", "b": {"c":"value"}}`,
+			remove: _NS(_P("b", "c")),
+			expect: `{"a": "value", "b": {}}`,
+		},
+		{
+			object: `{"a": "value", "b": []}`,
+			remove: _NS(_P("b")),
+			expect: `{"a": "value"}`,
+		},
+		{
+			object: `{"a": "value", "b": ["c"]}`,
+			remove: _NS(_P("b")),
+			expect: `{"a": "value"}`,
+		},
+		{
+			object: `{"a": "value", "b": ["c"]}`,
+			remove: _NS(_P("b", "c")),
+			// `c` won't get removed, as it is not a field, while being addressed as one in MakePathOrDie.
+			// There is a test for removing `c` below.
+			expect: `{"a": "value", "b": ["c"]}`,
+		},
+		{
+			object: `{"a": "value", "b": ["c", "d"]}`,
+			remove: _NS(_P("b", "c")),
+			// `c` won't get removed, as it is not a field, while being addressed as one in MakePathOrDie.
+			// There is a test for removing `c` below.
+			expect: `{"a": "value", "b": ["c", "d"]}`,
+		},
+		{
+			object: `{"a": "value", "b": ["c"]}`,
+			remove: _NS(_P("b", 0)),
+			expect: `{"a": "value", "b": []}`,
+		},
+		{
+			object: `{"a": "value", "b": ["c", "d"]}`,
+			remove: _NS(_P("b", 1)),
+			expect: `{"a": "value", "b": ["c"]}`,
+		},
+		{
+			object: `{"a": "value", "b": [{"c": "value"}, {"d": "value"}]}`,
+			remove: _NS(_P("b", "c")),
+			// `c` won't get removed, as it is not a field, but a field inside a list item, MakePath does not address it accordingly.
+			// It would be unexpected to remove all `c`s from a lists items.
+			// To remove only the list item or `c` the path must be specified differently (see below).
+			expect: `{"a": "value", "b": [{"c": "value"}, {"d": "value"}]}`,
+		},
+		{
+			object: `{"a": "value", "b": [{"c": "value"}, {"d": "value"}]}`,
+			remove: _NS(_P("b", 0)),
+			expect: `{"a": "value", "b": [{"d": "value"}]}`,
+		},
+		{
+			object: `{"a": "value", "b": [{"c": "value"}, {"d": "value"}]}`,
+			remove: _NS(_P("b", 0, "c")),
+			expect: `{"a": "value", "b": [{}, {"d": "value"}]}`,
+		},
+		{
+			object: `{"a": "value", "b": {"c":"value", "d":{"e":"value"}}}`,
+			remove: _NS(_P("b", "d")),
+			expect: `{"a": "value", "b": {"c":"value"}}`,
+		},
+		{
+			object: `{"a": "value", "b": {"c":"value", "d": {"e":"value"}}}`,
+			remove: _NS(_P("b", "d", "e")),
+			expect: `{"a": "value", "b": {"c":"value", "d": {}}}`,
+		},
+	}
+
+	for i, c := range cases {
+		c := c
+		t.Run(fmt.Sprintf("case-%d", i), func(t *testing.T) {
+			t.Parallel()
+
+			obj, err := DeducedParseableType.FromYAML(c.object)
+			if err != nil {
+				t.Fatalf("unable to parse/validate object: %v\n%v", err, c.object)
+			}
+
+			parseable := &ParseableType{
+				Schema:  obj.schema,
+				TypeRef: obj.typeRef,
+			}
+			expectTyped, err := parseable.FromYAML(c.expect)
+			if err != nil {
+				t.Fatalf("unable to parse/validate expected object: %v\n%v", err, c.expect)
+			}
+			expect := expectTyped.AsValue()
+
+			result := obj.Remove(c.remove).AsValue()
+			if !value.Equals(result, expect) {
+				t.Fatalf("unexpected result after Remove:\ngot: %v\nexp: %v",
+					value.ToString(result), value.ToString(expect),
+				)
+			}
+
+			result = obj.RemoveItems(c.remove).AsValue()
+			if !value.Equals(result, expect) {
+				t.Fatalf("unexpected result after RemoveItems:\ngot: %v\nexp: %v",
+					value.ToString(result), value.ToString(expect),
+				)
+			}
+		})
+	}
+}
+
+func TestRemove(t *testing.T) {
+	var cases = []struct {
+		object          YAMLObject
+		schema          YAMLObject
+		remove          *fieldpath.Set
+		expect          YAMLObject
+		expectItemsOnly YAMLObject
+	}{
+		{
+			schema: `types:
+- name: type
+  map:
+    fields:
+    - name: a
+      type:
+        scalar: string
+    - name: b
+      type:
+        map:
+          elementType:
+            scalar: string
+`,
+			object:          `{"a": "value", "b":{"c":"value", "d": "value"}}`,
+			remove:          _NS(_P("b", "d")),
+			expect:          `{"a": "value", "b":{"c":"value"}}`,
+			expectItemsOnly: `{"a": "value", "b":{"c":"value"}}`,
+		},
+		{
+			schema: `types:
+- name: type
+  map:
+    fields:
+    - name: a
+      type:
+        scalar: string
+    - name: b
+      type:
+        map:
+          elementType:
+            scalar: string
+`,
+			object:          `{"a": "value", "b":{"c":"value", "d": "value"}}`,
+			remove:          _NS(_P("b")),
+			expect:          `{"a": "value"}`,
+			expectItemsOnly: `{"a": "value", "b":{"c":"value", "d": "value"}}}`,
+		},
+		{
+			schema: `types:
+- name: type
+  map:
+    fields:
+    - name: a
+      type:
+        scalar: string
+    - name: b
+      type:
+        map:
+          fields:
+          - name: c
+            type:
+              scalar: string
+`,
+			object:          `{"a": "value", "b":{"c":"value"}}`,
+			remove:          _NS(_P("b")),
+			expect:          `{"a": "value"}`,
+			expectItemsOnly: `{"a": "value", "b":{"c":"value"}}`,
+		},
+		{
+			schema: `types:
+- name: type
+  map:
+    fields:
+    - name: a
+      type:
+        scalar: string
+    - name: b
+      type:
+        map:
+          fields:
+          - name: c
+            type:
+              scalar: string
+`,
+			object:          `{"a": "value", "b":{"c":"value"}}`,
+			remove:          _NS(_P("b")),
+			expect:          `{"a": "value"}`,
+			expectItemsOnly: `{"a": "value", "b":{"c":"value"}}`,
+		},
+		{
+			schema: `types:
+- name: type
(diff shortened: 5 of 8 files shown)


typed/remove.go
```diff
@@ -23,13 +23,15 @@ type removingWalker struct {
 	value    *value.Value
 	schema   *schema.Schema
 	toRemove *fieldpath.Set
+	shallow  bool
```
kwiesmueller: Not entirely happy with the name `shallow`.
Thought about recursive, but didn't feel right as well.

Left it this way for now, happy for suggestions.

typed/remove_test.go
```diff
@@ -0,0 +1,139 @@
...
+	"testing"
+
+	"sigs.k8s.io/structured-merge-diff/fieldpath"
+	"sigs.k8s.io/structured-merge-diff/schema"
+	"sigs.k8s.io/structured-merge-diff/value"
+)
+
+func TestRemoveShallowRemovesParents(t *testing.T) {
+
+	test, err := value.FromJSON([]byte(`
```
kwiesmueller: Could use more test cases

typed/remove_test.go
```diff
@@ -0,0 +1,139 @@
...
+
+	remove := fieldpath.NewSet(
+		fieldpath.MakePathOrDie("removeMap"),
+		fieldpath.MakePathOrDie("removeList"),
+	)
+
+	str := schema.String
+
+	atom := schema.Atom{
+		Map: &schema.Map{
```
kwiesmueller: This is probably easier, was a quick start to get a test.

value/value.go
```diff
@@ -187,7 +187,7 @@ func Equals(lhs, rhs Value) bool {
 	return true
 }
 
-// String returns a human-readable representation of the value.
+// ToString returns a human-readable representation of the value.
```
apelisse: That brings inconsistency. I suspect we could rename them `AsXXX`, like `AsString()`

typed/typed.go
```diff
@@ -140,7 +140,13 @@ func (tv TypedValue) Compare(rhs *TypedValue) (c *Comparison, err error) {
 
 // RemoveItems removes each provided list or map item from the value.
 func (tv TypedValue) RemoveItems(items *fieldpath.Set) *TypedValue {
-	tv.value = removeItemsWithSchema(tv.value, items, tv.schema, tv.typeRef)
+	tv.value = removeItemsWithSchema(tv.value, items, tv.schema, tv.typeRef, false)
+	return &tv
+}
+
+// Remove removes each provided path from the value.
+func (tv TypedValue) Remove(paths *fieldpath.Set) *TypedValue {
```
apelisse: I don't know if that name is explicit enough (nor is the comment/doc above).

I would suggest `RemoveDeep()`, which will sound counter-intuitive to you, since you used shallow where I would have used `deep` :-))

typed/remove_test.go
apelisse: That test is really hard to parse and follow. I think you could do the same thing with a "deduced" schema (since you wouldn't have to specify the schema). There are also other schemas available throughout this test suite that you can use.

And yeah, it's good that we're testing how the two functions behave differently, but we need more examples :-/

typed/typed.go
kwiesmueller: I'm okay with using deep instead if shallow.
Thought about it and deep didn't feel right looking at what it does, but I'm okay with it any way.

value/value.go
kwiesmueller: I just renamed the comment so golint is not complaining.
But I could open another PR renaming the functions.
Did you see my question regarding the naming in slack: https://kubernetes.slack.com/archives/C9VRU4PRB/p1576069598284000

typed/remove_test.go
kwiesmueller: Right, I'll try deduced and add more cases.

value/value.go
apelisse: Yes, I agree that this was a poor choice for naming. I'd be happy to see a PR to rename all the `Type` methods into `AsType`.

typed/typed.go
apelisse: Shallow and Deep are both used for copy, comparison, etc. I think it'll be idiomatic for people here too.

typed/remove.go
```diff
@@ -86,6 +88,13 @@ func (w *removingWalker) doMap(t *schema.Map) ValidationErrors {
 	m.Iterate(func(k string, val value.Value) bool {
 		pe := fieldpath.PathElement{FieldName: &k}
 		path, _ := fieldpath.MakePath(pe)
+
+		if w.deep {
+			if w.toRemove.Has(path) {
```
apelisse: Nit: maybe merge these two conditions? `if w.deep && w.toRemove...`

value/value.go
apelisse: Can we keep this change for a different pull-request?

typed/remove_test.go
```diff
@@ -0,0 +1,133 @@
...
+			object: `{"a": "value", "b": {"c":"value"}}`,
+			remove: fieldpath.NewSet(fieldpath.MakePathOrDie("b")),
+			deep:   true,
+			expect: `{"a": "value"}`,
+		},
+		{
+			object: `{"a": "value", "b":{"c":"value"}}`,
+			remove: fieldpath.NewSet(fieldpath.MakePathOrDie("b")),
+			deep:   false,
+			// TODO: this seems to behave differently with a non deduced schema: {"a": "value", "b":{"c":"value"}} which is the reason for this PR
```
apelisse: @jennybuckley probably knows, I suspect it's because we treat deduced fields as map fields?

typed/remove_test.go
```diff
@@ -0,0 +1,133 @@
...
+			object: `{}`,
+			remove: fieldpath.NewSet(fieldpath.MakePathOrDie("a")),
+			deep:   true,
+			expect: ``,
+		},
+		{
+			object: `{"a": "value"}`,
+			remove: fieldpath.NewSet(fieldpath.MakePathOrDie("a")),
+			deep:   true,
+			expect: `null`,
```
apelisse: I guess it's a little odd how `{}` goes to `` and this one goes to `null`. Do you happen to know why?

typed/remove_test.go
```diff
@@ -0,0 +1,133 @@
...
+			object: `{"a": "value", "b": ["c"]}`,
+			remove: fieldpath.NewSet(fieldpath.MakePathOrDie("b", "c")),
+			deep:   true,
+			expect: `{"a": "value", "b": null}`,
+		},
+		{
+			object: `{"a": "value", "b": ["c", "d"]}`,
+			remove: fieldpath.NewSet(fieldpath.MakePathOrDie("b", "c")),
+			deep:   true,
+			// TODO: is this expected?
```
apelisse: Certainly not? You're not supposed to remove `d` are you?

apelisse: Good thing we have tests :-)

typed/remove_test.go
```diff
@@ -0,0 +1,133 @@
...
+			expect: `{"a": "value", "b": null}`,
+		},
+		{
+			object: `{"a": "value", "b": [{"c": "value"}, {"d": "value"}]}`,
+			remove: fieldpath.NewSet(fieldpath.MakePathOrDie("b", "c")),
+			deep:   true,
+			// TODO: is this expected?
```
apelisse: Neither

apelisse: These are probably very ugly bugs...

typed/remove_test.go
```diff
@@ -0,0 +1,133 @@
...
+			remove: fieldpath.NewSet(fieldpath.MakePathOrDie("b")),
+			deep:   false,
+			// TODO: this seems to behave differently with a non deduced schema: {"a": "value", "b":{"c":"value"}} which is the reason for this PR
+			expect: `{"a": "value"}`,
+		},
+		{
+			object: `{"a": "value", "b": {"c":"value"}}`,
+			remove: fieldpath.NewSet(fieldpath.MakePathOrDie("b", "c")),
+			deep:   true,
+			// TODO: is a null result expected (insted of a {})
```
apelisse: I agree that it's a little odd. What do you think @jennybuckley

typed/remove_test.go
```diff
@@ -0,0 +1,133 @@
...
+			remove: fieldpath.NewSet(fieldpath.MakePathOrDie("b", "d")),
+			deep:   true,
+			expect: `{"a": "value", "b": {"c":"value"}}`,
+		},
+		{
+			object: `{"a": "value", "b": {"c":"value", "d": {"e":"value"}}}`,
+			remove: fieldpath.NewSet(fieldpath.MakePathOrDie("b", "d", "e")),
+			deep:   true,
+			expect: `{"a": "value", "b": {"c":"value", "d": null}}`,
+		},
```
apelisse: Would you mind taking all these tests and running them with `deep: false`, so that we can also test that behavior?

value/value.go
kwiesmueller: It's still just the comment. But sure, I'll remove it.

typed/remove_test.go
kwiesmueller: I am not sure. I thought maybe it's not possible to remove items of a list by their data. But then this behavior still would be unexpected.

typed/remove_test.go
kwiesmueller: I think it has something to do with how the test creates the typedvalues.
So the input gets converted as expected, after removal it results in {}

Maybe the result is not null, but null converts to the same as the result of an empty object?

Not sure if the way I create the objects (input, expect) is correct.

kwiesmueller: I'll step through it a little more in the debugger to find out.

typed/remove_test.go
```diff
@@ -0,0 +1,133 @@
...
+		c := c
+		t.Run(fmt.Sprintf("case-%d", i), func(t *testing.T) {
+			t.Parallel()
+
+			obj, err := DeducedParseableType.FromYAML(c.object)
+			if err != nil {
+				t.Fatalf("unable to parse/validate object: %v\n%v", err, c.object)
+			}
+
+			expectTyped, err := DeducedParseableType.FromYAML(c.expect)
```
apelisse: These two `FromYAML` (here and below) look good to me. You might have to deep into the code a little bit more to undesrtand why you end-up with null instead of empty dir.

typed/remove_test.go
apelisse: Yeah, I commented below, I think the test code is fine, the tested code is probably not.

kwiesmueller: Okay, so i found the cause.
https://github.com/kubernetes-sigs/structured-merge-diff/pull/133/files#diff-11232f9aff4ebc184b767f3a3520214aR78
There we just return when the map (same for list) is nil or empty.
This means the resulting object after remove is nil and the expect `` matches this.

kwiesmueller: Not sure about the null one yet.

typed/remove_test.go
kwiesmueller: I'm thinking about adding a second test that defines the schema for all the tests above to verify.
If the behavior differs, we can at least be sure we want it to do.

typed/remove_test.go
kwiesmueller: same reason as: https://github.com/kubernetes-sigs/structured-merge-diff/pull/133/files#r364479060

typed/remove_test.go
jennybuckley: Yes I think this is because deduced schema would see "b":{"c":"value"} as a map item, and so it would be removable, but if you defined a schema which has the fields named, then it would be seen as a struct field, and I don't think we expect to remove those with RemoveItems, since this function was created to remove list/map items that no appliers owns anymore, a behavior which doesn't apply to struct fields (since we didn't want removing replicas from a config and reapplying to reset replicas to 1)

typed/remove_test.go
jennybuckley: fieldpath.MakePathOrDie("b", "c") does this path refer to the list item "c"? I think the path being invalid might be why this is happening, but either way the output isn't what I'd expect.

jennybuckley: If you want to reference the list item by it's data you might have to do: fieldpath.MakePathOrDie("b", value.Value{StringValue: "c"}) or something like this which would work with the new value interface

typed/remove_test.go
jennybuckley: I think this wouldn't matter because in an actual serialized kubernetes object with an empty map, it should look the same as one with a null map, right? but I'm not totally sure

typed/remove_test.go
kwiesmueller: Okay, then I'll check tomorrow why an invalid path might remove the entire thing.
But yes, that was the goal. I just wanted to try of this works and stumbled into this, so it's a test now.

typed/remove_test.go
kwiesmueller: Not sure. Don't we treat null and empty differently at least for apply?

kwiesmueller: So when we strip fields before running apply, I think resulting in null actually might be better as the key would be removed from the object and apply ignores it (right?).
Probably the reason why it is this way?

apelisse: Actually, we tend to treat null as an opinion so it's not ideal

kwiesmueller: Do we? The opinion to remove something/reset a list?
I only remember when I added the tests for resetting managedFields, that the managedFields must not be null or [] but [{}] but I think that's a different place.

apelisse: You're right, we don't on list, we should be good then.

kwiesmueller:
/retest 
to see if everything breaks now (as it should with the race test)

typed/remove_test.go
kwiesmueller: @jennybuckley the reason is https://github.com/kubernetes-sigs/structured-merge-diff/pull/133/files#diff-11232f9aff4ebc184b767f3a3520214aR49 
because in this case the ElementRelationship in the List is atomic, so the function just returns and the remove ends up with a nil output for this atom.

typed/remove.go
```diff
@@ -46,6 +48,7 @@ func (w *removingWalker) doList(t *schema.List) (errs ValidationErrors) {
 
 	// If list is null, empty, or atomic just return
 	if l == nil || l.Length() == 0 || t.ElementRelationship == schema.Atomic {
+		w.out = l
```
apelisse: This is it. `l` is a `listUnstructured` in your case. I'm a little concerned that this is not consistent with the map case, so we must have a bug there and a scenario that is mis-tested.

apelisse: This doesn't do the same thing, is this expected: https://github.com/kubernetes-sigs/structured-merge-diff/pull/133/files#diff-11232f9aff4ebc184b767f3a3520214aR80

go.mod
```diff
@@ -5,8 +5,7 @@ require gopkg.in/yaml.v2 v2.2.1
 require (
 	github.com/google/gofuzz v1.0.0
 	github.com/json-iterator/go v1.1.6
-	github.com/modern-go/concurrent v0.0.0-20180306012644-bacd9c7ef1dd // indirect
-	github.com/modern-go/reflect2 v1.0.1 // indirect
+	sigs.k8s.io/structured-merge-diff v1.0.2
```
apelisse: I don't know if you rebased or what, but you're going to have fun fixing these :-)

typed/remove.go
```diff
@@ -92,6 +95,11 @@ func (w *removingWalker) doMap(t *schema.Map) ValidationErrors {
 	m.Iterate(func(k string, val value.Value) bool {
 		pe := fieldpath.PathElement{FieldName: &k}
 		path, _ := fieldpath.MakePath(pe)
+
+		if w.toRemove.Has(path) {
+			return true
+		}
+
```
apelisse: Line 107 below, we're doing another similar test that I think you can remove.

value/value.go
```diff
@@ -214,7 +214,7 @@ func EqualsUsing(a Allocator, lhs, rhs Value) bool {
 	return true
 }
 
-// ToString returns a human-readable representation of the value.
+// String returns a human-readable representation of the value.
```
apelisse: Did you forget something?

typed/remove.go
apelisse: I'm also very surprised that you don't have to do that for lists (in case we want to remove an associative list-item for example). I guess I'll see in the tests if that's a use-case you missed.

typed/typed.go
```diff
@@ -138,7 +138,8 @@ func (tv TypedValue) Compare(rhs *TypedValue) (c *Comparison, err error) {
 	return c, nil
 }
 
-// RemoveItems removes each provided list or map item from the value.
+// RemoveItems removes each provided path from the value, including matching child paths.
```
apelisse: > including matching child paths

That part of the sentence is not super clear. I think the word recursive would definitely remind me of `rm -r` as in "removes the path and all of its sub-path"?

typed/remove_test.go
```diff
@@ -0,0 +1,340 @@
...
+	}{
+		{
+			object: `{}`,
+			remove: fieldpath.NewSet(fieldpath.MakePathOrDie("a")),
+			expect: `{}`,
+		},
+		{
+			object: `{"a": "value"}`,
+			remove: fieldpath.NewSet(fieldpath.MakePathOrDie("a")),
+			expect: ``,
```
apelisse: I'm really confused by the inconsistency here. If we have an empty object, we keep the empty object. But if we remove the single item from the object, we don't have an object anymore? :-))

typed/remove_test.go
```diff
@@ -0,0 +1,340 @@
...
+			remove: fieldpath.NewSet(fieldpath.MakePathOrDie("a")),
+			expect: `{"b": {}}`,
+		},
+		{
+			object: `{"a": "value", "b": {"c":"value"}}`,
+			remove: fieldpath.NewSet(fieldpath.MakePathOrDie("b")),
+			expect: `{"a": "value"}`,
+		},
+		{
+			// this is okay, because in a deduced schema `b` is handled as a map not a struct
```
apelisse: I don't see a difference between this test and the test above. Also I think the comment is obsolete now, since removing should just remove the entire thing anyway?

typed/remove_test.go
```diff
@@ -0,0 +1,340 @@
...
+			object: `{"a": "value", "b":{"c":"value"}}`,
+			remove: fieldpath.NewSet(fieldpath.MakePathOrDie("b")),
+			expect: `{"a": "value"}`,
+		},
+		{
+			object: `{"a": "value", "b": {"c":"value"}}`,
+			remove: fieldpath.NewSet(fieldpath.MakePathOrDie("b", "c")),
+			expect: `{"a": "value", "b": null}`,
```
apelisse: now it's not {} or empty (it cant' be empty), but null. I know we talked about it, but it's always confusing.

typed/remove_test.go
```diff
@@ -0,0 +1,340 @@
...
+		{
+			object: `{"a": "value", "b": []}`,
+			remove: fieldpath.NewSet(fieldpath.MakePathOrDie("b")),
+			expect: `{"a": "value"}`,
+		},
+		{
+			object: `{"a": "value", "b": ["c"]}`,
+			remove: fieldpath.NewSet(fieldpath.MakePathOrDie("b", "c")),
+			// TODO: this should fail
+			expect: `{"a": "value", "b": ["c"]}`,
```
apelisse: I would expect ...
```
expect: `{"a": "value", "b": []}`, 
```

typed/remove_test.go
```diff
@@ -0,0 +1,340 @@
...
+		},
+		{
+			object: `{"a": "value", "b": ["c", "d"]}`,
+			remove: fieldpath.NewSet(fieldpath.MakePathOrDie("b", "c")),
+			expect: `{"a": "value", "b": ["c", "d"]}`,
```
apelisse: I would expect ...
```
expect: `{"a": "value", "b": ["d"]}`,
```

typed/remove_test.go
```diff
@@ -0,0 +1,340 @@
...
+		},
+		{
+			object: `{"a": "value", "b": [{"c": "value"}, {"d": "value"}]}`,
+			remove: fieldpath.NewSet(fieldpath.MakePathOrDie("b", "c")),
+			expect: `{"a": "value", "b": [{"c": "value"}, {"d": "value"}]}`,
```
apelisse: I would expect
```
expect: `{"a": "value", "b": [{"d": "value"}]}`
```

typed/remove_test.go
```diff
@@ -0,0 +1,340 @@
...
+		},
+		{
+			object: `{"a": "value", "b": {"c":"value", "d":{"e":"value"}}}`,
+			remove: fieldpath.NewSet(fieldpath.MakePathOrDie("b", "d")),
+			expect: `{"a": "value", "b": {"c":"value"}}`,
```
apelisse: This one is correct!

typed/remove_test.go
```diff
@@ -0,0 +1,340 @@
...
+		},
+		{
+			object: `{"a": "value", "b": {"c":"value", "d": {"e":"value"}}}`,
+			remove: fieldpath.NewSet(fieldpath.MakePathOrDie("b", "d", "e")),
+			expect: `{"a": "value", "b": {"c":"value", "d": null}}`,
```
apelisse: I would expect ...
```
expect: `{"a": "value", "b": {"c":"value", "d": {}}}`
```

typed/remove_test.go
```diff
@@ -0,0 +1,340 @@
...
+                fields:
+                - name: name
+                  type:
+                    scalar: string
+                elementRelationship: atomic
+            elementRelationship: associative
+    elementType:
+      scalar: string
+`,
+			remove: fieldpath.NewSet(fieldpath.MakePathOrDie("b", fieldpath.PathElement{Key: &value.FieldList{{Name: "name", Value: value.NewValueInterface("c")}}})),
```
apelisse: You have a test for that, good. Not sure I agree with null, but good.

typed/remove_test.go
kwiesmueller: Not really. I expected that as well, but you can not address list items this way using makepath.
So that's actually what should happen.
To delete a list item, there is the test below.

typed/remove_test.go
kwiesmueller: That's what we have to decide.
I think we had this discussion before.
What does it mean to null and what to {}.

go.mod
kwiesmueller: I did rebased, but not a clue what happened there.
Will clean up

typed/remove_test.go
kwiesmueller: I think that's  something different, but let me verify.
As I remember it, it did not matter if expect is empty or {} because of the way this string gets parsed and compare works.
At least for deduced

typed/remove_test.go
kwiesmueller: Yeah, that one slipped when I removed all the old duplicated tests :-/

typed/remove_test.go
kwiesmueller: Yeah, I think we should discuss the {} vs null thing and put it down so we know why we do it that way (or not)
Not sure myself anymore...

typed/remove_test.go
kwiesmueller: Same as above

typed/remove_test.go
kwiesmueller: Same, the remove path has to be different for this. At least that's how Jenny explained it to me.
But I can add another test for this.

value/value.go
kwiesmueller: Shame on me

typed/remove_test.go
apelisse: Ah, yeah, then I would make the tests consistent if possible

typed/remove_test.go
apelisse: Removing an item from a directory should not "remove that directory" if it ends-up being empty. I'm sort of following the semantics of `rm -r` in my mind, and that seems wrong.

typed/remove_test.go
apelisse: Why not? what if `b` is a set, can't you remove an item from the set?

apelisse: From the "status wiping" perspective, I think we would never have this kind of use-case, so we're probably good, but still, that seems odd to me.

typed/remove_test.go
apelisse: OK, I agree that I'm wrong on this one.

apelisse: c is the key of an object in the list, it wouldn't make sense to remove "c" from all the objects in the list I think.

typed/remove_test.go
apelisse: I would keep it as `{}` rather than null tbh. Because the goal is not to change the type or meaning of the parent object. Also we do typically give "null" a stronger meaning than empty list as far as I can tell, and I don't think that's good here, even though it proably doesn't make a huge difference.

typed/remove.go
kwiesmueller: I think i missed a test case for this.

typed/remove_test.go
kwiesmueller: Okay, i'll change it and see what breaks.
Just for reference, i think that's one of the previous discussions about this: https://github.com/kubernetes-sigs/structured-merge-diff/pull/133#discussion_r364017082

typed/remove_test.go
kwiesmueller: not sure that's a remove thing, but fieldpath.
if you would like your result, you'd have to specify it like here:
https://github.com/kubernetes-sigs/structured-merge-diff/pull/133/files#diff-2243880310b5933acfd4d1204d47c1f0R284

at least that's what i understood from this comment: https://github.com/kubernetes-sigs/structured-merge-diff/pull/133#discussion_r364503785

I wrote this test initially because i expected what you did.
If we agree that expectation should be right, i'll look how, but it makes sense after reading Jenny's comment.

apelisse: I see, the `MakePath` method has a slightly weird semantic where "strings" mean "fields of map with that name". In this case, it's not a map with a field with that name, but we want to refer to a value in the list. So yes, Jenny is right (as usual). Still, we might want a test for that.

kwiesmueller:
@apelisse i pushed my current state including the failing tests.
The `TestUpdateLeaf` ones are the ones i mentioned a week ago. They fail because of pruning and the new remove.

The `TestMultipleAppliers` are new and should be caused by the different null behavior.
I am not sure yet why they fail, but am looking at it.

typed/remove.go
kwiesmueller: Added a test, but list already does this.

typed/remove_test.go
kwiesmueller: Test added below

typed/remove_test.go
```diff
@@ -0,0 +1,369 @@
...
```
apelisse: OK I'm happy you have this example. There is a better way to build this though:
```suggestion
			remove: _NS(_P("b", _KBF("name", "c"))),
```

And as you can see, you can use these function aliases to simplify things throughout this file.

Also maybe you should have more than one item and show that only this one gets removed?

typed/remove_test.go
```diff
@@ -0,0 +1,369 @@
...
+                scalar: string
+            elementRelationship: atomic
+    elementType:
+      scalar: string
+`,
+			remove: fieldpath.NewSet(fieldpath.MakePathOrDie("b")),
+			expect: `{"a": "value"}`,
+		},
+		{
+			object: `{"a": "value", "b": ["c"]}`,
```
apelisse: Explain why this one does nothing

typed/remove_test.go
```diff
@@ -0,0 +1,369 @@
...
+          list:
+            keys: 
+            - name
+            elementType:
+              map:
+                fields:
+                - name: name
+                  type:
+                    scalar: string
+                elementRelationship: atomic
```
apelisse: I don't know if that line is useful

typed/remove_test.go
apelisse: Also maybe add other fields, so that we can see that the entire things gets removed.

merge/multiple_appliers_test.go
```diff
@@ -670,7 +670,7 @@ func TestMultipleAppliersNestedType(t *testing.T) {
 			},
 			Object: `
 				mapOfMapsRecursive:
-				  a:
+				  a: {}
```
apelisse: This is a little unfortunate :-/ but I think that's fine.

internal/fixture/state.go
```diff
@@ -504,3 +508,46 @@ func (tc TestCase) TestWithConverter(parser Parser, converter merge.Converter) e
 
 	return nil
 }
+
+// PrintState is an Operation printing the current state to help with debugging tests
+type PrintState struct{}
```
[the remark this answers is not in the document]
kwiesmueller: Yes, multiple tests would solve that as well.
While creating (and debugging) the remove tests, it was already getting less comfortable working with them due to the amount of tests. Also because of my unfortunate decision not to name them...

For now I only wanted to understand and debug the broken tests that already exist.
Pulling them apart into multiple tests would have been something for another PR and then still might cause more clutter while the Print and Expect Operations would allow for the same result in the existing test fixture thing.

typed/remove_test.go
```diff
@@ -0,0 +1,369 @@
...
+			expect: `{"a": "value"}`,
+		},
+		{
+			object: `{"a": "value", "b": ["c", "d"]}`,
+			remove: fieldpath.NewSet(fieldpath.MakePathOrDie("b", "c")),
+			expect: `{"a": "value", "b": ["c", "d"]}`,
+		},
+		{
+			object: `{"a": "value", "b": ["c", "d"]}`,
+			remove: fieldpath.NewSet(fieldpath.MakePathOrDie("b", fieldpath.NewIndexPathElement(1))),
```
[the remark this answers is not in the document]
kwiesmueller: I tried a few ways. If you know a better way, i'm glad to hear it.

typed/remove_test.go
```diff
@@ -0,0 +1,457 @@
...
+	_V   = value.NewValueInterface
+	_I   = fieldpath.NewIndexPathElement
+)
+
+func TestRemoveDeduced(t *testing.T) {
+	var cases = []struct {
+		object          YAMLObject
+		remove          *fieldpath.Set
+		expect          YAMLObject
+		expectItemsOnly YAMLObject
```
apelisse: I like that you're testing both at the same, that's so much simpler :-)

typed/remove_test.go
```diff
@@ -0,0 +1,457 @@
...
+		},
+		{
+			object:          `{"a": "value", "b": ["c"]}`,
+			remove:          _NS(_P("b", _I(0))),
+			expect:          `{"a": "value", "b": []}`,
+			expectItemsOnly: `{"a": "value", "b": []}`,
+		},
+		{
+			object:          `{"a": "value", "b": ["c", "d"]}`,
+			remove:          _NS(_P("b", _I(1))),
```
apelisse: I don't see a test that removes by value, do we have one? Maybe below, but I suspect it belongs here

typed/remove_test.go
apelisse: BUT, for CRDs everything is treated as a map, so everything is an item, so I can't see any test where the two are not the same?

typed/remove_test.go
```diff
@@ -0,0 +1,457 @@
...
+			result = obj.RemoveItems(c.remove).AsValue()
+			if !value.Equals(result, expectItemsOnly) {
+				t.Fatalf("unexpected result after RemoveItems:\ngot: %v\nexp: %v",
+					value.ToString(result), value.ToString(expectItemsOnly),
+				)
+			}
+		})
+	}
+}
+func TestRemove(t *testing.T) {
```
apelisse: missing space here before function :D

typed/remove_test.go
```diff
@@ -0,0 +1,457 @@
...
+              scalar: string
+        `,
+			remove:          _NS(_P("b")),
+			expect:          `{"a": "value"}`,
+			expectItemsOnly: `{"a": "value", "b":{"c":"value"}}`,
+		},
+		{
+			object: `{"a": "value", "b":{"c":"value"}}`,
+			schema: `types:
+- name: type
```
apelisse: somehow this one is not indented the same way as the other

apelisse: it's not the only one either.

typed/remove_test.go
```diff
@@ -0,0 +1,457 @@
...
+            - name: c
+              type:
+                scalar: string
+            elementRelationship: separable
+    elementType:
+      scalar: string
+`,
+			remove: _NS(_P("b", "c")),
+			expect: `{"a": "value", "b": {}}`,
+			// This seems wrong
```
apelisse: What is wrong? That doesn't look bad to me?

typed/remove_test.go
```diff
@@ -0,0 +1,457 @@
...
+                  scalar: string
+              - name: b
+                type:
+                  map:
+                    fields:
+                    - name: c
+                      type:
+                        scalar: string
+                    elementRelationship: separable
+            elementType:
```
apelisse: I'm confused why you have this here?

apelisse: I can see how it can be useful but you're not using it here.

typed/remove_test.go
```diff
@@ -0,0 +1,457 @@
...
+            - name: c
+              type:
+                scalar: string
+            elementRelationship: atomic
+    elementType:
+      scalar: string
+`,
+			remove: _NS(_P("b", "c")),
+			expect: `{"a": "value", "b": {}}`,
+			// This seems wrong
```
apelisse: I don't see it either

typed/remove_test.go
```diff
@@ -0,0 +1,457 @@
...
+                    elementRelationship: atomic
+            elementType:
+              scalar: string
+        `,
+			remove: _NS(_P("b", "c")),
+			// `c` won't get removed, as it is not a field, while being addressed as one in the path.
+			expect:          `{"a": "value", "b": ["c"]}`,
+			expectItemsOnly: `{"a": "value", "b": ["c"]}`,
+		},
+	}
```
apelisse: You still don't have a test that removes from a set.

typed/remove_test.go
kwiesmueller: Yeah, missed that.
I commented the other tests for debugging only broken ones and removing the comments messed up indendation.

typed/remove_test.go
kwiesmueller: Right. Previously there was the "skip atomic" part. Would that make it compatible with crds?

typed/remove.go
```diff
@@ -46,8 +48,10 @@ func (w *removingWalker) doScalar(t *schema.Scalar) ValidationErrors {
 func (w *removingWalker) doList(t *schema.List) (errs ValidationErrors) {
 	l := w.value.AsListUsing(w.allocator)
 	defer w.allocator.Free(l)
-	// If list is null, empty, or atomic just return
-	if l == nil || l.Length() == 0 || t.ElementRelationship == schema.Atomic {
```
apelisse: @jennybuckley Do you remember why we had the atomic check here? Would it be a problem to remove it?

internal/fixture/state.go
```diff
@@ -504,3 +504,48 @@ func (tc TestCase) TestWithConverter(parser Parser, converter merge.Converter) e
...
+		return fmt.Errorf("expected live and config to be the same:\n%v\nConfig: %v\nLive:   %v\n", comparison, value.ToString(config.AsValue()), value.ToString(state.Live.AsValue()))
+	}
+	return nil
+}
+
+func (op ExpectState) preprocess(parser Parser) (Operation, error) {
+	return op, nil
+}
+
+func (s *State) parseConfig(apiVersion fieldpath.APIVersion, obj typed.YAMLObject) (*typed.TypedValue, error) {
```
apelisse: nit, why is this method lost here?

kwiesmueller:
@jennybuckley did you have time to look at this?

apelisse:
Can you please squash? Thanks

fieldpath/element.go
```diff
@@ -47,6 +47,16 @@ type PathElement struct {
 	Index *int
 }
 
+// NewValuePathElement creates a PathElement from the provided value.Value
```
apelisse, line 50: Can you remove these two functions if we don't need them? I suspect we already have something like this in `MakePath`

apelisse:
I think that's been replaced.

PR #134, opened 2019-12-10
Update everything to v2

apelisse:
Backward compatibility has definitely changed, so I think it's time that we make it v2.

### DIFF
--- a/fieldpath/element.go
+++ b/fieldpath/element.go
@@ -21,7 +21,7 @@ import (
 	"sort"
 	"strings"
 
-	"sigs.k8s.io/structured-merge-diff/value"
+	"sigs.k8s.io/structured-merge-diff/v2/value"
 )
 
 // PathElement describes how to select a child field given a containing object.
--- a/fieldpath/element_test.go
+++ b/fieldpath/element_test.go
@@ -19,7 +19,7 @@ package fieldpath
 import (
 	"testing"
 
-	"sigs.k8s.io/structured-merge-diff/value"
+	"sigs.k8s.io/structured-merge-diff/v2/value"
 )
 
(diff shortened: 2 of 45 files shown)


jpbetz:
The mechanical change looks good.

Are we setting precedence for using v2 go modules in k8s? Or has the been done before for any of the other sub-projects? I haven't had the time to think through how version tags would work, any thing we can reference or add to this PR about that might be valuable.

apelisse:
This follows the recommandation from: https://github.com/golang/go/wiki/Modules#semantic-import-versioning

liggitt:
The go.mod file module name needs /v2 appended. The fact that tests passed without that is concerning as well… it means the code in this PR isn't what CI is testing.

apelisse:
Wait, it has the name appended, I'm not sure what you're talking about ...

liggitt:
Weird, I totally missed that somehow.

liggitt:
Agree the mechanics look fine. I'm not quite sure when we would want to add v2.x.y tags. We may want to tag this commit with v2.0.0-alpha.0 or something similar so the first commit with a v2 go.mod has a v2 tag.

liggitt:
Separately, it would also be good to document / formalize how we determined go API compatibility had been broken and a major revision was required. An automated test that uses [apidiff](https://github.com/golang/exp/blob/master/cmd/apidiff/main.go) to record the public API of the last tagged release in a file, and a CI job that flags breakages against it would be really useful in avoiding accidental API breakages.

apelisse:
Absolutely a great idea!

With regard to tagging, we can't tag until the commit is merged, would you mind approving?

apelisse:
Thanks!

PR #135, opened 2019-12-11
Change equal for path element

apelisse:
This brings a 5 to 10% time-wise on some of the operations. Might be tremendous improvements on some specific resources though.

### DIFF
--- a/fieldpath/element.go
+++ b/fieldpath/element.go
@@ -49,50 +49,90 @@ type PathElement struct {
 
 // Less provides an order for path elements.
 func (e PathElement) Less(rhs PathElement) bool {
+	return e.Compare(rhs) < 0
+}
+
+// Compare provides an order for path elements.
+func (e PathElement) Compare(rhs PathElement) int {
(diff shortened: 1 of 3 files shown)


fieldpath/element.go
```diff
@@ -49,50 +49,90 @@ type PathElement struct {
...
 	}
 
-	return false
+	return 0
 }
 
 // Equals returns true if both path elements are equal.
 func (e PathElement) Equals(rhs PathElement) bool {
-	return !e.Less(rhs) && !rhs.Less(e)
+	if e.FieldName != nil {
```
jpbetz: For readability, could use this idiom for taking care of the nil cases up-front:
```
if (e.FieldName == nil) != (rhs.FieldName == nil) {
  return false
}
```

Just a suggestion. The logic already appears correct.

apelisse: Maybe it's just me, but I had to squint to make sure that statement was correct, I somehow find it harder to read?

apelisse:
yay

PR #136, opened 2019-12-18
Add reflector based implementation of value interface

jpbetz:
Part 1 of the reflector based value interface implementation. This does not include all of the improvements from https://github.com/jennybuckley/structured-merge-diff/pull/1, but I'd like to get it reviewed and merged before sending out PRs for the other improvements.

cc @jennybuckley @apelisse


value/mapreflect.go
```diff
@@ -0,0 +1,105 @@
...
+		return true
+	})
+	return result
+}
+
+func (r mapReflect) Equals(m Map) bool {
+	if r.Length() != m.Length() {
+		return false
+	}
+	return m.Iterate(func(key string, value Value) bool {
```
apelisse: Since you know the type of `r`, I think you can do better than the closure here? Only important if the benchmarks says it is...

value/mapreflect.go
```diff
@@ -0,0 +1,105 @@
...
+
+func (r mapReflect) Iterate(fn func(string, Value) bool) bool {
+	return eachMapEntry(r.Value, func(s string, value reflect.Value) bool {
+		mapVal := mustWrapValueReflect(value)
+		defer mapVal.Recycle()
+		return fn(s, mapVal)
+	})
+}
+
+func eachMapEntry(val reflect.Value, fn func(string, reflect.Value) bool) bool {
```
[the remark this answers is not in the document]
jpbetz: This is an interesting idea. We tried it out (https://github.com/jpbetz/structured-merge-diff/tree/value-range-iterator) and it makes the value reflector perform better, but it’s worse for value unstructured maps. The experimented used a `valuemap.Range()` function that returns an interface with `.Key()`, `.Value()` and `.Next()` functions.  I suspect this it might be able to turn this experiment into a net win somehow, but I'm shelving in the short term while we land some of the more obvious performance wins.

value/valuereflect.go
```diff
@@ -0,0 +1,301 @@
...
+func (r valueReflect) List() List {
+	if r.IsList() {
+		return listReflect{r.Value}
+	}
+	panic("value is not a list")
+}
+
+func (r valueReflect) Bool() bool {
+	if r.IsBool() {
+		return r.Value.Bool()
```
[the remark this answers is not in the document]
jpbetz: Yes, looks like I could drop this check for `Bool()` and `Float()`. I'm leaving them in for consistency with `String()`, `List()` and `Int()`, but agree that they're not strictly needed.

value/mapreflect.go
jpbetz: It haven't yet it seen it as a major contributor in benchmarks, but I'll add a comment about this potential optimization and circle back on it. It could certainly be more efficient.

jpbetz:
> It'd be really really nice if we could have some structs and decode them so that we can have benchmarks for this code directly in smd rather than through kubernetes/kubernetes.

Added a benchmark in smd that performs the validate step, which is fairly representative of the benchmark we have in kubernetes/kubernetes.

jpbetz:
Rebased.

value/mapreflect.go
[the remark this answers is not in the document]
apelisse: > I suspect this it might be able to turn this experiment into a net win somehow, but I'm shelving in the short term while we land some of the more obvious performance wins.

That sounds great.

apelisse:
/lgtm

Thanks for doing it Joe!
@lavalamp do you want to have a look that?

value/jsontagutil.go
```diff
@@ -0,0 +1,89 @@
...
+
+import (
+	"reflect"
+	"strings"
+)
+
+// TODO: This implements the same functionality as https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apimachinery/pkg/runtime/converter.go#L236
+// but is based on the highly efficient approach from https://golang.org/src/encoding/json/encode.go
+
+func lookupJsonTags(f reflect.StructField) (string, bool, bool) {
```
lavalamp: Since these won't ever change, it might make sense to build a shared global cache in a map?

value/jsontagutil.go
```diff
@@ -0,0 +1,89 @@
...
+	case reflect.Bool:
+		return !v.Bool()
+	case reflect.Int, reflect.Int8, reflect.Int16, reflect.Int32, reflect.Int64:
+		return v.Int() == 0
+	case reflect.Uint, reflect.Uint8, reflect.Uint16, reflect.Uint32, reflect.Uint64, reflect.Uintptr:
+		return v.Uint() == 0
+	case reflect.Float32, reflect.Float64:
+		return v.Float() == 0
+	case reflect.Interface, reflect.Ptr:
+		return v.IsNil()
```
lavalamp: Consider panic on funcs and channels?

value/jsontagutil.go
```diff
@@ -0,0 +1,89 @@
...
+	"strings"
+)
+
+// TODO: This implements the same functionality as https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apimachinery/pkg/runtime/converter.go#L236
+// but is based on the highly efficient approach from https://golang.org/src/encoding/json/encode.go
+
+func lookupJsonTags(f reflect.StructField) (string, bool, bool) {
+	var name string
+	tag := f.Tag.Get("json")
+	if tag == "-" {
```
lavalamp: We should probably pretend such fields don't exist at all.

value/jsontagutil.go
lavalamp: Oh, it looks like you already did.

value/structreflect.go
```diff
@@ -0,0 +1,238 @@
...
+func (c *structCache) Get(t reflect.Type) (map[string]*fieldCacheEntry, bool) {
+	entry, ok := c.value.Load().(structCacheMap)[t]
+	return entry, ok
+}
+
+// Update sets the fieldCacheEntry for the given type via a copy-on-write update to the struct cache.
+func (c *structCache) Update(t reflect.Type, m map[string]*fieldCacheEntry) {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+
```
lavalamp: Possible optimization: check if `t` is already present (due to multiple simultaneous computations when we lose the race) and bail, since the build function is idempotent (right?). Might make a difference on startup.

value/structreflect.go
```diff
@@ -0,0 +1,238 @@
...
+	return ok
+}
+
+func (r structReflect) Set(key string, val Value) {
+	fieldVal, ok := r.findJsonNameField(key)
+	if !ok {
+		panic(fmt.Sprintf("key %s may not be set on struct %T: field does not exist", key, r.Value.Interface()))
+	}
+	if !fieldVal.CanSet() {
+		// See https://blog.golang.org/laws-of-reflection for details on why a struct may not be settable
```
lavalamp: In other reflection stuff, we require that a user pass a pointer for the top level object.

The other case, a struct as a map item, we can actually handle (with a lot of complexity) by storing the map and the key and copying and re-setting the whole item. Hopefully we don't need to do that.

I'm not sure if panic is going to be OK here unless we're sure it's not going to happen at run time?

value/structreflect.go
```diff
@@ -0,0 +1,238 @@
...
+func (r structReflect) Delete(key string) {
+	fieldVal, ok := r.findJsonNameField(key)
+	if !ok {
+		panic(fmt.Sprintf("key %s may not be deleted on struct %T: field does not exist", key, r.Value.Interface()))
+	}
+	if !fieldVal.CanSet() {
+		// See https://blog.golang.org/laws-of-reflection for details on why a struct may not be settable
+		panic(fmt.Sprintf("key %s may not be deleted on struct: %T: struct is not settable", key, r.Value.Interface()))
+	}
+	fieldVal.Set(reflect.Zero(fieldVal.Type()))
```
lavalamp: I think we can only properly "delete" something if we're setting a pointer to nil. If we're setting a e.g. string to "" or an int to 0, I don't think that's doing what the caller expects.

value/structreflect.go
```diff
@@ -0,0 +1,238 @@
...
+		v := mustWrapValueReflect(value)
+		defer v.Recycle()
+		return fn(s, v)
+	})
+}
+
+func eachStructField(structVal reflect.Value, fn func(string, reflect.Value) bool) bool {
+	for jsonName, fieldCacheEntry := range getStructCacheEntry(structVal.Type()) {
+		fieldVal := fieldCacheEntry.getFieldFromStruct(structVal)
+		if fieldCacheEntry.isOmitEmpty && (safeIsNil(fieldVal) || isZero(fieldVal)) {
```
lavalamp: I think callers expect to visit every field in the *schema* whether or not it is set in the given object? @apelisse @jennybuckley

value/structreflect.go
```diff
@@ -0,0 +1,238 @@
...
+	return eachStructField(r.Value, func(s string, value reflect.Value) bool {
+		v := mustWrapValueReflect(value)
+		defer v.Recycle()
+		return fn(s, v)
+	})
+}
+
+func eachStructField(structVal reflect.Value, fn func(string, reflect.Value) bool) bool {
+	for jsonName, fieldCacheEntry := range getStructCacheEntry(structVal.Type()) {
+		fieldVal := fieldCacheEntry.getFieldFromStruct(structVal)
```
lavalamp: Will this skip fields with `json:"-"`? I think it should?

value/valuereflect.go
```diff
@@ -0,0 +1,301 @@
...
+		return NewValueInterface(nil), nil
+	}
+	return wrapValueReflect(reflect.ValueOf(value))
+}
+
+func wrapValueReflect(value reflect.Value) (Value, error) {
+	// TODO: conversion of json.Marshaller interface types is expensive. This can be mostly optimized away by
+	// introducing conversion functions that do not require going through JSON and using those here.
+	if marshaler, ok := getMarshaler(value); ok {
+		return toUnstructured(marshaler, value)
```
lavalamp: Will this actually be used often?

value/structreflect.go
apelisse: No I don't think so. In the unstructured world we unmarshal into a map[string]interface{} and only know about the specified fields.

typed/validate_test.go
```diff
@@ -288,3 +288,98 @@ func TestSchemaSchema(t *testing.T) {
 		t.Fatalf("failed to create schemaschema: %v", err)
 	}
 }
+
+func BenchmarkValidateStructured(b *testing.B) {
+	type Primitives struct {
```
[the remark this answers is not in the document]
jpbetz: Yes, we'll have benchmarks covering k8s types in k8s/k8s. I've got those, and we'll get them merged after this PR.

value/structreflect.go
jpbetz: Nice, yes, that might help, and certainly won't hurt. The map entry puts are idempotent. Added.

value/structreflect.go
jpbetz: The value interfaces are read-only with the exception of Map,  which is currently call exclusively by `discriminator.Set` for noramlization. This set function feels out-of-place to me for this interface. I'll sort out what we need to do about this case tomorrow.
^ @apelisse

value/structreflect.go
jpbetz: I'll enforce that the value is ptr for now. This has the same issue you mentioned above `Set`, so we'll look at both cases together.

value/structreflect.go
jpbetz: Fixed and added a test.

value/structreflect.go
jpbetz: That was my understanding as well.

In a subsequent PR, I'll be merging a test that I have in k8s/k8s to go through all native kubernetes types, fuzz them, wrap them with this reflect value interface, convert them to unstructured and wrap them with the unstructured value interface and make sure they match. That's passing for me locally.

value/valuereflect.go
jpbetz: Yeah, Time and Quantity are common enough that this gets called quite a bit. I have a follow up PR to optimize those conversions more since round tripping through json is expensive.

value/jsontagutil.go
jpbetz: Fixed.

value/structreflect.go
jpbetz: Humm, actually, for omitempty fields, setting a non-pointer to zero value deletes as a caller would expect, I'll allow that case.

value/structreflect.go
jpbetz: Looks like we do always expect a pointer for the top level object. I'm going add some checks to enforce that, but looks like we're okay.

jpbetz: Added a `TestReflectMutateNestedStruct` test to cover these cases. The tests failed when a struct being changed is contained in a map, so I've created a workaround that puts a replacement into the map.

jpbetz:
Thanks for the review @lavalamp! Feedback addressed with new commits. This is ready for another pass.

value/valuereflect_test.go
```diff
@@ -0,0 +1,520 @@
...
+
+	expectedMap := map[string]interface{}{"key1": int64(2), "key4": "string4"}
+	unstructured := rv.Unstructured()
+	if !reflect.DeepEqual(unstructured, expectedMap) {
+		t.Errorf("expected %v but got: %v", expectedMap, unstructured)
+	}
+}
+
+// TestReflectMutateNestedStruct ensures a structs nested within various typed can be modified.
+func TestReflectMutateNestedStruct(t *testing.T) {
```
apelisse: That looks pretty exhaustive, thank you for writing this test. It's probably not needed, but I'm curious if we could write this test slightly more generically ... It doesn't have to be done now though.

internal/fixture/state.go
```diff
@@ -100,7 +100,7 @@ func (s *State) UpdateObject(tv *typed.TypedValue, version fieldpath.APIVersion,
 	return nil
 }
 
-// Update the current state with the passed in object
+// Set the current state with the passed in object
```
apelisse: You've updated the comment but it doesn't match the name of the function anymore?

internal/fixture/state.go
```diff
@@ -308,7 +308,7 @@ func (f ForceApplyObject) preprocess(parser typed.ParseableType) (Operation, err
 	return f, nil
 }
 
-// Update is a type of operation. It is a controller type of
+// Set is a type of operation. It is a controller type of
```
apelisse: Same, the comment doesn't match the name anymore?

internal/fixture/state.go
apelisse: I don't think this should be renamed actually.

internal/fixture/state.go
apelisse: I don't think this should be renamed actually.

merge/real_test.go
```diff
@@ -101,7 +101,7 @@ func BenchmarkOperations(b *testing.B) {
 					},
 				},
 				{
-					name: "Update",
+					name: "Set",
```
apelisse: I don't think this should be renamed actually.

merge/update.go
```diff
@@ -114,7 +114,7 @@ func (s *Updater) update(oldObject, newObject *typed.TypedValue, version fieldpa
 	return managers, compare, nil
 }
 
-// Update is the method you should call once you've merged your final
+// Set is the method you should call once you've merged your final
```
apelisse: Shouldn't be renamed

typed/parser.go
```diff
@@ -102,12 +102,28 @@ func (p ParseableType) FromYAML(object YAMLObject) (*TypedValue, error) {
...
+// unstructured object in Kubernetes world, to a TypedValue. It returns an
 // error if the resulting object fails schema validation.
+// The provided interface{} must be one of: map[string]interface{},
+// map[interface{}]interface{}, []interface{}, int types, float types,
+// string or boolean. Nested interface{} must also be one of these types.
 func (p ParseableType) FromUnstructured(in interface{}) (*TypedValue, error) {
 	return AsTyped(value.NewValueInterface(in), p.Schema, p.TypeRef)
 }
 
+// FromStructured converts a go "interface{}" type, typically an structured object in
```
apelisse: Should you make it clear in the doc that this should be a pointer?

typed/union.go
```diff
@@ -243,7 +243,7 @@ func (u *union) Normalize(old, new, out value.Map) error {
 		return fmt.Errorf("multiple fields set without discriminator change: %v", ns)
 	}
 
-	// Update discriminiator if it needs to be deduced.
+	// Set discriminiator if it needs to be deduced.
```
apelisse: :-) Your script went crazy didn't it?

typed/union.go
```diff
@@ -265,7 +265,7 @@ func (u *union) NormalizeApply(applied, merged, out value.Map) error {
 		return fmt.Errorf("applied discriminator (%v) doesn't match applied field (%v)", u.d.Get(applied), *as.One())
 	}
 
-	// Update discriminiator if needed
+	// Set discriminiator if needed
```
apelisse: Same

typed/union.go
jpbetz: Heh. Sigh. I think I did a editor based refactor when I saw it had been change to Set, but then when I merged it was no longer named Set :P

typed/parser.go
jpbetz: Yes, I missed this one. Fixed by applying same text I had added to NewValueInterface about pointers.

value/valuereflect_test.go
jpbetz: I went ahead and generalized it now.

value/fields.go
```diff
@@ -27,7 +27,7 @@ type Field struct {
 	Value Value
 }
 
-// FieldList is a list of key-value pairs. Each field is expected to
+// FieldList is a list of key-value pairs. Each field is expectUpdated to
```
apelisse: :D

jpbetz: Fixed in followup

PR #137, opened 2019-12-23
Use Compare when serializing rather than Less

apelisse:
Compare the PathElement rather than calling twice in both direction. I don't think this has a "exponential" impact as it did in other places, but it's still the best thing to do.

I get a 2.5% time improvement on serialize tests. Nothing fancy.

@tedyu @jennybuckley


apelisse:
@jpbetz

jpbetz:
Makes sense.

/lgtm

PR #138, opened 2019-12-24
Change import for release-v2 branch to v2

apelisse:


apelisse:
@liggitt @kwiesmueller

PR #139, opened 2019-12-24
Since release-2.0 is now v2, master is becoming v3

apelisse:


apelisse:
@liggitt @kwiesmueller

liggitt:
might be helpful to explain what is going on in the PR description

iiuc, https://github.com/kubernetes-sigs/structured-merge-diff/pull/134 was too late, because kubernetes 1.17 had already shipped relying on a SMD go API which was incompatible both with the SMD v1 *and* SMD master at the time https://github.com/kubernetes-sigs/structured-merge-diff/pull/134 was merged

that means we want to:
* tag a v2 release representing the SMD go API kubernetes 1.17 shipped depending on
* create a v3 stream representing the state of master

is that right?

kwiesmueller:
Yes, that's right.

PR #140, opened 2019-12-24
Back-port Fix from Jenny for ugly errors

apelisse:
The validation error is quite ugly for users, I think back-porting this to the branch for 1.16 is reasonable.

/assign @kwiesmueller @jennybuckley


apelisse:
I want to clarify the pull-request name and branch name, we're back-porting Jenny's PR to fix the ugly error in #131

apelisse:
We'll also probably need one for release-2.0, once #138 is merged.

PR #142, opened 2020-01-08
rename value conversion methods

kwiesmueller:
Renames all conversion methods for value.Value
e.g. `Map()` -> `AsMap()`

This removes the implementation of `String() string`.
Without this change, calling String() on a value would behave differently than usually expected in go.

### DIFF
--- a/value/value.go
+++ b/value/value.go
@@ -52,30 +52,30 @@ type Value interface {
 	// IsMap returns true if the Value is null, false otherwise.
 	IsNull() bool
 
-	// Map converts the Value into a Map (or panic if the type
+	// AsMap converts the Value into a Map (or panic if the type
 	// doesn't allow it).
-	Map() Map
-	// List converts the Value into a List (or panic if the type
+	AsMap() Map
+	// AsList converts the Value into a List (or panic if the type
 	// doesn't allow it).
-	List() List
-	// Bool converts the Value into a bool (or panic if the type
+	AsList() List
+	// AsBool converts the Value into a bool (or panic if the type
 	// doesn't allow it).
-	Bool() bool
-	// Int converts the Value into an int64 (or panic if the type
+	AsBool() bool
+	// AsInt converts the Value into an int64 (or panic if the type
 	// doesn't allow it).
-	Int() int64
-	// Float converts the Value into a float64 (or panic if the type
+	AsInt() int64
+	// AsFloat converts the Value into a float64 (or panic if the type
 	// doesn't allow it).
-	Float() float64
-	// String converts the Value into a string (or panic if the type
+	AsFloat() float64
+	// AsString converts the Value into a string (or panic if the type
 	// doesn't allow it).
-	String() string
+	AsString() string
 
-	// Returns a value of this type that is no longer needed. The
+	// Recycle returns a value of this type that is no longer needed. The
 	// value shouldn't be used after this call.
 	Recycle()
 
-	// Converts the Value into an Unstructured interface{}.
+	// Unstructured converts the Value into an Unstructured interface{}.
 	Unstructured() interface{}
 }
 
@@ -131,17 +131,17 @@ func Equals(lhs, rhs Value) bool {
 	if lhs.IsFloat() || rhs.IsFloat() {
 		var lf float64
 		if lhs.IsFloat() {
-			lf = lhs.Float()
+			lf = lhs.AsFloat()
 		} else if lhs.IsInt() {
-			lf = float64(lhs.Int())
+			lf = float64(lhs.AsInt())
 		} else {
 			return false
 		}
 		var rf float64
 		if rhs.IsFloat() {
-			rf = rhs.Float()
+			rf = rhs.AsFloat()
 		} else if rhs.IsInt() {
-			rf = float64(rhs.Int())
+			rf = float64(rhs.AsInt())
 		} else {
 			return false
 		}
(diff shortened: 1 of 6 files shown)


fieldpath/element.go
```diff
@@ -143,12 +143,12 @@ func (e PathElement) String() string {
 	case e.Key != nil:
 		strs := make([]string, len(*e.Key))
 		for i, k := range *e.Key {
-			strs[i] = fmt.Sprintf("%v=%v", k.Name, value.ToString(k.Value))
+			strs[i] = fmt.Sprintf("%v=%v", k.Name, value.AsString(k.Value))
```
apelisse: Here I think it's exactly NOT what we want. We don't want to cast the value into a string, but to print the value.

fieldpath/element.go
```diff
@@ -143,12 +143,12 @@ func (e PathElement) String() string {
...
 		}
 		// Keys are supposed to be sorted.
 		return "[" + strings.Join(strs, ",") + "]"
 	case e.Value != nil:
-		return fmt.Sprintf("[=%v]", value.ToString(*e.Value))
+		return fmt.Sprintf("[=%v]", value.AsString(*e.Value))
```
apelisse: Same as above

internal/fixture/state.go
```diff
@@ -450,7 +450,7 @@ func (tc TestCase) TestWithConverter(parser typed.ParseableType, converter merge
 			return fmt.Errorf("failed to compare live with config: %v", err)
 		}
 		if !comparison.IsSame() {
-			return fmt.Errorf("expected live and config to be the same:\n%v\nConfig: %v\n", comparison, value.ToString(state.Live.AsValue()))
+			return fmt.Errorf("expected live and config to be the same:\n%v\nConfig: %v\n", comparison, value.AsString(state.Live.AsValue()))
```
apelisse: Same as above I think

merge/multiple_appliers_test.go
```diff
@@ -985,7 +985,7 @@ func (r repeatingConverter) Convert(v *typed.TypedValue, version fieldpath.APIVe
 	if err != nil {
 		return nil, missingVersionError
 	}
-	y, err := yaml.Marshal(v.AsValue().Unstructured())
+	y, err := yaml.Marshal(v.AsValue().AsUnstructured())
```
apelisse: That's not really a case either, that's slightly different, I'd keep it as `Unstructured`

typed/deduced_test.go
```diff
@@ -203,7 +203,7 @@ func TestMergeDeduced(t *testing.T) {
 			} else {
 				if !value.Equals(got.AsValue(), out.AsValue()) {
 					t.Errorf("Expected\n%v\nbut got\n%v\n",
-						value.ToString(out.AsValue()), value.ToString(got.AsValue()),
+						value.AsString(out.AsValue()), value.AsString(got.AsValue()),
```
apelisse: Same

typed/merge.go
```diff
@@ -58,10 +58,10 @@ type mergeRule func(w *mergingWalker)
 var (
 	ruleKeepRHS = mergeRule(func(w *mergingWalker) {
 		if w.rhs != nil {
-			v := w.rhs.Unstructured()
+			v := w.rhs.AsUnstructured()
 			w.out = &v
 		} else if w.lhs != nil {
-			v := w.lhs.Unstructured()
+			v := w.lhs.AsUnstructured()
```
apelisse: Same

typed/merge_test.go
```diff
@@ -399,7 +399,7 @@ func (tt mergeTestCase) test(t *testing.T) {
 			} else {
 				if !value.Equals(got.AsValue(), out.AsValue()) {
 					t.Errorf("Expected\n%v\nbut got\n%v\n",
-						value.ToString(out.AsValue()), value.ToString(got.AsValue()),
+						value.AsString(out.AsValue()), value.AsString(got.AsValue()),
```
apelisse: Same

value/value.go
```diff
@@ -52,31 +52,31 @@ type Value interface {
...
+	AsInt() int64
+	// AsFloat converts the Value into a float64 (or panic if the type
 	// doesn't allow it).
-	Float() float64
-	// String converts the Value into a string (or panic if the type
+	AsFloat() float64
+	// AsString converts the Value into a string (or panic if the type
 	// doesn't allow it).
-	String() string
+	AsString() string
```
apelisse: All of these (above) are good

value/value.go
```diff
@@ -52,31 +52,31 @@ type Value interface {
...
 
 	// Returns a value of this type that is no longer needed. The
 	// value shouldn't be used after this call.
 	Recycle()
 
-	// Converts the Value into an Unstructured interface{}.
-	Unstructured() interface{}
+	// AsUnstructured converts the Value into an Unstructured interface{}.
+	AsUnstructured() interface{}
```
apelisse: I think you should revert that change

value/value.go
```diff
@@ -187,30 +187,30 @@ func Equals(lhs, rhs Value) bool {
 	return true
 }
 
-// String returns a human-readable representation of the value.
-func ToString(v Value) string {
+// AsString returns a human-readable representation of the value.
+func AsString(v Value) string {
```
apelisse: I'd also revert that change (can you make it `String`)?

fieldpath/element.go
kwiesmueller: Right, my bad.
value.ToString actually does something different than AsString.

This would be fitting the Stringer interface.
Shouldn't it be a String method for the Value interface or do you think having `AsString()` and `String()` would be error prone as it's to easy to use the wrong one?

apelisse: The problem is that `String` does not depend on the specific implementation of the string, so we can't make it part of the interface (or we could, but it'd be tedious).

kwiesmueller: Ah, okay.
Then i'll just revert the two things you pointed out.

kwiesmueller: Done

PR #145, opened 2020-01-13
Make fieldset v1 serialization alphabetical

jennybuckley:
Move '.' to beginning of fieldset, so it serialized in the same order as default json serialization.


internal/testdata/endpointsset.json
```diff
@@ -1 +1 @@
-{"f:apiVersion":{},"f:kind":{},"f:metadata":{"f:creationTimestamp":{},"f:labels":{"f:app":{},".":{}},"f:name":{},"f:namespace":{},"f:resourceVersion":{},"f:selfLink":{},"f:uid":{},".":{}},"f:subsets":{}}
```
apelisse: Maybe we should pretty print these so that it's easier to review. The review of API changes like this are super important.

jennybuckley: should I add a flag to the cli to pretty print, then?

lavalamp: We probably should, it can come in another PR though.

This particular change is unusual in that all prior output is still accepted, but it changes the output. We should probably include a test that the previous data remains valid.

PR #146, opened 2020-01-14
Allow tests to specify objects in different versions

apelisse:
This will make it possible to parse input objects with different
schemas, and test what happens if the schema changes, especially the
topology.

A few things need to happen in order to make this possible:

1. We need to be able to parse multiple different versions. A
ParseableType can only parse one type, so we need to pass in a parser
instead. The problem with the parser was that we needed a mapping to
know what was the name of the root type. Let's use the version name as
the root type. Now we can have multiple parseable type in the test.

2. Since most test actually use the same type with multiple different
version, we add a wrapper that let's you re-use the same type with
different versions.

3. DeducedType are usually all parsed the same for all versions, so
let's also create a helper for that.

4. Now we need to know which version of the object we want to compare to
at the end. We need to specify that version in each individual test, but
that's better becasue we don't have to necessarily compare in the
version of the last operation (implicit becomes explicit).

### DIFF
--- a/internal/fixture/state.go
+++ b/internal/fixture/state.go
@@ -26,11 +26,39 @@ import (
 	"sigs.k8s.io/structured-merge-diff/v3/value"
 )
 
+// For the sake of tests, a parser is something that can retrieve a
+// ParseableType.
+type Parser interface {
+	Type(string) typed.ParseableType
+}
+
+// SameVersionParser can be used if all the versions are actually using the same type.
+type SameVersionParser struct {
+	T typed.ParseableType
+}
+
+func (p SameVersionParser) Type(_ string) typed.ParseableType {
+	return p.T
+}
+
+// DeducedParser is a parser that is deduced no matter what the version
+// specified.
+var DeducedParser = SameVersionParser{
+	T: typed.DeducedParseableType,
+}
+
 // State of the current test in terms of live object. One can check at
 // any time that Live and Managers match the expectations.
+//
+// The parser will look for the type by using the APIVersion of the
+// object it's trying to parse. If trying to parse a "v1" object, a
+// corresponding "v1" type should exist in the schema. If all the
+// versions should map to the same type, or to a DeducedParseableType,
+// one can use the SameVersionParser or the DeducedParser types defined
+// in this package.
 type State struct {
 	Live     *typed.TypedValue
-	Parser   typed.ParseableType
+	Parser   Parser
 	Managers fieldpath.ManagedFields
 	Updater  *merge.Updater
 }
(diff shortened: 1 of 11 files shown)


merge/real_test.go
```diff
@@ -105,44 +106,44 @@ func BenchmarkOperations(b *testing.B) {
...
-						},
-						Update{
-							Manager:    "other-controller",
-							APIVersion: "v2",
+							APIVersion: fieldpath.APIVersion(bench.typename),
 							Object:     bench.obj,
 						},
 					},
 				},
+				// XXX: How do we do that?
```
apelisse: I need to think about that a little bit

apelisse: This test doesn't have a lot of value, maybe we should just drop it ...

apelisse:
I've been a compatibility test example on top of this PR: https://github.com/kubernetes-sigs/structured-merge-diff/pull/147/commits/6cfef667f12a2b60cb7545c177aee731595e87ea

internal/fixture/state.go
```diff
@@ -364,6 +389,9 @@ type TestCase struct {
 	// Object, if not empty, is the object as it's expected to
 	// be after all the operations are run.
 	Object typed.YAMLObject
+	// APIVersion should be set if the object is non-empty and
```
lavalamp: The diff would be smaller if this defaulted to v1 or something.

I think the comment should explain that for the tests, we implemented this as selecting the root object in the schema? Some comment should at least explain that.

apelisse: I'm really not a big fan of using a default. I considered maybe "default to the version of the latest change", which is what we implicitly used to do.

> I think the comment should explain that for the tests, we implemented this as selecting the root object in the schema? Some comment should at least explain that.

Yeah, I agree.

merge/real_test.go
lavalamp: This condition should be tested, is it tested elsewhere?

We'd expect the managed fields to no longer reference v1 (assuming all fields are in v2), right? That seems useful to check somehow.

apelisse: Yes, but we don't have to do that with these objects because it's much harder here.

internal/fixture/state.go
lavalamp: I think an explicit hard coded default is less magical than going with the prior thing.

If you want to test multi-version stuff, you should specify this with every change for consistency. If you don't want to test multi version stuff, you don't really care what this defaults to, no?

apelisse: Yeah, I really dislike the default, it really looks much more magic. I'm also fine with just specifying the version everywhere to be honest, I don't think it hurts that much.

lavalamp:
I think the approach needs to be documented somewhere obvious; get an LGTM from anyone once you add that. I'm fine with whatever you decide about the default.

apelisse:
Fixed documentation. PTAL

merge/real_test.go
jennybuckley: Can you just replace `tc.Bench(parser)` below, with something like `tc.Bench(SameVersionParser{T: parser.Type(bench.typename)})` (I think you would also have to wire typename through tc since this is in a different loop)

apelisse: I think that can work yeah ... thanks!

jennybuckley:
/lgtm
We can add back the UpdateVersion benchmark in a future PR

PR #147, opened 2020-01-15
WIP: Add example of testing compatibility change

apelisse:
This creates an example of a compatibility test. It's not great, but we get it mostly for free.

In this case it would show that going from atomic to associative (and vice-versa) is bad because we wouldn't be able to "resolve" the managed fields from one to the other. Even though, maybe we can fix the code to do the right thing.

I'm still not completely convinced that this test does the right thing so I have to think a little bit about it.

cc @mariantalla @lavalamp @jennybuckley

### DIFF
--- a/internal/fixture/state.go
+++ b/internal/fixture/state.go
@@ -26,11 +26,32 @@ import (
 	"sigs.k8s.io/structured-merge-diff/v3/value"
 )
 
+// For the sake of tests, a parser is something that can retrieve a
+// ParseableType.
+type Parser interface {
+	Type(string) typed.ParseableType
+}
+
+// SameVersionParser can be used if all the versions are actually using the same type.
+type SameVersionParser struct {
+	T typed.ParseableType
+}
+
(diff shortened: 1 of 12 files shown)


apelisse:
I suspect we could split this PR in two, the part that adds the feature, and the part that has the test. The test is no good but the feature should be doing the right thing.

jennybuckley:
in `broken_atomic_doesnt_own_former_associative` I would expect that "one" would not manage anything, and that "other" would only manage ".list"
in `broken_associative_doesnt_own_former_atomic` I think I would expect that "one" would not manage anything, and "other" would manage {".list[b]", ".list[b].name", ".list[b].value"}
So these tests do seem to do something unexpected.

jennybuckley:
I think the issue is that it's using "associativeListParser" instead of "atomicListParser" actually

apelisse:
Yeah, the test is clearly broken in many ways. Thanks for looking at it Jenny. We clearly need more tests like this though.

PR #151, opened 2020-01-21
Optimize sync.Pool usage and value.Equals implementations

jpbetz:
- Avoid using `sync.Pool` for iteration loops where we can simply reuse an object directly.
- Introduce `List.Range` to make manage object reuse when looping over lists
- Use `List.Range` to optimize `ListEqual` and `ListCompare`

Benchmarks:
- `sync.Pool` change improves `BenchmarkConvertObjectToTyped` by ~20% for CPU

/sig api-machinery
/cc @apelisse @jennybuckley

### DIFF
--- a/value/listreflect.go
+++ b/value/listreflect.go
@@ -16,7 +16,10 @@ limitations under the License.
 
 package value
 
-import "reflect"
+import (
+	"reflect"
+	"sync"
+)
 
 type listReflect struct {
 	Value reflect.Value
@@ -41,32 +44,44 @@ func (r listReflect) Unstructured() interface{} {
 	return result
 }
 
+var lrrPool = sync.Pool{
+	New: func() interface{} {
+		return &listReflectRange{vr: &valueReflect{}}
+	},
+}
+
 func (r listReflect) Range() ListRange {
-	return &listReflectRange{r.Value, newTempValuePooler(), -1, r.Value.Len()}
+	length := r.Value.Len()
+	if length == 0 {
+		return EmptyRange
+	}
+	rr := lrrPool.Get().(*listReflectRange)
+	rr.list = r.Value
+	rr.i = -1
+	return rr
 }
 
 type listReflectRange struct {
-	val    reflect.Value
-	pooler *tempValuePooler
-	i      int
-	length int
+	list reflect.Value
+	vr   *valueReflect
+	i    int
 }
 
 func (r *listReflectRange) Next() bool {
 	r.i += 1
-	return r.i < r.length
+	return r.i < r.list.Len()
 }
 
 func (r *listReflectRange) Item() (index int, value Value) {
 	if r.i < 0 {
 		panic("Item() called before first calling Next()")
 	}
-	if r.i >= r.length {
+	if r.i >= r.list.Len() {
 		panic("Item() called on ListRange with no more items")
 	}
-	return r.i, r.pooler.NewValueReflect(r.val.Index(r.i))
+	return r.i, r.vr.reuse(r.list.Index(r.i))
 }
 
 func (r *listReflectRange) Recycle() {
-	r.pooler.Recycle()
+	lrrPool.Put(r)
 }
--- a/value/listreflect.go
+++ b/value/listreflect.go
@@ -40,3 +40,33 @@ func (r listReflect) Unstructured() interface{} {
 	}
 	return result
 }
+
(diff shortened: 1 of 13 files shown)


value/listreflect.go
```diff
@@ -40,3 +40,33 @@ func (r listReflect) Unstructured() interface{} {
 	}
 	return result
 }
+
+func (r listReflect) Range() ListRange {
+	return &listReflectRange{r.Value, newTempValuePooler(), -1, r.Value.Len()}
```
apelisse: Note that `listReflectRange` will escape here too

value/listreflect.go
```diff
@@ -40,3 +40,33 @@ func (r listReflect) Unstructured() interface{} {
...
+}
+
+type listReflectRange struct {
+	val    reflect.Value
+	pooler *tempValuePooler
+	i      int
+	length int
```
apelisse: Very important optimization: you can make this a few bytes smaller by removing `length` since it can be computed from value?

value/valuereflect_test.go
```diff
@@ -560,6 +561,18 @@ func TestReflectList(t *testing.T) {
 			if !reflect.DeepEqual(iterateResult, tc.expectedIterate) {
 				t.Errorf("expected iterate to produce %#v but got %#v", tc.expectedIterate, iterateResult)
 			}
+
+			iter := m.Range()
+			iter.Recycle()
```
lavalamp: is this supposed to be a defer?

value/listreflect.go
jpbetz: Sure, I've pooled both together. It works out nicely, actually.

value/listreflect.go
jpbetz: Done. I checked and computing length appears to be a cheap O(1) for all paths, so this should be fine.

value/valuereflect_test.go
jpbetz: Yikes, yes. I'll review all the other Recycle calls too.

value/listreflect.go
```diff
@@ -40,3 +40,33 @@ func (r listReflect) Unstructured() interface{} {
...
 	return result
 }
+
+func (r listReflect) Range() ListRange {
+	return &listReflectRange{r.Value, newTempValuePooler(), -1, r.Value.Len()}
+}
+
+type listReflectRange struct {
+	val    reflect.Value
+	pooler *tempValuePooler
```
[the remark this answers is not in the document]
jpbetz: I was using `valueUnstructured` in the value reflect code to hold values converted from `json.Marshal`, but that doesn't really make things faster than wrapping the result of `json.Marshal` in `valueReflect`, so I've removed `tempValuePooler` and simplified this.

value/listunstructured.go
```diff
@@ -25,3 +25,41 @@ func (l listUnstructured) Length() int {
...
+	}
+	vv := viPool.Get().(*valueUnstructured)
+	return &listUnstructuredRange{l, vv, -1, len(l)}
+}
+
+type listUnstructuredRange struct {
+	list   listUnstructured
+	vv     *valueUnstructured
+	i      int
+	length int
```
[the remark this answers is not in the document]
jpbetz: Removed length here. I'll verify performance is unchanged.

value/mapunstructured.go
```diff
@@ -72,14 +80,12 @@ func (m mapUnstructuredInterface) Equals(other Map) bool {
 		if !ok {
 			return false
 		}
-		vv := NewValueInterface(v)
+		vv.Value = v
```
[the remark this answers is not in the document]
jpbetz: oops, forgot to optimize this one like I had for the other map implementations. Done.

We might want to do the `GetInto/AtInto` optimization as a future improvement It would help in a few ways, like eliminating the need for having custom Equals functions on each map implementation. I'd like to pursue it in a future PR since it's not entirely trivial Callers that are using the map interface don't know if the Into functions are using a underlying `valueReflect` or a `valueUnstructured` so the signature would need to be something like: `GetInto(key string, ValueReuser)` or something where `ValueReuser` (we probably wouldn't call it that) would be able to reuse either a `valueReflect` or a `valueUnstructured` as needed.

jpbetz:
@apelisse, thanks for the review feedback, it has been applied. PTAL

apelisse:
/lgtm
/approve

Thanks!

PR #152, opened 2020-01-23
Optimize conversion to unstructured

jpbetz:
Extend the reflection cache introduced in https://github.com/kubernetes-sigs/structured-merge-diff/pull/136 to also track if a type implements `json.Marshal` and `json.Unmarshal`.

Changes:
- Avoid expensive `reflect.Implements` check on every object being converted by caching implements checks.
- Introduce a `UnstructuredConverter` interface that types like `Time` and `Quantity` can implement to directly convert to unstructured without going through JSON.

Net improvement is roughly 25% for CPU and 20% for allocations for `ObjectToTyped` and about 10% for CPU and allocations for `ToUnstructured`/ `FromUnstructured`.

This has been implemented such that `runtime.ToUnstructured` in `kubernetes/kubernetes` can be trivially modified to use this improvement simply by doing:

```
entry := value.TypeReflectEntryOf(sv.Type())
if entry.CanConvertToUnstructured() {
	v, err := entry.ToUnstructured(sv)
	if err != nil {
		return err
	}
	if v != nil {
		dv.Set(reflect.ValueOf(v))
	}
	return nil
}
```

We will adopt this change up in `kubernetes/kubernetes` in a follow up PR. We will use it instead of [the existing reflection cache in kubernetes for unstructured conversion](https://github.com/kubernetes/kubernetes/blob/dc090f80d14a04eecfe3e46bb5e9e17ed13b5356/staging/src/k8s.io/apimachinery/pkg/runtime/converter.go#L57) and get the performance boost for all `ToUnstructured`/ `FromUnstructured` conversions performed by kubernetes.

Example of using `UnstructuredConverter` to optimize a type:

```
func (t Time) ToUnstructured() interface{} {
	if t.IsZero() {
		return nil
	}
	buf := make([]byte, 0, len(time.RFC3339))
	buf = t.UTC().AppendFormat(buf, time.RFC3339)
	return string(buf)
}
```

We plan to initially adopt `UnstructuredConverter` for `Time` and `Quantity` since they are the most heavily used. We can incrementally optimize or other `json.Marshal` types like `IntToString` as needed.

Baseline:

```
BenchmarkConvertObjectToTyped/Pod/structured-12                    61731             97134 ns/op           11206 B/op        288 allocs/op
BenchmarkConvertObjectToTyped/Pod/unstructured-12                  43767            134630 ns/op           25996 B/op        473 allocs/op
BenchmarkConvertObjectToTyped/Node/structured-12                   29714            201174 ns/op           31905 B/op        737 allocs/op
BenchmarkConvertObjectToTyped/Node/unstructured-12                 23811            252946 ns/op           53644 B/op       1155 allocs/op
BenchmarkConvertObjectToTyped/Endpoints/structured-12               1952           3205514 ns/op          306575 B/op       7041 allocs/op
BenchmarkConvertObjectToTyped/Endpoints/unstructured-12             1524           4241015 ns/op          981595 B/op      16072 allocs/op
```

With optimizations enabled:

```
BenchmarkConvertObjectToTyped/Pod/structured-12                    83481             72445 ns/op            8677 B/op        232 allocs/op
 BenchmarkConvertObjectToTyped/Pod/unstructured-12                  49674            123063 ns/op           24211 B/op        443 allocs/op
 BenchmarkConvertObjectToTyped/Node/structured-12                   43626            136374 ns/op           21611 B/op        538 allocs/op
 BenchmarkConvertObjectToTyped/Node/unstructured-12                 27535            213865 ns/op           45515 B/op       1037 allocs/op
 BenchmarkConvertObjectToTyped/Endpoints/structured-12               2238           2595218 ns/op          306162 B/op       7035 allocs/op
 BenchmarkConvertObjectToTyped/Endpoints/unstructured-12             1515           3989870 ns/op          981368 B/op      16069 allocs/op
```

/sig api-machinery

### DIFF
--- a/value/fields.go
+++ b/value/fields.go
@@ -27,7 +27,7 @@ type Field struct {
 	Value Value
 }
 
-// FieldList is a list of key-value pairs. Each field is expectUpdated to
+// FieldList is a list of key-value pairs. Each field is expected to
 // have a different name.
 type FieldList []Field
 
--- a/value/listreflect.go
+++ b/value/listreflect.go
@@ -58,13 +58,15 @@ func (r listReflect) Range() ListRange {
 	rr := lrrPool.Get().(*listReflectRange)
 	rr.list = r.Value
 	rr.i = -1
+	rr.entry = nil
 	return rr
 }
 
 type listReflectRange struct {
-	list reflect.Value
-	vr   *valueReflect
-	i    int
+	list  reflect.Value
+	vr    *valueReflect
+	i     int
+	entry *TypeReflectCacheEntry
 }
 
 func (r *listReflectRange) Next() bool {
@@ -79,7 +81,11 @@ func (r *listReflectRange) Item() (index int, value Value) {
 	if r.i >= r.list.Len() {
 		panic("Item() called on ListRange with no more items")
 	}
-	return r.i, r.vr.mustReuse(r.list.Index(r.i), nil, nil)
+	v := r.list.Index(r.i)
+	if r.entry == nil {
+		r.entry = TypeReflectEntryOf(v.Type())
+	}
+	return r.i, r.vr.mustReuse(v, r.entry, nil, nil)
 }
 
 func (r *listReflectRange) Recycle() {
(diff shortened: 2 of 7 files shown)


apelisse:
/lgtm
/approve

Thanks!

PR #154, opened 2020-02-04
Remove non-leaf fields from updater's fieldsets

jennybuckley:
The presence of non-leaf fields in managed fieldsets doesn't actually do anything, it can never result in a conflict, and it can never change the result of an operation. I think we should remove them, since they can ultimately lead to some managers getting stuck in the managedFields list that own nothing but non-leaf fields, and since they never conflict with anything they end up sticking around forever.

This bug does not usually present itself in k8s built-in types, because fields with struct values are not usually nullable, and so they get defaulted to a struct with 0-values for all it's fields, instead of nil, meaning they never appear as 'added' in a compare result. Unfortunately CRs are backed by unstructured, where all 'structs' are actually maps, so we should probably fix this to try to be consistent with builtin types.


jennybuckley:
rebased

typed/typed.go
```diff
@@ -238,6 +238,8 @@ func merge(lhs, rhs *TypedValue, rule, postRule mergeRule) (*TypedValue, error)
 	mw.typeRef = lhs.typeRef
 	mw.rule = rule
 	mw.postItemHook = postRule
+	mw.postHook = postRule
+	mw.postItemHook = postItemRule
```
apelisse: You're overriding `postItemHook`

apelisse: I suspect we could do a better job finding a good name for these variables?

typed/typed.go
```diff
@@ -99,7 +99,7 @@ func (tv TypedValue) ToFieldSet() (*fieldpath.Set, error) {
 // match), or an error will be returned. Validation errors will be returned if
 // the objects don't conform to the schema.
 func (tv TypedValue) Merge(pso *TypedValue) (*TypedValue, error) {
-	return merge(&tv, pso, ruleKeepRHS, nil)
+	return merge(&tv, pso, ruleKeepRHS, nil, nil)
```
apelisse: The new parameter is nil everywhere?

apelisse:
It's very questionnable that we still need that after Joe's work on atomic <-> associative. I'll close for now, but we can revisit.

PR #156, opened 2020-02-07
Convert numbers like Kubernetes does for json.Marshaler types

jpbetz:
`k8s.io/apimachinery/pkg/runtime` tests failures from https://github.com/kubernetes/kubernetes/pull/87908 showed that `json.Marshaler` was not handled exactly right. We need to handle numbers in collections exactly as Kubernetes does.

### DIFF
--- a/value/reflectcache.go
+++ b/value/reflectcache.go
@@ -210,7 +210,7 @@ func (e TypeReflectCacheEntry) ToUnstructured(sv reflect.Value) (interface{}, er
 
 		case data[0] == '"':
 			var result string
-			err := json.Unmarshal(data, &result)
+			err := unmarshal(data, &result)
 			if err != nil {
 				return nil, fmt.Errorf("error decoding string from json: %v", err)
 			}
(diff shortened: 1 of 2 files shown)


jpbetz:
cc @apelisse

value/reflectcache.go
```diff
@@ -363,3 +363,101 @@ func (c *typeReflectCache) update(updates reflectCacheMap) {
...
+
+// Below json Unmarshal is fromk8s.io/apimachinery/pkg/util/json
+// to handle number conversions as expected by Kubernetes
+
+// limit recursive depth to prevent stack overflow errors
+const maxDepth = 10000
+
+// Unmarshal unmarshals the given data
+// If v is a *map[string]interface{}, numbers are converted to int64 or float64
+func Unmarshal(data []byte, v interface{}) error {
```
liggitt: keep private for now? I'd prefer not to pick up new callers here if we don't need to

liggitt: can you also bring over the unit tests around this?

jpbetz: Done and done!

apelisse:
Would you mind squashing? Thanks

PR #157, opened 2020-02-13
Fix build for go < 1.12

saschagrunert:
The `MapRange()` function has been added in go 1.12. To maintain
backwards compatiblity with older versions we switch back to the usage
of `MapKeys()` and `MapIndex()`.

May I ask you to cut a new release after we could agree on this fix? :innocent:


apelisse:
Any change here is going to make me very nervous.

WDYT @jpbetz

jpbetz:
This needs to be bechmarked before we can discuss it more meaningfully. `BenchmarkConvertObjectToTyped` from kubernetes/kubernetes like done in https://github.com/kubernetes-sigs/structured-merge-diff/pull/152 would be a good start.

apelisse:
My understanding, from looking at the mentioned issue, is that it's no longer needed.

saschagrunert:
Yeah let’s close this one.

PR #143, opened 2020-01-08
make schema.Equals pointer receivers for Map and Schema

kwiesmueller:
This fixes data races caused by copied mutexes

/assign @apelisse

PR #159, opened 2020-03-16
Nil on noop apply

apelisse:
This is #158 for release-2.0


apelisse:
We probably need to pull #143 if we want to pass the tests ... Doing that now.

apelisse:
Done, this now included #143, and that did fix the tests.

PR #161, opened 2020-05-02
Add ignored fields to update

kwiesmueller:
The fieldset provided to the update/apply will be excluded from any managedFields changes, but still be reflected in the actual update.

- [ ] Needs more documentation or a better name to clarify this behavior.

/sig api-machinery
/wg api-expression
/cc @apelisse


apelisse:
I'm a little confused why there are all these commits in there, do you need #133? You also have a commit that has already been merged.

fieldpath/set.go
```diff
@@ -128,6 +128,31 @@ func (s *Set) Has(p Path) bool {
 	}
 }
 
+// HasPartial returns true if PathElement of p is a member of the set.
+// Example:
+// For a set containing .obj
+// any path starting with .obj (like .obj.a) is part of the set.
+func (s *Set) HasPartial(p Path) bool {
```
apelisse, line 135: I'm not completely sure what this does nor why it's needed yet. But I'm a little curious.

internal/fixture/state.go
```diff
@@ -1,12 +1,9 @@
 /*
 Copyright 2018 The Kubernetes Authors.
-
```
apelisse, line 3: I don't think you should change these.

internal/fixture/state.go
```diff
@@ -184,6 +181,10 @@ func (s *State) CompareLive(obj typed.YAMLObject, version fieldpath.APIVersion)
 	return live.Compare(tv)
 }
 
+func (s *State) parseConfig(apiVersion fieldpath.APIVersion, obj typed.YAMLObject) (*typed.TypedValue, error) {
```
apelisse, line 184: While I don't disagree that this is useful, this is making that change more complicated than it has to be.

internal/fixture/state.go
```diff
@@ -471,12 +475,12 @@ func (tc TestCase) TestWithConverter(parser Parser, converter merge.Converter) e
 
 	// If LastObject was specified, compare it with LiveState
 	if tc.Object != typed.YAMLObject("") {
-		comparison, err := state.CompareLive(tc.Object, tc.APIVersion)
-		if err != nil {
-			return fmt.Errorf("failed to compare live with config: %v", err)
+		expect := ExpectState{
```
apelisse, line 478: too many changes, do we really need this here?

internal/testdata/node.yaml
```diff
@@ -7,14 +7,14 @@ metadata:
     volumes.kubernetes.io/controller-managed-attach-detach: "true"
   creationTimestamp: "2019-07-09T16:17:29Z"
   labels:
-    beta.kubernetes.io/arch: amd64
```
apelisse, line 10: no idea what this is doing here.

merge/update.go
```diff
@@ -261,7 +262,11 @@ func (s *Updater) addBackOwnedItems(merged, pruned *typed.TypedValue, managedFie
 		if err != nil {
 			return nil, fmt.Errorf("failed to create field set from pruned object at version %v: %v", version, err)
 		}
-		pruned = merged.RemoveItems(mergedSet.Difference(prunedSet.Union(managed)))
+
```
apelisse, line 265: all of this is completely unrelated now

typed/typed.go
```diff
@@ -109,12 +109,28 @@ func (tv TypedValue) Merge(pso *TypedValue) (*TypedValue, error) {
...
 	c = &Comparison{
 		Removed:  fieldpath.NewSet(),
 		Modified: fieldpath.NewSet(),
 		Added:    fieldpath.NewSet(),
 	}
+	if ignored == nil {
+		ignored = fieldpath.NewSet()
+	}
 	_, err = merge(&tv, rhs, func(w *mergingWalker) {
+		if ignored.HasPartial(w.path) {
```
apelisse, line 131: I didn't really expect the merge code to be altered by the ignored field. Can't we just remove the fields from the set after we're done?

fieldpath/set.go
kwiesmueller: The current Has behaves like this:

set = [.obj]
set.Has(".obj") -> true
set.Has(".obj.a") -> false

HasPartial behaves like this:

set = [.obj]
set.Has(".obj") -> true
set.Has(".obj.a") -> true

It does what the new Remove does, match children if the parent path is defined.

internal/fixture/state.go
kwiesmueller: Yeah, it's from the rebase. I'll cleanup.

merge/update.go
kwiesmueller: > This PR includes #133 for using both of them in k/k testing.
I can cleanup this.

typed/typed.go
kwiesmueller: We could, but that would mean to go over all managers. I thought this way is more efficient as we could do the other thing in k/k as well.

apelisse: not if you do it at the right time? We can change the list of fields changed before we update the other managers (I think).

fieldpath/set.go
apelisse: OK, If it's not has, there is already a method that does that.

apelisse: `WithPrefix` I think is somewhat similar to what you're trying to do

kwiesmueller: Yeah, i tried around a lot with WithPrefix and didn't get it to work.
It never resulted in a positive match.
Only the Set has Prefix not the Path and prefixing the set didn't help. Also because the merge code does not expose at which step/path it is.
So i tried adding that, but without success.

This is the only way i could get it to work so far :-/

typed/typed.go
kwiesmueller: You mean removing the fields from *typed.Comparison?
That would work, but needs to go over added, modified and removed again and then do the same logic/remove from there.

apelisse: Yeah, I think that's better than altering the merge code though

go.sum
```diff
@@ -17,5 +17,4 @@ gopkg.in/check.v1 v0.0.0-20161208181325-20d25e280405 h1:yhCVgyC4o1eVCa2tZl7eS0r+
 gopkg.in/check.v1 v0.0.0-20161208181325-20d25e280405/go.mod h1:Co6ibVJAznAaIkqp8huTwlJQCZ016jof/cbN4VW5Yz0=
 gopkg.in/yaml.v2 v2.2.1 h1:mUhvW9EsL+naU5Q3cakzfE91YhliOondGd6ZrsDBHQE=
 gopkg.in/yaml.v2 v2.2.1/go.mod h1:hI93XBmqTisBFMUTm0b8Fm+jr3Dg1NNxqwp+5A1VGuI=
-sigs.k8s.io/structured-merge-diff v1.0.1 h1:LOs1LZWMsz1xs77Phr/pkB4LFaavH7IVq/3+WTN9XTA=
```
kwiesmueller, line 20: I don't understand why this always gets done.

merge/ignore_test.go
```diff
@@ -0,0 +1,228 @@
...
+					Manager: "default",
+					Fields: fieldpath.NewSet(
+						fieldpath.MakePathOrDie("numeric"),
+						fieldpath.MakePathOrDie("string"),
+					),
+				},
+				ExpectManagedFields{
+					Manager: "default2",
+					Fields: fieldpath.NewSet(
+						fieldpath.MakePathOrDie("numeric"),
```
kwiesmueller, line 191: Is this actually correct?

merge/update.go
```diff
@@ -128,15 +130,18 @@ func (s *Updater) Update(liveObject, newObject *typed.TypedValue, version fieldp
...
 	}
 	if _, ok := managers[manager]; !ok {
 		managers[manager] = fieldpath.NewVersionedSet(fieldpath.NewSet(), version, false)
 	}
+	if ignored == nil {
+		ignored = fieldpath.NewSet()
+	}
 	managers[manager] = fieldpath.NewVersionedSet(
-		managers[manager].Set().Union(compare.Modified).Union(compare.Added).Difference(compare.Removed),
+		managers[manager].Set().Union(compare.Modified).Union(compare.Added).Difference(compare.Removed).RecursiveDifference(ignored),
```
kwiesmueller, line 144: I'm not entirely sure why this is required yet and if it should be solved somewhere else.

merge/update.go
```diff
@@ -174,12 +179,19 @@ func (s *Updater) Apply(liveObject, configObject *typed.TypedValue, version fiel
 	if err != nil {
 		return nil, fieldpath.ManagedFields{}, fmt.Errorf("failed to get field set: %v", err)
 	}
+	if ignored != nil {
+		set = set.RecursiveDifference(ignored)
+		// TODO: is this correct. If we don't remove from lastSet pruning might remove the fields?
```
kwiesmueller, line 184: After thinking about it again, this might not be required.
Pruning should remove fields from the object if they are not present in newObject and ignored.
Please confirm

typed/typed.go
```diff
@@ -291,3 +291,13 @@ func (c *Comparison) String() string {
 	}
 	return bld.String()
 }
+
+// Remove fields from the compare
```
apelisse, line 295: Maybe make the name and description a little clearer. "Remove" is not clear enough, especially since it has a field name removed, I want to make sure we don't forget what this is.

typed/comparison_test.go
```diff
@@ -0,0 +1,185 @@
+package typed_test
+
+import (
+	"testing"
+
+	"sigs.k8s.io/structured-merge-diff/v3/fieldpath"
+	"sigs.k8s.io/structured-merge-diff/v3/typed"
+)
+
+func TestComparisonRemove(t *testing.T) {
```
apelisse, line 10: I know this sounds weird, but there are too many tests here. "Compare" is merely applying `RecursiveDifference` to three different sets. You should have a test to make sure that you apply it to all three sets and a test to check the empty set and that's it. 

On the other hand, you don't have any tests for `RecursiveDifference`, which is not good :-) And you should add that to the benchmark suite too.

merge/update.go
```diff
@@ -83,6 +83,8 @@ func (s *Updater) update(oldObject, newObject *typed.TypedValue, version fieldpa
 			versions[managerSet.APIVersion()] = compare
 		}
 
+		compare.Remove(ignored)
```
apelisse, line 86: compare is coming out of a cache here (`versions`), can we do the removal before we store in the cache?

merge/ignore_test.go
```diff
@@ -0,0 +1,228 @@
...
+					APIVersion: "v1",
+					Object: `
+						numeric: 1
+						string: "some string"
+					`,
+					IgnoredFields: fieldpath.NewSet(
+						fieldpath.MakePathOrDie("string"),
+					),
+				},
+				ExpectState{
```
apelisse, line 26: See, now we have more code, and you don't even know which feature to use. You technically don't have to use `ExpectState` here, not `ExpectManagedFields`, I don't know if that's needed. You can always split the tests ...

apelisse: Or just remove the other way to check?

apelisse: Also, `ExpectState` is also not the best name, it's actually expecting a specific object?

merge/update.go
```diff
@@ -41,7 +41,7 @@ func (s *Updater) EnableUnionFeature() {
 	s.enableUnions = true
 }
 
-func (s *Updater) update(oldObject, newObject *typed.TypedValue, version fieldpath.APIVersion, managers fieldpath.ManagedFields, workflow string, force bool) (fieldpath.ManagedFields, *typed.Comparison, error) {
+func (s *Updater) update(oldObject, newObject *typed.TypedValue, version fieldpath.APIVersion, managers fieldpath.ManagedFields, ignored *fieldpath.Set, workflow string, force bool) (fieldpath.ManagedFields, *typed.Comparison, error) {
```
apelisse, line 44: I was looking at this, a little sad that we have to add yet another parameter :-/
But then I realized that we don't need to change "Ignored" per request, it's not supposed to change is it?

merge/update.go
apelisse: While thinking that "ignored" isn't per request, I realized that it's "per version". which means that this might be incorrect because ignored doesn't work for all versions.

fieldpath/set.go
```diff
@@ -94,6 +94,53 @@ func (s *Set) Difference(s2 *Set) *Set {
...
 }
 
+// RecursiveDifference returns a Set containing elements which:
+// * appear in s
+// * do not appear in s2
+//
+// Compared to a regular difference, this recursively removes
+// all children from s
+// that are either children or members in s2
+func (s *Set) RecursiveDifference(s2 *Set) *Set {
```
apelisse, line 104: I haven't looked at this code a lot, but it seems pretty long. Other methods of this object delegate a lot to Children and Members, is this possible here?

kwiesmueller: I'll look at it.
I know the difference was done for members and children separately, but in this case both are needed to do the diff.
Feel free to correct me on that.

merge/update.go
kwiesmueller: But we pass different ignores when creating the fieldManager and they get used here, don't they?
So we could put ignored in the updater as subresources and versions all get their own fieldManager

merge/ignore_test.go
kwiesmueller: Okay, I remember having this discussion.
I'll split the tests and remove the operations.

But I'll have to add a way for checking managed fields.

Without a way to check intermediate state, I'll have to split and duplicate where necessary. That will be more code, but probably also easier to debug.

merge/update.go
kwiesmueller: Also I first hat the remove above on line 82, but that didn't work because it got skipped when the version was already present.

fieldpath/set.go
kwiesmueller: I moved the code to SetNodeMap which reduced it a little as well.

typed/comparison_test.go
kwiesmueller: Moved most of the tests to the set_test and added a benchmark.
It's a little worse than the normal Difference, because of the additional call(s) to s2.Members.Has (i assume).

typed/typed.go
```diff
@@ -291,3 +291,14 @@ func (c *Comparison) String() string {
 	}
 	return bld.String()
 }
+
+// ExcludeFields fields from the compare recursively removes the fields
+// from the entire comparison
+func (c *Comparison) ExcludeFields(fields *fieldpath.Set) {
```
apelisse, line 297: Thanks, I think that's a better name!

merge/update.go
apelisse: Right, at that line we do set thing inside the cache, but if you look at it, we're also setting into the cache at the very beginning of the method.

merge/update.go
```diff
@@ -41,14 +42,16 @@ func (s *Updater) EnableUnionFeature() {
...
-func (s *Updater) update(oldObject, newObject *typed.TypedValue, version fieldpath.APIVersion, managers fieldpath.ManagedFields, ignored *fieldpath.Set, workflow string, force bool) (fieldpath.ManagedFields, *typed.Comparison, error) {
+func (s *Updater) update(oldObject, newObject *typed.TypedValue, version fieldpath.APIVersion, managers fieldpath.ManagedFields, workflow string, force bool) (fieldpath.ManagedFields, *typed.Comparison, error) {
 	conflicts := fieldpath.ManagedFields{}
 	removed := fieldpath.ManagedFields{}
 	compare, err := oldObject.Compare(newObject)
 	if err != nil {
 		return nil, nil, fmt.Errorf("failed to compare objects: %v", err)
 	}
 
+	compare.ExcludeFields(s.IgnoredFields[version])
```
apelisse, line 53: mostly a nit, but you could make `ExcludeFields` return self, so that you can just do `version: compare.ExcludeFields(...)` below.

merge/update.go
```diff
@@ -80,11 +83,10 @@ func (s *Updater) update(oldObject, newObject *typed.TypedValue, version fieldpa
 			if err != nil {
 				return nil, nil, fmt.Errorf("failed to compare objects: %v", err)
 			}
+			compare.ExcludeFields(s.IgnoredFields[managerSet.APIVersion()])
```
apelisse, line 86: Right, so that's much better because it's going to ignore the proper fields :-)

internal/fixture/state.go
```diff
@@ -146,7 +147,8 @@ func (s *State) ApplyObject(tv *typed.TypedValue, version fieldpath.APIVersion,
 	if err != nil {
 		return err
 	}
-	new, managers, err := s.Updater.Apply(s.Live, tv, version, s.Managers, ignored, manager, force)
+	defer setIgnoredFieldsAndCleanup(s.Updater, ignored)()
```
apelisse, line 150: Isn't this wrong? You're trying to move something that per-operation to something that should be a state? This should be a parameter of the `Test` method.

kwiesmueller: Yes, but in the current test cases i wanted to be able to change the ignoredFields per operation.
Aside from that i had the same thought.
Mainly the `update_does_not_steal_ignored` test is the reason i change the ignoredFields to emulate behavior like the one of a subresource (where they differ per operation).

apelisse:
This is really starting to look good, a few things to fix and we'll be good to go!

merge/update.go
```diff
@@ -30,7 +30,8 @@ type Converter interface {
 // Updater is the object used to compute updated FieldSets and also
 // merge the object on Apply.
 type Updater struct {
-	Converter Converter
+	Converter     Converter
+	IgnoredFields map[fieldpath.APIVersion]*fieldpath.Set
```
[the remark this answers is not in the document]
kwiesmueller: No, we said that we only have one ignoredFields per endpoint.
I don't remember Daniels exact reasoning but it made sense and we agreed.

merge/update.go
```diff
@@ -135,8 +136,13 @@ func (s *Updater) Update(liveObject, newObject *typed.TypedValue, version fieldp
 	if _, ok := managers[manager]; !ok {
 		managers[manager] = fieldpath.NewVersionedSet(fieldpath.NewSet(), version, false)
 	}
+
+	ignored := s.IgnoredFields[version]
+	if ignored == nil {
```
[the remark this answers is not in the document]
kwiesmueller: Thought so too, could move the check from Comparison there or just check again.
I did it in Comparison to only check once, but on the other hand it's just a nil check.

go.sum
apelisse: I think that's good! Thanks :-)

go.mod
```diff
@@ -7,6 +7,7 @@ require (
 	github.com/json-iterator/go v1.1.6
 	github.com/modern-go/concurrent v0.0.0-20180306012644-bacd9c7ef1dd // indirect
 	github.com/modern-go/reflect2 v1.0.1 // indirect
+	github.com/stretchr/testify v1.3.0 // indirect
```
apelisse, line 10: I don't think that's ok though? What uses it?

kwiesmueller: I don't know, I looked a while back, it was some dependency I think.
Every time I run go mod tidy that gets added.

fieldpath/set.go
```diff
@@ -333,6 +347,49 @@ func (s *SetNodeMap) Difference(s2 *Set) *SetNodeMap {
...
+// all children from s
+// that are either children or members in s2
+func (s *SetNodeMap) RecursiveDifference(s2 *Set) *SetNodeMap {
+	out := &SetNodeMap{}
+
+	i, j := 0, 0
+	for i < len(s.members) && j < len(s2.Children.members) {
+		if s.members[i].pathElement.Less(s2.Children.members[j].pathElement) {
+			if !s2.Members.Has(s.members[i].pathElement) {
+				out.members = append(out.members, setNode{
```
apelisse, line 362: the method above `Difference` is creating the whole `setNode` on one line, you should probably use the same style.

apelisse:
Daniel, do you want to have a look at that "RecursiveDifference" code?
/assign @lavalamp

fieldpath/set_test.go
```diff
@@ -447,6 +453,71 @@ func TestSetIntersectionDifference(t *testing.T) {
 	})
 }
 
+func TestSetRecursiveDifference(t *testing.T) {
+	table := []struct {
+		name   string
+		a      *Set
+		b      *Set
+		expect *Set
```
lavalamp, line 461: Can you add in the expectation for the regular difference? I'm actually not following how this is different?

fieldpath/set_test.go
```diff
@@ -447,6 +453,71 @@ func TestSetIntersectionDifference(t *testing.T) {
...
+		},
+		{
+			name: "remove all matching children",
+			a: NewSet(
+				MakePathOrDie("a", "aa", "aab"),
+				MakePathOrDie("a", "ab", "aba"),
+			),
+			b:      NewSet(MakePathOrDie("a")),
+			expect: NewSet(),
+		},
```
lavalamp, line 509: Missing case of "child" minus "parent": e.g. `"a"` - `"a", "aa"`

fieldpath/set.go
```diff
@@ -94,6 +94,20 @@ func (s *Set) Difference(s2 *Set) *Set {
 	}
 }
 
+// RecursiveDifference returns a Set containing elements which:
+// * appear in s
+// * do not appear in s2
+//
+// Compared to a regular difference, this recursively removes
+// all children from s
+// that are either children or members in s2
```
lavalamp, line 103: Can we align this comment more with the one for `Difference`? I'm having a hard time understanding the difference (no pun intended). Feel free to adjust both comments, the existing one isn't amazingly clear.

go.mod
kwiesmueller: `go mod why github.com/stretchr/testify/...`
```
# github.com/stretchr/testify
(main module does not need package github.com/stretchr/testify)

# github.com/stretchr/testify/assert
sigs.k8s.io/structured-merge-diff/v3/fieldpath
github.com/json-iterator/go
github.com/json-iterator/go.test
github.com/stretchr/testify/require
github.com/stretchr/testify/assert

# github.com/stretchr/testify/http
(main module does not need package github.com/stretchr/testify/http)

# github.com/stretchr/testify/mock
(main module does not need package github.com/stretchr/testify/mock)

# github.com/stretchr/testify/require
sigs.k8s.io/structured-merge-diff/v3/fieldpath
github.com/json-iterator/go
github.com/json-iterator/go.test
github.com/stretchr/testify/require

# github.com/stretchr/testify/suite
(main module does not need package github.com/stretchr/testify/suite)
```

liggitt: go.mod includes all dependencies, even test ones

fieldpath/set_test.go
kwiesmueller: I added both differences to the test and added the test i forgot for both.

merge/update.go
apelisse: > lastSet.Set().RecursiveDifference(ignored)

I think that has to be wrong, we're not editing "VersionedSet" in place, and also `RecursiveDifference` returns a new object, so that's a no-op right now.

I would assume that the `lastSet` has been trimmed already (during the operation that set it), but it's possible that the ignored fields have changed since that manager was written, so doing it again seem reasonable.

fieldpath/set.go
kwiesmueller: Can you check again?

The difference is, that a Difference only respects fields that are directly specified.
So to remove `a.b.c` and `a.b.d` from s, both need to be in s2.
But RecursiveDifference behaves more like deleting a folder that contains folders. It's enough to specify `a` in s2 and it's members are deleted recursively, or in the case of RecursiveDifference not added to the result.

I tried a few different explanations, but none was really good. I'm open for suggestions :-/

lavalamp:
Thanks, it's much clearer now.

/approve

(I only looked at the tests & documentation, maybe @apelisse or @jennybuckley can review the rest and lgtm)

apelisse:
I had a look at the code, thanks!
/lgtm

PR #162, opened 2020-05-07
Add Conflicts.ToSet()

julianvmodesto:
Adds a method to merge.Conflicts to convert all conflicts to a fieldpath.Set.

We use this in https://github.com/kubernetes/kubernetes/pull/84134 when comparing merge.Conflicts to another Set (a Set of allowed conflicts).

### DIFF
--- a/merge/conflict.go
+++ b/merge/conflict.go
@@ -95,6 +95,15 @@ func (c Conflicts) Equals(c2 Conflicts) bool {
 	return true
 }
 
+// ToSet aggregates conflicts for all managers into a single Set.
+func (c Conflicts) ToSet() *fieldpath.Set {
+	set := fieldpath.NewSet()
+	for _, conflict := range []Conflict(c) {
+		set.Insert(conflict.Path)
+	}
+	return set
+}
+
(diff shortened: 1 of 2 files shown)


merge/conflict.go
```diff
@@ -95,6 +95,14 @@ func (c Conflicts) Equals(c2 Conflicts) bool {
 	return true
 }
 
+func (c Conflicts) ToSet() *fieldpath.Set {
```
apelisse, line 98: Looks good, maybe write a quick test?

merge/conflict_test.go
```diff
@@ -61,3 +61,34 @@ conflicts with "Bob":
 		t.Errorf("Got %v, wanted %v", got.Error(), wanted)
 	}
 }
+
+func TestToSet(t *testing.T) {
```
apelisse, line 65: Thanks for adding the test, now that you've done that, I realize that it's aggregating the conflicts one has with all users. You need to document the method above to explain that.

julianvmodesto: Ah yes, done

PR #164, opened 2020-05-24
always drop manager on conversion errors

kwiesmueller:
We decided to drop managers that fail conversions in any case.

Relates to: https://github.com/kubernetes/kubernetes/issues/90904

/sig api-machinery
/wg api-expression
/assign @apelisse 
/cc @apelisse @liggitt

### DIFF
--- a/merge/update.go
+++ b/merge/update.go
@@ -62,19 +62,13 @@ func (s *Updater) update(oldObject, newObject *typed.TypedValue, version fieldpa
 			var err error
 			versionedOldObject, err := s.Converter.Convert(oldObject, managerSet.APIVersion())
 			if err != nil {
-				if s.Converter.IsMissingVersionError(err) {
-					delete(managers, manager)
-					continue
-				}
-				return nil, nil, fmt.Errorf("failed to convert old object: %v", err)
+				delete(managers, manager)
+				continue
 			}
 			versionedNewObject, err := s.Converter.Convert(newObject, managerSet.APIVersion())
 			if err != nil {
-				if s.Converter.IsMissingVersionError(err) {
-					delete(managers, manager)
-					continue
-				}
-				return nil, nil, fmt.Errorf("failed to convert new object: %v", err)
+				delete(managers, manager)
+				continue
 			}
 			compare, err = versionedOldObject.Compare(versionedNewObject)
 			if err != nil {


merge/update.go
```diff
@@ -62,19 +62,13 @@ func (s *Updater) update(oldObject, newObject *typed.TypedValue, version fieldpa
 			var err error
 			versionedOldObject, err := s.Converter.Convert(oldObject, managerSet.APIVersion())
 			if err != nil {
-				if s.Converter.IsMissingVersionError(err) {
-					delete(managers, manager)
-					continue
-				}
-				return nil, nil, fmt.Errorf("failed to convert old object: %v", err)
```
kwiesmueller, line 69: Did i remember correctly that we explicitly didn't want to fail but drop?

apelisse: Correct, I think the new code is right.

liggitt:
this looks correct... a covering test would be great if possible

merge/update.go
```diff
@@ -62,19 +62,13 @@ func (s *Updater) update(oldObject, newObject *typed.TypedValue, version fieldpa
 			var err error
 			versionedOldObject, err := s.Converter.Convert(oldObject, managerSet.APIVersion())
```
liggitt, line 63: I also expected to see the check that the API version was non-empty before trying to call conversion at all

apelisse: Should we make this validation somewhere else? That really shouldn't happen.

liggitt: > That really shouldn't happen.

Isn't that exactly what happened in https://github.com/kubernetes/kubernetes/issues/91006#issuecomment-628293331?

apelisse: :-)

It happened because we used to push such an object for "clearing managedFields". Since we're moving away from that model, it shouldn't happen anymore!

liggitt: if it was an approach we documented and people have picked up using, I'd expect us to handle it

apelisse: https://kubernetes.io/docs/reference/using-api/api-concepts/#clearing-managedfields

Even if we want to keep that pattern, we probably shouldn't reach that low in the stack with the empty `apiVersion`.

kwiesmueller:
I thought we decided to check for that when decoding managedFields in the fieldManager.

merge/update.go
kwiesmueller: I thought we decided to check for that when decoding managedFields in the fieldManager.

merge/update.go
kwiesmueller: After the meeting yesterday, do we still want this behavior (at least for apply).

apelisse:
This is being discussed in https://github.com/kubernetes/kubernetes/issues/90904.

apelisse:
I think we should keep the error the way it is and drop invalid type/versions but fail for error in conversions (since they should mostly be transient. We could also reject automatically the entries that have a obviously wrong `apiVersion`. But all of that should probably be done in k/k.

PR #166, opened 2020-06-22
Remove fields that are omitted from applied configurations if there are no other owners

jpbetz:
Implements the first part of the [WG Apply: Unsetting fields design](https://docs.google.com/document/d/11gUQY7Qvs1m53RTWzDzbOcDUZYMNRQCBQbTukpxmpSg/edit#heading=h.a715f9g8o54d):

Omitting a shared field from an applied configuration:
- Applier gives up ownership but the field remains unchanged.

Omit an exclusively owned field from an applied configuration:
- Optional field: field is removed or defaulted.
- Required field: the apply operation will fail with a validation error.

This makes fields consistent with how associative list/map items already behave.

Added a test converter that renames fields (e.g. conversion of v1 to v2 renames "field_v1"  to "field_v2") and used it to ensure that fields are unset when the unset is applied at a different version than the previous apply.

cc @apelisse @jennybuckley @lavalamp

### DIFF
--- a/merge/leaf_test.go
+++ b/merge/leaf_test.go
@@ -378,6 +378,42 @@ func TestUpdateLeaf(t *testing.T) {
 				),
 			},
 		},
+		"update_apply_omits": {
+			Ops: []Operation{
+				Apply{
+					Manager:    "default",
+					APIVersion: "v1",
+					Object: `
+						numeric: 2
+					`,
+				},
+				Update{
+					Manager:    "controller",
+					APIVersion: "v1",
+					Object: `
+						numeric: 1
+					`,
+				},
+				Apply{
+					Manager:    "default",
+					APIVersion: "v1",
+					Object:     ``,
+				},
+			},
+			Object: `
+						numeric: 1
+			`,
+			APIVersion: "v1",
+			Managed: fieldpath.ManagedFields{
+				"controller": fieldpath.NewVersionedSet(
+					_NS(
+						_P("numeric"),
+					),
+					"v1",
+					false,
+				),
+			},
+		},
 		"apply_twice_remove_different_version": {
 			Ops: []Operation{
 				Apply{
--- a/merge/leaf_test.go
+++ b/merge/leaf_test.go
@@ -345,7 +345,7 @@ func TestUpdateLeaf(t *testing.T) {
 				),
 			},
 		},
-		"apply_twice_dangling": {
+		"apply_twice_remove": {
 			Ops: []Operation{
 				Apply{
 					Manager:    "default",
@@ -365,9 +365,7 @@ func TestUpdateLeaf(t *testing.T) {
 				},
 			},
 			Object: `
-				numeric: 1
 				string: "new string"
-				bool: false
 			`,
 			APIVersion: "v1",
 			Managed: fieldpath.ManagedFields{
@@ -380,7 +378,7 @@ func TestUpdateLeaf(t *testing.T) {
 				),
 			},
 		},
-		"apply_twice_dangling_different_version": {
+		"apply_twice_remove_different_version": {
 			Ops: []Operation{
 				Apply{
 					Manager:    "default",
@@ -400,9 +398,7 @@ func TestUpdateLeaf(t *testing.T) {
 				},
 			},
 			Object: `
-				numeric: 1
 				string: "new string"
-				bool: false
 			`,
 			APIVersion: "v1",
 			Managed: fieldpath.ManagedFields{
@@ -462,7 +458,6 @@ func TestUpdateLeaf(t *testing.T) {
 				},
 			},
 			Object: `
-				string: "string"
 			`,
 			APIVersion: "v1",
 			Managed:    fieldpath.ManagedFields{},
--- a/merge/multiple_appliers_test.go
+++ b/merge/multiple_appliers_test.go
@@ -384,7 +384,7 @@ func testMultipleAppliersFieldUnsetting(t *testing.T, v1, v2, v3 fieldpath.APIVe
 				),
 			},
 		},
-		"unset_scalar_shared_owner": {
+		"unset_scalar_shared_with_applier": {
 			Ops: []Operation{
 				Apply{
 					Manager:    "apply-one",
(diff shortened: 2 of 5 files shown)


apelisse:
Could you please specifically pinpoint the limitation of the code in this pull-request? I'm assuming it won't work perfectly across versions?

apelisse:
By the way, while you're looking at this code, we should understand and fix the bug with "complex nested types".

jpbetz:
> Could you please specifically pinpoint the limitation of the code in this pull-request? I'm assuming it won't work perfectly across versions?

This goes through Updater.prune which does the required conversions and field set comparisons to handle multi version. I‘ll add more tests to showcase this.

jpbetz:
> By the way, while you're looking at this code, we should understand and fix the bug with "complex nested types".

Complex types should be handled properly. I’ll add test cases for this too. If there are specific bugs I should be aware of please send them my way.

apelisse:
> This goes through Updater.prune which does the required conversions and field set comparisons to handle multi version. I‘ll add more tests to showcase this.

Awesome yes, I'd be very surprised if this actually works to be honest. Let me think about it for a minute.

apelisse:
@jpbetz Let me know if I can help you write additional tests!

jpbetz:
> > This goes through Updater.prune which does the required conversions and field set comparisons to handle multi version. I‘ll add more tests to showcase this.
> 
> Awesome yes, I'd be very surprised if this actually works to be honest. Let me think about it for a minute.

I spend a while with venn diagrams and couldn't find anything wrong with it. Let me know if you find any problems.

jpbetz:
> @jpbetz Let me know if I can help you write additional tests!

I added field unsetting tests for scalar and complex fields. I also added multi version testing with a converter that does field renames to make sure that prune handles multi version apply correctly.

I need to clean it up still, but I think it demonstrates that this approach works well for many of the non-trivial cases.

apelisse:
> I added field unsetting tests for scalar and complex fields. I also added multi version testing with a converter that does field renames to make sure that prune handles multi version apply correctly.

Awesome, I was looking forward to it, and will definitely look closely at that code!

jpbetz:
@apelisse I had a look at how defaulting w/r/t this PR.

- if I add defaulting to conversion (https://github.com/jpbetz/structured-merge-diff/pull/new/test-defaults-in-converter) and re-ran the `TestMultipleAppliersFieldUnsetting` test added by this PR. It fails as we expected. `prune` is unable to observe the field is being deleted, so the field ends up ownership orphaned, but its value is not unset as it should be.
- If I use go modules to replace the SMD dependency in k8s to depend on this PR, and then try to reproduce the defaulting problem, using fields like `replicas`, I don't observe the issue. The fields are unset as they should be. I've reviewed the code and it looks like while defaulting does happen after apply (https://github.com/kubernetes/kubernetes/blob/ad29e168dc21aae6abe342a9501f40ec1cfe365d/staging/src/k8s.io/apiserver/pkg/endpoints/handlers/fieldmanager/structuredmerge.go#L172), it does not happen during conversion, which is performed using the `UnsafeObjectConverter` (https://github.com/kubernetes/kubernetes/blob/1ff857e1269fc100c8dda64287180cbfb5b2e657/staging/src/k8s.io/apimachinery/pkg/runtime/helper.go#L45). I've looked over the converter and it doesn't seem to do defaulting anywhere.

So it looks like this approach is safe?

apelisse:
As discussed offline, this is probably because we don't run the conversion (and defaulting) if the version is the same in k8s:
https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apiserver/pkg/endpoints/handlers/fieldmanager/internal/versionconverter.go#L78-L80.

jpbetz:
> As discussed offline, this is probably because we don't run the conversion (and defaulting) if the version is the same in k8s:
> https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apiserver/pkg/endpoints/handlers/fieldmanager/internal/versionconverter.go#L78-L80.

I checked and we don't run defaulting across versions either. I've added some [examples](https://docs.google.com/document/d/11gUQY7Qvs1m53RTWzDzbOcDUZYMNRQCBQbTukpxmpSg/edit#bookmark=id.d4np45pyq3zs) to the design that the behavior is what we want.

I also instrumented the default function for the "singular" field for a customresourcedefinitions object and verified directly that it is not called from within apply, even when multiple conversions are called from prune, but that it is called from `structuredMergeManager.Apply`, after apply completes, when `defaulter.Default` is called explicitly.

```
E0629 21:07:50.056798    3186 update.go:164] Start Apply
E0629 21:07:50.056903    3186 update.go:229] prune: Converting merged to apiextensions.k8s.io/v1beta1
E0629 21:07:50.057022    3186 update.go:265] addBackOwnedItems: Converting merged to apiextensions.k8s.io/v1
E0629 21:07:50.060735    3186 update.go:273] addBackOwnedItems: Converting pruned to apiextensions.k8s.io/v1
E0629 21:07:50.060894    3186 update.go:298] addBackDanglingItems: Converting pruned to apiextensions.k8s.io/v1beta1
E0629 21:07:50.061051    3186 update.go:247] prune: Converting pruned to apiextensions.k8s.io/v1
E0629 21:07:50.061167    3186 update.go:210] End Apply
```

```
E0629 21:07:21.335972    3186 structuredmerge.go:176] Start Calling default
E0629 21:07:21.336211    3186 defaults.go:46] setDefaults_CustomResourceDefinitionSpec called
E0629 21:07:21.336663    3186 defaults.go:46] setDefaults_CustomResourceDefinitionSpec called
E0629 21:07:21.336676    3186 structuredmerge.go:178] End calling default
```

merge/leaf_test.go
```diff
@@ -380,7 +378,7 @@ func TestUpdateLeaf(t *testing.T) {
 				),
 			},
 		},
-		"apply_twice_dangling_different_version": {
+		"apply_twice_remove_different_version": {
```
apelisse, line 381:
```suggestion
		"update_apply_omits": {
			Ops: []Operation{
				Apply{
					Manager:    "default",
					APIVersion: "v1",
					Object: `
						numeric: 2
					`,
				},
				Update{
					Manager:    "controller",
					APIVersion: "v1",
					Object: `
						numeric: 1
					`,
				},
				Apply{
					Manager:    "default",
					APIVersion: "v1",
					Object:     ``,
				},
			},
			Object: `
						numeric: 1
			`,
			APIVersion: "v1",
			Managed: fieldpath.ManagedFields{
				"controller": fieldpath.NewVersionedSet(
					_NS(
						_P("numeric"),
					),
					"v1",
					false,
				),
			},
		},
		"apply_twice_remove_different_version": {
```

merge/multiple_appliers_test.go
```diff
@@ -243,6 +244,300 @@ func TestMultipleAppliersSet(t *testing.T) {
...
+			}
+		}
+	}
+}
+
+func testMultipleAppliersFieldUnsetting(t *testing.T, v1, v2, v3 fieldpath.APIVersion) {
+	tests := map[string]TestCase{
+		"unset_scalar_sole_owner": {
+			Ops: []Operation{
+				Apply{
```
apelisse, line 354: Can we add a test that shares ownership with an updater too?

jpbetz: Added the test. It failed so I poked around and found that the merge logic found:

https://github.com/kubernetes-sigs/structured-merge-diff/blob/068223b26491dad6419fc6712b70a57e0a162ef1/merge/update.go#L249

This seems to suggest that the merge logic is designed to drop associative map/list items from updaters rather than merge it in as it does with other appliers. Now that I'm also having fields be merged using the same logic, this is more problematic, because it would not allow a field like "replica" to be ownership transferred from an applier to an updater.

And thinking about it, it also seems like the wrong thing to do for associative map/list items since an applier might with to ownership transfer those to an updater as well... So I've updated the `appliers_remove_from_controller_real_conversion` test to reflect the behavior change.

But I'm not 100% comfortable with this because I don't understand the reasoning behind why the merge logic would be that way in the first place.

jpbetz: Figured this out.

Offline @apelisse showed me one of the cases we want to preserve the behavior of:

- Applier creates a Deployment with a Container in the PodSpec
- Some controller modifies something about the container using update (not apply)
- The Applier removes the container from their applied configuration
- Desired result: the container is deleted

This case works fine with this PR. It does exactly what we want.

The reason the `appliers_remove_from_controller_real_conversion` test fails the object is first created with an updater, which then owns all the maps, from the root down, which doesn't reflect what we want to test for.  If I change the test so that the applier creates the object initially, and creates map a and b, then the test passes.

One of my misunderstandings here is how updater ownership works. They don't own all fields of an object, only fields they have changed.

jpbetz: I've added an `applier_updater_shared_ownership_real_conversion` test and updated the `appliers_remove_from_controller_real_conversion` to show that this PR does what we want. This ensures both cases:

- An applier creates an object, a controller changes something about an associative map item, the applier removes the item from its applied configuration and the item is deleted
- An updater creates an object, an applier changes something about an associative map item, the applier removes the item from its applied configuration and the item is not deleted (without this PR, the item is deleted, which is not what I would expect or want)

jpbetz:
/hold

Existing k8s unit tests and integration are passing for this change (https://github.com/kubernetes/kubernetes/pull/92661). I'm going to remove this hold once I've added a couple of integration tests in k8s that ensure unsetting fields happens as expected for native types. Once those tests are passing I'll remove this hold.

Feel free to tag this when it looks ready.

jpbetz:
Tests added to https://github.com/kubernetes/kubernetes/pull/92661

/hold cancel

jpbetz:
This is ready to be merged once reviewers are happy with it. I'd like to get it going today if possible so we can bump k8s to use it for 1.19.

PR #167, opened 2020-08-05
Udpate go mod package to v4

apelisse:
We need to update the tag then.


jpbetz:
/lgtm

Scanned the changes and did a grep against a checkout to ensure there are no remaining structured-merge-diff/v3 usages still in the source.

PR #169, opened 2020-08-10
Add test for multiple appliers with atomic maps

jpbetz:
I wanted to ensure the behavior was what I expected so I wrote a test.

If there is already test coverage for this somewhere let me know, otherwise I think we should add it.

/sig api-machinery
/cc @apelisse


apelisse:
Thanks!
/lgtm
/approve

PR #170, opened 2020-08-12
Support granular<->atomic schema changes

jpbetz:
xref: [Design proposal](https://docs.google.com/document/d/1DUipHmrg0PTFsnS7QtMzTQWTC6num9GP0doVQmA3b8k/edit#)

Enables developers to safely add/remove these tags to Kubernetes types, by reconciling existing managed field data with the schema change:

- atomic tags: +listType=atomic, +structType=atomic, +mapType=atomic
- granular tags: +listType=map with +listMapKey, +listType=set, +mapType=granular

The corresponding x-kubernetes OpenAPI annotations for CRDs are also supported.

Need to do more careful benchmarking in k8s, but a simple benchmark in SMD suggests a 3.8% bump in CPU utilization and less than a 0.5% bump in allocations. The low allocation number is expected because the traversal is read-only unless there is an actual reconciliation change to apply afterward.

Before:

```
BenchmarkMultipleApplierRecursiveRealConversion 	   11803	   1547320 ns/op	  708431 B/op	    7441 allocs/op
```

After:

```
BenchmarkMultipleApplierRecursiveRealConversion 	   10000	   1606807 ns/op	  711170 B/op	    7471 allocs/op
```

Unblocks https://github.com/kubernetes/kubernetes/issues/92913 and https://github.com/kubernetes/kubernetes/pull/91256. 

xref https://github.com/kubernetes/kubernetes/pull/93901 which enables `structType=atomic` usage in k8s

### DIFF
--- a/merge/schema_change_test.go
+++ b/merge/schema_change_test.go
@@ -1,0 +1,252 @@
+/*
+Copyright 2018 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package merge_test
+
+import (
+	"testing"
+
+	"sigs.k8s.io/structured-merge-diff/v4/fieldpath"
+	. "sigs.k8s.io/structured-merge-diff/v4/internal/fixture"
+	"sigs.k8s.io/structured-merge-diff/v4/merge"
+	"sigs.k8s.io/structured-merge-diff/v4/typed"
+)
+
+var structParser = func() *typed.Parser {
+	oldParser, err := typed.NewParser(`types:
+- name: v1
+  map:
+    fields:
+      - name: struct
+        type:
+          namedType: struct
+- name: struct
+  map:
+    fields:
+    - name: numeric
+      type:
+        scalar: numeric
+    - name: string
+      type:
+        scalar: string`)
+	if err != nil {
+		panic(err)
+	}
+	return oldParser
+}()
+
+var structWithAtomicParser = func() *typed.Parser {
+	newParser, err := typed.NewParser(`types:
+- name: v1
+  map:
+    fields:
+      - name: struct
+        type:
+          namedType: struct
+- name: struct
+  map:
+    fields:
+    - name: numeric
+      type:
+        scalar: numeric
+    - name: string
+      type:
+        scalar: string
+    elementRelationship: atomic`)
+	if err != nil {
+		panic(err)
+	}
+	return newParser
+}()
+
+func TestGranularToAtomicSchemaChanges(t *testing.T) {
+	tests := map[string]TestCase{
+		"to-atomic": {
+			Ops: []Operation{
+				Apply{
+					Manager: "one",
+					Object: `
+						struct:
+						  numeric: 1
+					`,
+					APIVersion: "v1",
+				},
+				ChangeParser{Parser: structWithAtomicParser},
+				Apply{
+					Manager: "two",
+					Object: `
+						struct:
+						  string: "string"
+					`,
+					APIVersion: "v1",
+					Conflicts: merge.Conflicts{
+						merge.Conflict{Manager: "one", Path: _P("struct")},
+					},
+				},
+				ForceApply{
+					Manager: "two",
+					Object: `
+						struct:
+						  string: "string"
+					`,
+					APIVersion: "v1",
+				},
+			},
+			Object: `
+				struct:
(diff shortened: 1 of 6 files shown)


jpbetz:
@apelisse I'll look into if what is required to have an atomic clear nested managed fields from other field managers when applied.

apelisse:
> Also, the second apply does not require a force, but probably should when it clears the ownership of the fields owned by the first applier

Correct. If we look at both cases (going from atomic to non-atomic and vice versa), both should probably generate a conflict:
- changing an atomic list that already had items MUST be a conflict
- adding an item to a list that used to be atomic should maybe generate a conflict. Note that we also need to somehow transfer ownership of existing items to the owner of the former atomic list.

Trying to see what happens on HA clusters that have multiple versions ... I don't think it has additional consequences beside the typical "things are blurry during transition".

merge/schema_change_test.go
```diff
@@ -0,0 +1,121 @@
...
+			Ops: []Operation{
+				Apply{
+					Manager: "one",
+					Object: `
+						struct:
+						  numeric: 1
+					`,
+					APIVersion: "v1",
+				},
+				ChangeParser{Parser: structWithAtomicParser},
```
apelisse, line 87: That's really cool, I'm really happy you did it that way.

apelisse:
Let me know when I should take a look, thanks :-)

jpbetz:
@apelisse This is ready for review.

jpbetz:
@apelisse to close the loop on the case of handling atomic->granular schema changes. I was concerned that it might be possible to mis-interpret a field set that is already granular as atomic if the fieldset only owned the root of a map/list but not any of it's children (i.e.`{ .: {}}`). I added a test case for this and it does what you had hoped-- it drops ownership of the entire map once no children are owned. It looks like lists are handled the same way. This is great because it means if we ever see `{}` we know for sure the field was atomically owned when the fieldset was written.

jpbetz:
@jennybuckley if you have any extra review bandwidth, this one could use some extra eyes

merge/update.go
```diff
@@ -152,13 +155,45 @@ func (s *Updater) Update(liveObject, newObject *typed.TypedValue, version fieldp
...
+	result := fieldpath.ManagedFields{}
+	for manager, versionedSet := range managers {
+		tv, err := s.Converter.Convert(liveObject, versionedSet.APIVersion())
+		if s.Converter.IsMissingVersionError(err) { // okay to skip, obsolete versions will be deleted automatically anyway
+			continue
+		}
+		if err != nil {
+			return nil, err
+		}
+		reconciled, err := typed.ReconcileFieldSetWithSchema(versionedSet.Set(), tv)
```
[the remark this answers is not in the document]
apelisse: Not asking you to change anything, just curious if we considered that and if that's any better.

typed/reconcile_schema.go
```diff
@@ -0,0 +1,297 @@
...
+		return nil, errorf("toFieldSet: %v", err)
+	}
+	if listValueFieldSet, ok := fieldSetAtPath(valueFieldSet, path); ok {
+		valueFieldSet = prefixWithPath(path, listValueFieldSet)
+		result = result.Union(valueFieldSet)
+	}
+	return result, nil
+}
+
+func prefixWithPath(prefix fieldpath.Path, set *fieldpath.Set) *fieldpath.Set {
```
apelisse, line 268: That makes me a little sad, there's probably a better way to do that :-)

merge/schema_change_test.go
```diff
@@ -0,0 +1,252 @@
...
+					APIVersion: "v1",
+				},
+			},
+			Object: `
+				struct:
+				  numeric: 1
+				  string: "b"
+			`,
+			APIVersion: "v1",
+			Managed: fieldpath.ManagedFields{
```
apelisse, line 233: Very nice!

apelisse (commented):
Looks good, a couple of minor questions

merge/update.go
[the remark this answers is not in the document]
jpbetz: The reason we need the live type is that when we go from atomic to granular, we need to add to the fieldset all the fields that exist in the live object under the field that was previously atomic. We don't know exactly what they are just from the schema, because we don't know which fields are absent, or which map keys or list items are present.

typed/reconcile_schema.go
jpbetz: I've switched to constructing the new field path in a more direct way that doesn't require the prefixing.

jpbetz:
Thanks for the review @apelisse! Feedback applied.

typed/reconcile_schema.go
```diff
@@ -283,6 +276,13 @@ func fieldSetAtPath(node *fieldpath.Set, path fieldpath.Path) (*fieldpath.Set, b
 	return node, ok
 }
 
+func descendToPath(node *fieldpath.Set, path fieldpath.Path) *fieldpath.Set {
```
apelisse, line 279: Nit for next time we look at it, this code should probably live in `fieldpath` rather than here.

apelisse:
We also probably need to cherry-pick this in as many versions as we can :-(

jpbetz:
> We also probably need to cherry-pick this in as many versions as we can :-(

Roger that.

Issue #172, opened 2020-09-17
Panic when converting a struct that embeds a pointer to another struct to Unstructured  


christarazi:
The following setup can lead to a panic. The issue seems to be with embedding a pointer to a struct, when using `runtime.DefaultUnstructuredConverter.ToUnstructured`.

```golang
type A struct {
  *B `json:",inline"`
}

type B struct {
  Foo string `json:"foo"`
}
```

The panic observed is:

```
... Panic: reflect: NumField of non-struct type *v2.CiliumNetworkPolicy (PC=0x46A7A8)

/usr/lib/go/src/runtime/panic.go:975
  in gopanic
/usr/lib/go/src/reflect/type.go:974
  in rtype.NumField
/home/chris/code/cilium/cilium/vendor/sigs.k8s.io/structured-merge-diff/v4/value/reflectcache.go:146
  in buildStructCacheEntry
/home/chris/code/cilium/cilium/vendor/sigs.k8s.io/structured-merge-diff/v4/value/reflectcache.go:153
  in buildStructCacheEntry
/home/chris/code/cilium/cilium/vendor/sigs.k8s.io/structured-merge-diff/v4/value/reflectcache.go:119
  in typeReflectEntryOf
/home/chris/code/cilium/cilium/vendor/sigs.k8s.io/structured-merge-diff/v4/value/reflectcache.go:94
  in TypeReflectEntryOf
/home/chris/code/cilium/cilium/vendor/k8s.io/apimachinery/pkg/runtime/converter.go:480
  in toUnstructured
/home/chris/code/cilium/cilium/vendor/k8s.io/apimachinery/pkg/runtime/converter.go:413
  in unstructuredConverter.ToUnstructured
unknown_fields.go:102
  in detectUnknownFields
validator.go:138
  in NPValidator.ValidateCCNP
validator_test.go:189
  in CNPValidationSuite.Test_GoodCCNP
/usr/lib/go/src/reflect/value.go:336
  in Value.Call
/usr/lib/go/src/runtime/asm_amd64.s:1374
  in goexit
```

The following patch fixes the issue:

```diff
diff --git a/vendor/sigs.k8s.io/structured-merge-diff/v4/value/reflectcache.go b/vendor/sigs.k8s.io/structured-merge-diff/v4/value/reflectcache.go
index 49e6dd169..ff16bd52f 100644
--- a/vendor/sigs.k8s.io/structured-merge-diff/v4/value/reflectcache.go
[...]

PR #173, opened 2020-09-17
Fix panic on pointer to embedded struct

christarazi:
This PR converts an embedded struct into its original type if it is
a pointer.

Fixes: https://github.com/kubernetes-sigs/structured-merge-diff/issues/172

### DIFF
--- a/value/reflectcache.go
+++ b/value/reflectcache.go
@@ -74,7 +74,7 @@ func (f *FieldCacheEntry) CanOmit(fieldVal reflect.Value) bool {
 func (f *FieldCacheEntry) GetFrom(structVal reflect.Value) reflect.Value {
 	// field might be nested within 'inline' structs
 	for _, elem := range f.fieldPath {
-		structVal = structVal.FieldByIndex(elem)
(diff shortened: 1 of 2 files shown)


jpbetz:
The fix looks right to me. Would you add a test to ensure this doesn't get broken in the future? It should be as simple as adding a new tests struct and case here: https://github.com/kubernetes-sigs/structured-merge-diff/blob/1cc6019316282e0491aa45d2880b33c802848cf1/value/valuereflect_test.go#L241

timebertt:
@christarazi what's the status on this?
We have also hit this issue (ref https://github.com/gardener/gardener/issues/2977) and would be very interested in having a upstream fix.

christarazi:
@timebertt Apologies, yes this is still on my radar. Just trying to find the time :). Hopefully in the next week or so.

christarazi:
I signed it

christarazi:
I signed it

apelisse:
ping @jpbetz

ialidzhikov:
/kind bug
/priority important-soon

@jpbetz, could you please recheck?

PR #175, opened 2020-09-30
Ignore fieldset paths not in schema when reconciling

jpbetz:
Fixes a bug in https://github.com/kubernetes-sigs/structured-merge-diff/pull/170. Reporting an error is problematic for rollbacks and HA upgrades where a new field is introduced.


apelisse (commented):
Thanks,
/lgtm
/approve

PR #176, opened 2020-10-01
Fix noop granular case for atomic<->granular schema changes

jpbetz:
If the managed fields for a complex type has no children fields (e.g. `emptyrDir: {}`), and the live object has not fields, skip reconciliation with schemas entirely.

Fixes the integration test failure in k8s: https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/93901/pull-kubernetes-integration/1311377692632813568. How I reproduced:

```bash
kubectl apply -f hack/testdata/retainKeys/deployment/deployment-before.yaml
kubectl apply -f hack/testdata/retainKeys/deployment/deployment-after.yaml
kubectl get deployment test-deployment-retainkeys -oyaml | grep "emptyDir"
```

Before this fix, `emptyDir: {}` would match in the fieldset because reconciliation created a replacement fieldset for the emptyDir that has both a Member and an empty Children list (created as a side effect of calling descendToPath), which prevented it from being properly removed when deployment-after.yaml was applied and the emptyDir was removed entirely and replaced with a  hostPath.

### DIFF
--- a/typed/reconcile_schema.go
+++ b/typed/reconcile_schema.go
@@ -258,12 +258,13 @@ func buildGranularFieldSet(path fieldpath.Path, value *TypedValue) (*fieldpath.S
 	if err != nil {
 		return nil, errorf("toFieldSet: %v", err)
 	}
-	result := fieldpath.NewSet(path)
-	resultAtPath := descendToPath(result, path)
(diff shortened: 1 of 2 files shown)


jpbetz:
@apelisse One last fix and then I think k8s tests should all pass.

apelisse:
shrug
/lgtm
/approve

PR #177, opened 2020-10-19
Allow defaulted keys

apelisse:
This is mostly a revival of #83 from @jennybuckley 

I've rebased it on master, which required a few changes. I've also decided to incorporate a couple of changes:
1. Allow validation of missing default-able associative keys rather than have a "unvalidated" parsing
2. Associate defaults to fields rather than strictly to associative lists (more general)

The tests that Jenny wrote are still passing, which I think is a good sign.


jpbetz:
The implementation fell out nicely on this one.

I don't see any problems with this.

/lgtm

PR #178, opened 2020-11-10
implement wiping managedFields

kwiesmueller:
/cc @apelisse @jennybuckley 
/wg api-expression

### DIFF
--- a/merge/wipe_fields.go
+++ b/merge/wipe_fields.go
@@ -1,0 +1,61 @@
+/*
+Copyright 2020 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+package merge
+
+import (
+	"fmt"
+
+	"sigs.k8s.io/structured-merge-diff/v4/fieldpath"
+	"sigs.k8s.io/structured-merge-diff/v4/typed"
+)
+
+// WipeManagedFields clears the managedFields for the current manager from any fields that
+// have been/will be removed from the object during preparation.
+// The fields to clear are determined based on the fields that were not added or modified in the preparedObject
+// compared to the liveObject
+func WipeManagedFields(
+	liveManagedFields, newManagedFields fieldpath.ManagedFields,
+	manager string,
+	liveObject, preparedObject *typed.TypedValue,
+) (fieldpath.ManagedFields, error) {
+	newManaged, exists := newManagedFields[manager]
+	if !exists {
+		return newManagedFields, nil
+	}
+
+	liveManaged, exists := liveManagedFields[manager]
+	if !exists {
+		return newManagedFields, nil
+	}
+
+	comparisonLivePrepared, err := liveObject.Compare(preparedObject)
+	if err != nil {
+		return nil, fmt.Errorf("comparing live and prepared: %w", err)
+	}
+
+	addedManagedFields := newManaged.Set().Difference(liveManaged.Set())
+	// remove fields from addedManagedFields that are gone after the object got prepared
+	addedManagedFields = addedManagedFields.Difference(comparisonLivePrepared.Added)
+	addedManagedFields = addedManagedFields.Difference(comparisonLivePrepared.Modified)
+
+	newManagedFields[manager] = fieldpath.NewVersionedSet(
+		newManaged.Set().Difference(addedManagedFields),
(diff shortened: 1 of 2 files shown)


merge/wipe_fields.go
```diff
@@ -0,0 +1,61 @@
...
+
+import (
+	"fmt"
+
+	"sigs.k8s.io/structured-merge-diff/v4/fieldpath"
+	"sigs.k8s.io/structured-merge-diff/v4/typed"
+)
+
+// WipeManagedFields clears the managedFields for the current manager from any fields that
+// have been/will be removed from the object during preparation.
```
apelisse, line 26: "removed" is inaccurate? They might be removed, they might be reset?

merge/wipe_fields.go
```diff
@@ -0,0 +1,61 @@
...
+// The fields to clear are determined based on the fields that were not added or modified in the preparedObject
+// compared to the liveObject
```
apelisse, line 28: Maybe give some context on what "preparation", "prepared" means. Since this is outside kubernetes, maybe try to explain that this is meant to detect fields that will be reset/changed, maybe give an example?

apelisse: It's a little unclear what the parameters are.

merge/wipe_fields.go
```diff
@@ -0,0 +1,61 @@
...
+	if !exists {
+		liveManaged = fieldpath.NewVersionedSet(fieldpath.NewSet(), newManaged.APIVersion(), newManaged.Applied())
+	}
+
+	comparisonLivePrepared, err := liveObject.Compare(preparedObject)
+	if err != nil {
+		return nil, fmt.Errorf("comparing live and prepared: %w", err)
+	}
+
+	addedManagedFields := newManaged.Set().Difference(liveManaged.Set())
```
apelisse, line 49: Can you try to explain what is the idea behind that technique? :-)

apelisse (commented):
I'm not sure I understand what this is doing, so I can't really tell if this is correct?

merge/wipe_fields.go
kwiesmueller: In here or in a comment?
The idea is, that in this line we find out which entries we're added by the Update compared to the current managedFields on the liveObject.

In the lines below we remove all entries that would be gone after the object ran through storage (PrepareForCreate/Update), so we end up with only the managedFields entry additions that will be persisted in the end. Then we only write those to the final managedFields.

merge/wipe_fields_test.go
```diff
@@ -0,0 +1,187 @@
...
+			newManagedFields: fieldpath.ManagedFields{
+				"test": fieldpath.NewVersionedSet(_NS(_P("a")), "v1", false),
+			},
+			manager:        "test",
+			liveObject:     `{"a": 1, "b": 2}`,
+			newObject:      `{"a": 1}`,
+			preparedObject: `{"a": 1, "b": 2}`,
+			expectedManagedFields: fieldpath.ManagedFields{
+				// TODO: should `b` still be owned despite the applier not intending to own it?
+				"test": fieldpath.NewVersionedSet(_NS(_P("a") /* _P("b") */), "v1", false),
```
kwiesmueller, line 127: should `b` still be owned despite the applier not intending to own it?

merge/wipe_fields.go
apelisse: In the top-level comment. I think we need to make it super clear what this does when someone has to modify/and understand what this does, why and how!

merge/wipe_fields_test.go
apelisse: `b` should still be owned by whoever owned it before, which is going to be a challenge :-(

merge/wipe_fields_test.go
```diff
@@ -0,0 +1,187 @@
...
+	"testing"
+
+	"sigs.k8s.io/structured-merge-diff/v4/fieldpath"
+	. "sigs.k8s.io/structured-merge-diff/v4/internal/fixture"
+	"sigs.k8s.io/structured-merge-diff/v4/merge"
+	"sigs.k8s.io/structured-merge-diff/v4/typed"
+)
+
+func TestWipeManagedFields(t *testing.T) {
+	type testCase struct {
```
apelisse, line 29: Any chance we can use the testing mechanism we already have to test these? That'd help build more sophisticated scenarios, and understand how these situation happen. I can't quite see how to represent the "prepare" in there though.

kwiesmueller: I looked at it, but the way it works right now I couldn't see a good way of adding the prepared stuff without making the tests much more complicated :-/

merge/wipe_fields_test.go
kwiesmueller: That's gonna be a problem.
Might mean it won't work this way and we'll have to do it directly in the update.
Which then means we come in a kind of chicken egg problem, because this mechanism needs the fields after update so we need to run update twice or something...

kwiesmueller (commented):
Replied to a few comments, might have been right that this was too easy.

apelisse:
I'm going to keep this around until we actually know how we'll solve that.

apelisse:
Looks like this is solved now, closing :-)

PR #179, opened 2020-11-13
Default keys float and int

apelisse:
This pull-request shows that even if the default is a float, this will still merge properly if someone applies with an integer for the field.

### DIFF
--- a/merge/default_keys_test.go
+++ b/merge/default_keys_test.go
@@ -256,7 +256,7 @@ var bookParser = func() *typed.Parser {
                         map:
                           fields:
                           - name: page
-                            default: 2
+                            default: 2.0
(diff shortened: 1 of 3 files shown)


merge/default_keys_test.go
```diff
@@ -322,6 +322,61 @@ func TestDefaultKeysNested(t *testing.T) {
 				),
 			},
 		},
+		"apply_default_type_differs": {
```
lavalamp, line 325: Perhaps a comment or make the name clear that it's float/int that you're testing?

lavalamp:
lgtm, can you squash?

apelisse:
OK I think that name is better.

PR #184, opened 2021-02-03
Fix remove structures

### LINKED ISSUE #171: Field deletion doesn't work well in SSA
k8s version is 1.19.1.

I first created the following by using `kubectl apply --server-side -f backendconfig.yaml`
```
apiVersion: cloud.google.com/v1
kind: BackendConfig
metadata:
  name: bar
  namespace: default
spec:
  connectionDraining:
    drainingTimeoutSec: 30
  securityPolicy:
    name: bol-default-policy
```

The following is what has been created and persisted:
```
apiVersion: cloud.google.com/v1
kind: BackendConfig
metadata:
  creationTimestamp: "2020-09-11T18:18:40Z"
  generation: 1
  managedFields:
  - apiVersion: cloud.google.com/v1
    fieldsType: FieldsV1
    fieldsV1:
      f:spec:
        f:connectionDraining:
          f:drainingTimeoutSec: {}
        f:securityPolicy:
          f:name: {}
    manager: kubectl
    operation: Apply
    time: "2020-09-11T18:18:40Z"
  name: bar
  namespace: default
  resourceVersion: "214762"
  selfLink: /apis/cloud.google.com/v1/namespaces/default/backendconfigs/bar
  uid: 8d90b477-994f-4091-b631-95f19ad10730
spec:
  connectionDraining:
    drainingTimeoutSec: 30
  securityPolicy:
    name: bol-default-policy
```

Then I tried to re-apply the following using `kubectl apply --server-side -f backendconfig.yaml`
```
apiVersion: cloud.google.com/v1
kind: BackendConfig
metadata:
  name: bar
  namespace: default
spec:
  # <=========== connectionDraining field is removed.
  securityPolicy:
    name: bol-default-policy
```
I got error:
```
The BackendConfig "bar" is invalid: spec.connectionDraining: Invalid value: "null": spec.connectionDraining in body must be of type object: "null"
```
I also tried to re-apply the following using `kubectl apply --server-side -f backendconfig.yaml`
```
apiVersion: cloud.google.com/v1
kind: BackendConfig
metadata:
  name: bar
  namespace: default
spec:
  connectionDraining: null  # <======= connectionDraining is set to null
  securityPolicy:
    name: bol-default-policy
```
I got the same error:
```
The BackendConfig "bar" is invalid: spec.connectionDraining: Invalid value: "null": spec.connectionDraining in body must be of type object: "null"
```

On the other hand, I tried to update the created object with `kubectl replace -f backendconfig.yaml`
```
apiVersion: cloud.google.com/v1
kind: BackendConfig
metadata:
  name: bar
  namespace: default
spec:
  connectionDraining: null  # <======= connectionDraining is set to null
  securityPolicy:
    name: bol-default-policy
```
I got the same schema validation error as SSA above.

But if I run `kubectl replace` with the following, it can successfully remove the field w/o error.
```
apiVersion: cloud.google.com/v1
kind: BackendConfig
metadata:
  name: bar
  namespace: default
spec:
  # <========== connectionDraining field is removed.
  securityPolicy:
    name: bol-default-policy
```

I have played with SMD v4 libary a bit, my understanding is that:
SMD v4 set the field as `null` in the merged object when it infers the user want to delete the field.
In this case, initially it is
```
spec:
  connectionDraining:
    drainingTimeoutSec: 30
  securityPolicy:
    name: bol-default-policy
```
Then I SSA:
```
spec:
  securityPolicy:
    name: bol-default-policy
```
In the apiserver, SMD v4 merged the object as the following before the object goes to the validation:
```
spec:
  connectionDraining: null
  securityPolicy:
    name: bol-default-policy
```

The OpenAPI schema validation rejects the object since the `connectionDraining` field is an object:
```
              connectionDraining:
                description: ConnectionDrainingConfig contains configuration for connection
                  draining. For now the draining timeout. May manage more settings
                  in the future.
                properties:
                  drainingTimeoutSec:
                    description: Draining timeout in seconds.
                    format: int64
                    type: integer
                type: object
```

apelisse:
Fixes #171 

This fixes a bug and slightly changes an existing behavior.

Struct that have a null value will now be owned, i.e.
```
struct:
```
will now be owned by whoever applies that (currently, these can't be owned, which means we may accidentally drop the empty struct).

Also, removing a struct will now remove all the items that are not owned by someone else in it (including things that are owned by an updater).

### DIFF
--- a/fieldpath/set.go
+++ b/fieldpath/set.go
@@ -19,6 +19,8 @@ package fieldpath
 import (
 	"sort"
 	"strings"
+
+	"sigs.k8s.io/structured-merge-diff/v4/schema"
 )
 
 // Set identifies a set of fields.
@@ -107,6 +109,30 @@ func (s *Set) RecursiveDifference(s2 *Set) *Set {
 	return &Set{
 		Members:  *s.Members.Difference(&s2.Members),
 		Children: *s.Children.RecursiveDifference(s2),
+	}
+}
+
+// EnsureNamedFieldsAreMembers returns a Set that contains all the
+// fields in s, as well as all the named fields that are typically not
+// included. For example, a set made of "a.b.c" will end-up also owning
+// "a" if it's a named fields but not "a.b" if it's a map.
+func (s *Set) EnsureNamedFieldsAreMembers(sc *schema.Schema, tr schema.TypeRef) *Set {
+	members := PathElementSet{
+		members: make(sortedPathElements, 0, s.Members.Size()+len(s.Children.members)),
+	}
+	atom, _ := sc.Resolve(tr)
+	members.members = append(members.members, s.Members.members...)
+	for _, node := range s.Children.members {
+		// Only insert named fields.
+		if node.pathElement.FieldName != nil && atom.Map != nil {
+			if _, has := atom.Map.FindField(*node.pathElement.FieldName); has {
+				members.Insert(node.pathElement)
+			}
+		}
+	}
+	return &Set{
+		Members:  members,
+		Children: *s.Children.EnsureNamedFieldsAreMembers(sc, tr),
 	}
 }
 
@@ -389,6 +415,31 @@ func (s *SetNodeMap) RecursiveDifference(s2 *Set) *SetNodeMap {
 	}
 
 	return out
+}
+
+// EnsureNamedFieldsAreMembers returns a set that contains all the named fields along with the leaves.
+func (s *SetNodeMap) EnsureNamedFieldsAreMembers(sc *schema.Schema, tr schema.TypeRef) *SetNodeMap {
+	out := make(sortedSetNode, 0, s.Size())
+	atom, _ := sc.Resolve(tr)
+	for _, member := range s.members {
+		tr := schema.TypeRef{}
+		if member.pathElement.FieldName != nil && atom.Map != nil {
+			tr = atom.Map.ElementType
+			if sf, ok := atom.Map.FindField(*member.pathElement.FieldName); ok {
+				tr = sf.Type
+			}
+		} else if member.pathElement.Key != nil && atom.List != nil {
+			tr = atom.List.ElementType
+		}
+		out = append(out, setNode{
+			pathElement: member.pathElement,
+			set:         member.set.EnsureNamedFieldsAreMembers(sc, tr),
+		})
+	}
+
+	return &SetNodeMap{
+		members: out,
+	}
 }
 
 // Iterate calls f for each PathElement in the set.
--- a/merge/ignore_test.go
+++ b/merge/ignore_test.go
@@ -190,6 +190,13 @@ func TestIgnoredFieldsUsesVersions(t *testing.T) {
 			`,
 			APIVersion: "v4",
 			Managed: fieldpath.ManagedFields{
+				"apply-one": fieldpath.NewVersionedSet(
+					_NS(
+						_P("mapOfMapsRecursive"),
+					),
+					"v4",
+					false,
+				),
 				"apply-two": fieldpath.NewVersionedSet(
 					_NS(
 						_P("mapOfMapsRecursive", "aa"),
--- a/merge/multiple_appliers_test.go
+++ b/merge/multiple_appliers_test.go
@@ -554,11 +554,10 @@ func testMultipleAppliersFieldUnsetting(t *testing.T, v1, v2, v3 fieldpath.APIVe
 					`,
 				},
 			},
-			Object: typed.YAMLObject(fmt.Sprintf(`
+			Object: typed.YAMLObject(`
 				struct:
 				  name: a
-				  complexField_%s: null
-			`, v3)),
+			`),
 			APIVersion: v3,
 			Managed: fieldpath.ManagedFields{
 				"apply-one": fieldpath.NewVersionedSet(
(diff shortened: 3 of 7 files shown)


jpbetz:
I just did a pass on the [WG API Expression: Unsetting fields](https://docs.google.com/document/d/11gUQY7Qvs1m53RTWzDzbOcDUZYMNRQCBQbTukpxmpSg/edit#heading=h.qm34ds8b1z19) to make sure we're not contradicting anything we decided on there. I think we're OK there.

jpbetz:
> Struct that have a null value will now be owned, i.e.
> 
> ```
> struct:
> ```

By "null" we really just mean "empty struct" here, right? In [WG API Expression: Unsetting fields](https://docs.google.com/document/d/11gUQY7Qvs1m53RTWzDzbOcDUZYMNRQCBQbTukpxmpSg/edit#heading=h.qm34ds8b1z19) I remember that we were deliberately trying to avoid the distinction between `struct: null` and `struct:`.

jpbetz:
> Also, removing a struct will now remove all the items that are not owned by someone else in it (including things that are owned by an updater).

Can we say more about the motivation for this in the PR description? This seems to complicate how the [ownership transfer](https://kubernetes.io/docs/reference/using-api/server-side-apply/#transferring-ownership) mechanism of Server Side Apply behaves-- updater ownership is now handled differently. Or were we already handling it differently and I just wasn't aware?

fieldpath/set.go
```diff
@@ -110,6 +110,23 @@ func (s *Set) RecursiveDifference(s2 *Set) *Set {
 	}
 }
 
+// NestedSet returns a Set that contains all the fields in s, as well as
```
lavalamp, line 113: `NestedSet` is not a good name, I think the name should include something about ensuring all children are also members. `RecursivelyCompletelyCoveredSet` or something like that? `EnsureAllChildrenAreMembers`?

lavalamp: `EnsureNodesAreMembers`

merge/ignore_test.go
```diff
@@ -174,10 +174,8 @@ func TestIgnoredFieldsUsesVersions(t *testing.T) {
 					APIVersion: "v2",
 				},
 				Apply{
-					Manager: "apply-one",
-					Object: `
-						mapOfMapsRecursive:
-					`,
+					Manager:    "apply-one",
+					Object:     ``,
```
lavalamp, line 178: hm I can see why you changed this test case but I think we need to know what the output is if clients change nothing?

typed/tofieldset.go
```diff
@@ -137,7 +137,9 @@ func (v *toFieldSetWalker) visitMapItems(t *schema.Map, m value.Map) (errs Valid
 		v2 := v.prepareDescent(pe, tr)
 		v2.value = val
 		errs = append(errs, v2.toFieldSet()...)
-		if _, ok := t.FindField(key); !ok {
+		if val.IsNull() || (val.IsMap() && val.AsMap().Length() == 0) {
```
lavalamp, line 140: zero length maps but not zero length lists? That's surprising?

merge/nested_test.go
```diff
@@ -500,6 +512,124 @@ func TestUpdateNestedType(t *testing.T) {
...
+					APIVersion: "v1",
+				},
+				Apply{
+					Manager: "default",
+					Object: `
+					`,
+					APIVersion: "v1",
+				},
+			},
+			Object: `
```
lavalamp, line 596: This result is wrong, `default` should no longer have owned any fields after the controller stomped them, so they should not be removed.

apelisse: The controller owns only the things that it changed, we can't know it wants to own name or not.

apelisse: But this is still wrong, because controller should still own value and that should be preserved. Yes, that's what I'm trying to change right now.

merge/nested_test.go
```diff
@@ -500,6 +512,124 @@ func TestUpdateNestedType(t *testing.T) {
...
+					Manager: "default",
+					Object: `
+						struct: {}
+					`,
+					APIVersion: "v1",
+				},
+			},
+			Object: `
+				struct:
+				  value: 1
```
lavalamp, line 629: This is also wrong, the user stopped expressing an opinion about `struct.name`. However, they co-owned that along with the controller. Since the controller still expresses an opinion, the field should remain.

merge/ignore_test.go
lavalamp: I guess my ask is, leave all the inputs the same and change the outputs in these tests. And copy the cases and modify them if you want to test different inputs.

merge/nested_test.go
lavalamp: oh that's a good point, it was an update. Yeah, the result here is that `struct.value` should exist and = 1.

merge/nested_test.go
lavalamp: I take it back, the update shouldn't take ownership of the name field. so this seems right.

apelisse: I'll still add a test where update took over the name too and see what happens.

typed/tofieldset.go
apelisse: Actually, this is specific to struct fields with no values, it's not relevant for lists.

merge/ignore_test.go
apelisse: I've made the change you requested. I understand that as a reviewer you want to see how the behavior changed, but I think the test is better the way it was written before.

merge/multiple_appliers_test.go
```diff
@@ -554,11 +554,10 @@ func testMultipleAppliersFieldUnsetting(t *testing.T, v1, v2, v3 fieldpath.APIVe
 					`,
 				},
 			},
-			Object: typed.YAMLObject(fmt.Sprintf(`
+			Object: typed.YAMLObject(`
 				struct:
 				  name: a
-				  complexField_%s: null
```
lavalamp, line 560: is this case still tested?

apelisse: Joe had to write the test with the "dangling" struct only because of this bug that I'm fixing. This is the bug fix :-)

PR #185, opened 2021-02-17
Expose ExtractItems as the opposite of RemoveItems

kevindelgado:
This exposes a new method on TypedValue called ExtractItems. Whereas
RemoveItems takes a fieldpath.Set adn returns the typed value with the
provided set items removed from it, ExtractItems returns ONLY those
items provided by the fieldpath.Set, effectively extracting them from
the original TypedValue.

### DIFF
--- a/fieldpath/set.go
+++ b/fieldpath/set.go
@@ -206,6 +206,40 @@ func (s *Set) WithPrefix(pe PathElement) *Set {
 	return subset
 }
 
+// Leaves returns a set containing only the leaf paths
+// of a set.
+func (s *Set) Leaves() *Set {
+	leaves := PathElementSet{}
+	im := 0
+	ic := 0
+
+	// any members that are not also children are leaves
+outer:
+	for im < len(s.Members.members) {
+		member := s.Members.members[im]
+
+		for ic < len(s.Children.members) {
+			d := member.Compare(s.Children.members[ic].pathElement)
+			if d == 0 {
+				ic++
+				im++
+				continue outer
+			} else if d < 0 {
+				break
+			} else /* if d > 0 */ {
+				ic++
+			}
+		}
+		leaves.members = append(leaves.members, member)
+		im++
+	}
+
+	return &Set{
+		Members:  leaves,
+		Children: *s.Children.Leaves(),
+	}
+}
+
 // setNode is a pair of PathElement / Set, for the purpose of expressing
 // nested set membership.
 type setNode struct {
@@ -454,4 +488,18 @@ func (s *SetNodeMap) iteratePrefix(prefix Path, f func(Path)) {
 		pe := n.pathElement
 		n.set.iteratePrefix(append(prefix, pe), f)
 	}
+}
+
+// Leaves returns a SetNodeMap containing
+// only setNodes with leaf PathElements.
+func (s *SetNodeMap) Leaves() *SetNodeMap {
+	out := &SetNodeMap{}
+	out.members = make(sortedSetNode, len(s.members))
+	for i, n := range s.members {
+		out.members[i] = setNode{
+			pathElement: n.pathElement,
+			set:         n.set.Leaves(),
+		}
+	}
+	return out
 }
--- a/fieldpath/set_test.go
+++ b/fieldpath/set_test.go
@@ -140,6 +140,12 @@ func BenchmarkFieldSet(b *testing.B) {
 				randOperand().RecursiveDifference(randOperand())
 			}
 		})
+		b.Run(fmt.Sprintf("leaves-%v", here.size), func(b *testing.B) {
+			b.ReportAllocs()
+			for i := 0; i < b.N; i++ {
+				randOperand().Leaves()
+			}
+		})
 	}
 }
 
@@ -454,6 +460,106 @@ func TestSetIntersectionDifference(t *testing.T) {
 			t.Errorf("expected: \n%v\n, got \n%v\n", i, got)
 		}
 	})
+}
+
+func TestSetLeaves(t *testing.T) {
+	table := []struct {
+		name     string
+		input    *Set
+		expected *Set
+	}{
+		{
+			name:     "empty set",
+			input:    NewSet(),
+			expected: NewSet(),
+		}, {
+			name: "all leaves",
+			input: NewSet(
+				_P("path1"),
+				_P("path2"),
+				_P("path3"),
+			),
+			expected: NewSet(
+				_P("path1"),
+				_P("path2"),
+				_P("path3"),
+			),
+		}, {
+			name: "only one leaf",
+			input: NewSet(
+				_P("root"),
+				_P("root", "l1"),
+				_P("root", "l1", "l2"),
+				_P("root", "l1", "l2", "l3"),
+			),
+			expected: NewSet(
+				_P("root", "l1", "l2", "l3"),
+			),
+		}, {
+			name: "multiple values, check for overwrite",
+			input: NewSet(
+				_P("root", KeyByFields("name", "a")),
+				_P("root", KeyByFields("name", "a"), "name"),
+				_P("root", KeyByFields("name", "a"), "value", "b"),
+				_P("root", KeyByFields("name", "a"), "value", "c"),
+			),
+			expected: NewSet(
+				_P("root", KeyByFields("name", "a"), "name"),
+				_P("root", KeyByFields("name", "a"), "value", "b"),
+				_P("root", KeyByFields("name", "a"), "value", "c"),
+			),
+		}, {
+			name: "multiple values and nested",
+			input: NewSet(
+				_P("root", KeyByFields("name", "a")),
+				_P("root", KeyByFields("name", "a"), "name"),
+				_P("root", KeyByFields("name", "a"), "value", "b"),
+				_P("root", KeyByFields("name", "a"), "value", "b", "d"),
+				_P("root", KeyByFields("name", "a"), "value", "c"),
+			),
+			expected: NewSet(
+				_P("root", KeyByFields("name", "a"), "name"),
+				_P("root", KeyByFields("name", "a"), "value", "b", "d"),
+				_P("root", KeyByFields("name", "a"), "value", "c"),
+			),
+		}, {
+			name: "all-in-one",
+			input: NewSet(
+				_P("root"),
+				_P("root", KeyByFields("name", "a")),
+				_P("root", KeyByFields("name", "a"), "name"),
+				_P("root", KeyByFields("name", "a"), "value", "b"),
+				_P("root", KeyByFields("name", "a"), "value", "b", "c"),
+				_P("root", KeyByFields("name", "a"), "value", "d"),
+				_P("root", KeyByFields("name", "a"), "value", "e"),
+				_P("root", "x"),
+				_P("root", "x", "y"),
+				_P("root", "x", "z"),
+				_P("root", KeyByFields("name", "p")),
+				_P("root", KeyByFields("name", "p"), "name"),
+				_P("root", KeyByFields("name", "p"), "value", "q"),
+			),
+			expected: NewSet(
+				_P("root", KeyByFields("name", "a"), "name"),
+				_P("root", KeyByFields("name", "a"), "value", "b", "c"),
+				_P("root", KeyByFields("name", "a"), "value", "d"),
+				_P("root", KeyByFields("name", "a"), "value", "e"),
+				_P("root", "x", "y"),
+				_P("root", "x", "z"),
+				_P("root", KeyByFields("name", "p"), "name"),
+				_P("root", KeyByFields("name", "p"), "value", "q"),
+			),
+		},
+	}
+
+	for _, tt := range table {
+		t.Run(tt.name, func(t *testing.T) {
+			if got := tt.input.Leaves(); !tt.expected.Equals(got) {
+				t.Errorf("expected %v, got %v for input %v", tt.expected, got, tt.input)
+			}
+		})
+	}
+
 }
 
 func TestSetDifference(t *testing.T) {
--- a/internal/fixture/state.go
+++ b/internal/fixture/state.go
@@ -341,6 +341,83 @@ func (f ForceApplyObject) preprocess(parser Parser) (Operation, error) {
 	return f, nil
 }
 
+// ExtractApply is a type of operation. It simulates extracting an object
+// the state based on the manager you have applied with, merging the
+// apply object with that extracted object and reapplying that.
+type ExtractApply struct {
+	Manager    string
+	APIVersion fieldpath.APIVersion
+	Object     typed.YAMLObject
+}
+
+var _ Operation = &ExtractApply{}
+
+func (e ExtractApply) run(state *State) error {
+	p, err := e.preprocess(state.Parser)
+	if err != nil {
+		return err
+	}
+	return p.run(state)
+}
+
+func (e ExtractApply) preprocess(parser Parser) (Operation, error) {
+
+	tv, err := parser.Type(string(e.APIVersion)).FromYAML(FixTabsOrDie(e.Object))
+	if err != nil {
+		return nil, err
+	}
+	return ExtractApplyObject{
+		Manager:    e.Manager,
+		APIVersion: e.APIVersion,
+		Object:     tv,
+	}, nil
+}
+
+type ExtractApplyObject struct {
+	Manager    string
+	APIVersion fieldpath.APIVersion
+	Object     *typed.TypedValue
+}
+
+var _ Operation = &ExtractApplyObject{}
+
+func (e ExtractApplyObject) run(state *State) error {
+	if state.Live == nil {
+		return state.ApplyObject(e.Object, e.APIVersion, e.Manager, true)
+	}
+	// Get object from state and convert it to current APIVersion
+	current, err := state.Updater.Converter.Convert(state.Live, e.APIVersion)
+	if err != nil {
+		return err
+	}
+	// Get set based on the manager you've applied with
+	set := fieldpath.NewSet()
+	mgr := state.Managers[e.Manager]
+	if mgr != nil {
+		// we cannot extract a set that is for a different version
+		if mgr.APIVersion() != e.APIVersion {
+			return fmt.Errorf("existing managed fieldpath set APIVersion (%s) differs from desired (%s), unable to extract", mgr.APIVersion(), e.APIVersion)
+		}
+		// trying to extract the fieldSet directly will return everything
+		// under the first path in the set, so we must filter out all
+		// the non-leaf nodes from the fieldSet
+		set = mgr.Set().Leaves()
+	}
+	// ExtractFields from the state object based on the set
+	extracted := current.ExtractItems(set)
+	// Merge ApplyObject on top of the extracted object
+	obj, err := extracted.Merge(e.Object)
+	if err != nil {
+		return err
+	}
+	// Reapply that to the state
+	return state.ApplyObject(obj, e.APIVersion, e.Manager, true)
+}
+
+func (e ExtractApplyObject) preprocess(parser Parser) (Operation, error) {
+	return e, nil
+}
+
 // Update is a type of operation. It is a controller type of
 // update. Errors are passed along.
 type Update struct {
--- a/typed/remove.go
+++ b/typed/remove.go
@@ -20,19 +20,26 @@ import (
 )
 
 type removingWalker struct {
-	value     value.Value
-	out       interface{}
-	schema    *schema.Schema
-	toRemove  *fieldpath.Set
-	allocator value.Allocator
+	value         value.Value
+	out           interface{}
+	schema        *schema.Schema
+	toRemove      *fieldpath.Set
+	allocator     value.Allocator
+	shouldExtract bool
 }
 
-func removeItemsWithSchema(val value.Value, toRemove *fieldpath.Set, schema *schema.Schema, typeRef schema.TypeRef) value.Value {
+// removeItemsWithSchema will walk the given value and look for items from the toRemove set.
+// Depending on whether shouldExtract is set true or false, it will return a modified version
+// of the input value with either:
+// 1. only the items in the toRemove set (when shouldExtract is true) or
+// 2. the items from the toRemove set removed from the value (when shouldExtract is false).
+func removeItemsWithSchema(val value.Value, toRemove *fieldpath.Set, schema *schema.Schema, typeRef schema.TypeRef, shouldExtract bool) value.Value {
 	w := &removingWalker{
-		value:     val,
-		schema:    schema,
-		toRemove:  toRemove,
-		allocator: value.NewFreelistAllocator(),
+		value:         val,
+		schema:        schema,
+		toRemove:      toRemove,
+		allocator:     value.NewFreelistAllocator(),
+		shouldExtract: shouldExtract,
 	}
 	resolveSchema(schema, typeRef, val, w)
 	return value.NewValueInterface(w.out)
@@ -59,11 +66,22 @@ func (w *removingWalker) doList(t *schema.List) (errs ValidationErrors) {
 		// Ignore error because we have already validated this list
 		pe, _ := listItemToPathElement(w.allocator, w.schema, t, i, item)
 		path, _ := fieldpath.MakePath(pe)
+		// save items on the path when we shouldExtract
+		// but ignore them when we are removing (i.e. !w.shouldExtract)
 		if w.toRemove.Has(path) {
-			continue
+			if w.shouldExtract {
+				newItems = append(newItems, item.Unstructured())
+			} else {
+				continue
+			}
 		}
 		if subset := w.toRemove.WithPrefix(pe); !subset.Empty() {
-			item = removeItemsWithSchema(item, subset, w.schema, t.ElementType)
+			item = removeItemsWithSchema(item, subset, w.schema, t.ElementType, w.shouldExtract)
+		} else {
+			// don't save items not on the path when we shouldExtract.
+			if w.shouldExtract {
+				continue
+			}
 		}
 		newItems = append(newItems, item.Unstructured())
 	}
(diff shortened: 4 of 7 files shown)


kevindelgado:
As we discussed, the changes to modify remove into extract don’t seem to justify needing an entirely separate walker. That being said, I’m not a fan of the naming. Here I’ve left  it as `removeWalker` with a `shouldExtract` flag.  I considered calling it `extractOrRemoveWalter`, but didn’t know if that was better or worse. Some other thoughts I had were, `isolatingWalker`, `separatingWalker`, etc. Definitely open to any suggestions.

kevindelgado:
For testing I mostly pattern-matched off of what seemed relevant from `merge_test` and `toset_test`.  I feel like I might be missing some, so let me know if any edge cases aren’t being tested or alternatively if any seem redundant and should be removed. 

The copypasta of shared testing schemas feels a little wrong. Let me know if we want to organize them better in some sort of shared util file or something (I can do it in a separate PR since it’s not really relevant here).

I attempted to test with a more nested schema such as https://github.com/kubernetes-sigs/structured-merge-diff/blob/master/merge/nested_test.go#L28-L112 but I ran into some difficulties. Maybe I was just constructing the typed values and fieldsets wrong, but it seemed like whatever fieldset I passed it would always remove nothing (leaving the original value) and extract nothing (leaving nil/empty value). Here’s an example of what I’m talking about for the schema above

```
                `{"listOfLists":[{"name": "a", "value":["b", "c"]}, {"name": "d"}]}`, // input typed value
                _NS(_P("listsOfLists", _KBF("name", "d"))),                                      // input fieldset to remove/extract
                `{"listOfLists":[{"name": "a", "value":["b", "c"]}]}`,                          // expected remove output
                `{"listOfLists":[{"name": "d"}]}`,                                                        // expected extract output
```
and got
```
--- FAIL: TestRemove (0.00s)
    --- FAIL: TestRemove/nested_types (0.00s)
        --- FAIL: TestRemove/nested_types/nested_types-valid-1 (0.00s)
            remove_test.go:449: Expected
                listOfLists=[name="a"value=["b","c"]]
                but got
                listOfLists=[name="a"value=["b","c"],name="d"]
            remove_test.go:464: Expected
                listOfLists=[name="d"]
                but got
                null

``` 
I noticed that none of the tests in the typed package use nested schemas (only tests in the merge pkg do). So I’m guessing there’s something I don’t know about nested types that makes unit testing them in the typed package not correct. Do you know?

typed/remove_test.go
```diff
@@ -0,0 +1,280 @@
...
+			t.Parallel()
+			pt := parser.Type(tt.rootTypeName)
+
+			tv, err := pt.FromYAML(quadruplet.object)
+			if err != nil {
+				t.Fatalf("unable to parser/validate object yaml: %v\n%v", err, quadruplet.object)
+			}
+
+			set := quadruplet.set
+			if err != nil {
```
apelisse, line 240: There's no new error here :-) Also not sure you need the `set` variable.

typed/remove.go
```diff
@@ -20,19 +20,26 @@ import (
...
+	shouldExtract bool
 }
 
-func removeItemsWithSchema(val value.Value, toRemove *fieldpath.Set, schema *schema.Schema, typeRef schema.TypeRef) value.Value {
+// removeItemsWithSchema with walk the given value and look for items form the toRemove set.
+// Depending on whether shouldExtract is set true or false, it will return a modified version
+// of the input value with either:
+// 1. only the items in the toRemove set (when shouldExtract is true) or
+// 2. the items from the toRemove set removed from the value (when shouldExtract is false).
+func removeItemsWithSchema(val value.Value, toRemove *fieldpath.Set, schema *schema.Schema, typeRef schema.TypeRef, shouldExtract bool) value.Value {
```
apelisse, line 36: At first I was curious if we wanted to rename this type/method since it's not doing the same thing now, but it's still actually removing items. So i think we're good.

typed/remove_test.go
```diff
@@ -0,0 +1,280 @@
...
+    - name: nv
+      type:
+        scalar: numeric
+- name: myValue
+  map:
+    elementType:
+      scalar: string
+`
+
+var removeCases = []removeTestCase{{
```
apelisse, line 144: These tests look good to me.

apelisse (commented):
One super minor nit, thanks!

typed/remove_test.go
kevindelgado: whoops, old code. fixed

typed/remove.go
kevindelgado: 👍

typed/remove_test.go
kevindelgado: 👍

kevindelgado:
feedback addressed, PTAL @apelisse (also, don't forget to /ok-to-test it because I cannot).

apelisse:
That nested test looks good, we probably need more of these, not sure why it's failing yet.

typed/remove_test.go
```diff
@@ -0,0 +1,275 @@
...
+
+			// test RemoveItems
+			rmOut, err := pt.FromYAML(quadruplet.removeOutput)
+			if err != nil {
+				t.Fatalf("unable to parser/validate removeOutput yaml: %v\n%v", err, quadruplet.removeOutput)
+			}
+
+			rmGot := tv.RemoveItems(quadruplet.set)
+			if !value.Equals(rmGot.AsValue(), rmOut.AsValue()) {
+				t.Errorf("Expected\n%v\nbut got\n%v\n",
```
apelisse, line 247: Can we change the string here to differentiate betwee tne "Removed" and "Extracted"?

kevindelgado: done

kevindelgado:
Fixed the issue with doList and added all the nested tests I could think of. Tried to be systematic about it so that we have full coverage of the same tests across all nested types (listOfLists, listOfMaps, mapOfLists, mapOfMaps). 

There are some discrepancies between list-as-top-level types and maps-as-top-level types which creates slightly different test results, I've highlighted these in the inline test comments, but I think these discrepancies are mainly due to the fact that the schema defines listOf* types as requiring name and value whereas mapOf* types do not (see the nested types schema). We can change the schema if we want the tests to look more symmetric, but the added listOf* constraints make it clear what the limitations of `ExtractItems` are. 

Let me know if I'm missing any tests or if you think the bahavior is not what we want. 
PTAL @apelisse

typed/remove_test.go
```diff
@@ -0,0 +1,646 @@
...
+		`{"mapOfMaps": {"b":{"a":"x","c":"z"}, "d":{"e":"y", "f":"w"}}}`,
+		`{"mapOfMaps": null}`,
+	}, {
+		// path with existant top-level but non-existant leaf element
+		`{"mapOfMaps": {"b":{"a":"x","c":"z"}, "d":{"e":"y", "f":"w"}}}`,
+		_NS(
+			_P("mapOfMaps", "b", "q"),
+		),
+		`{"mapOfMaps": {"b":{"a":"x","c":"z"}, "d":{"e":"y", "f":"w"}}}`,
+		`{"mapOfMaps": {"b":null}}`,
```
apelisse, line 536: Should we even keep `b` here? Or should we even keep anything at all since the one thing we're supposed to keep doesn't exist?

typed/remove_test.go
```diff
@@ -0,0 +1,646 @@
...
+		_NS(
+			_P(_V("a")),
+		),
+		`{"mapOfLists": {"b":["a","c"], "d":["e", "f"]}}`,
+		``,
+	}, {
+		// path to root type
+		`{"mapOfMaps": {"b":{"a":"x","c":"z"}, "d":{"e":"y", "f":"w"}}}`,
+		_NS(
+			_P("mapOfMaps"),
```
apelisse, line 501: we need to think about that too, that's not completely obvious to me

apelisse:
Considering the goal we have for this function, I think that a good way to test/agree on what the behavior should be could go this way:
1. Create an object with some fields
2. Apply only some fields to that object, look at the resulting fieldset for that applier
3. Extract from the object using the fieldset we got in 2., we should get back what was applied.

I suspect we should set-up a test for that. A good way to do/understand 2.  is to look at "ToFieldSet", I think if we make the behaviors match there, we should be good to go.

apelisse:
> 1. Create an object with some fields
> 2. Apply only some fields to that object, look at the resulting fieldset for that applier
> 3. Extract from the object using the fieldset we got in 2., we should get back what was applied.

In order to keep this test as minimal as possible, I would probably use `typed.Merge` to merge, and `typed.ToFieldSet` to get "apply" and "get the fieldset". That way you don't depend on any other package.

Basically:
1. `obj := typed.AsTyped(basedObject)`
2. `obj := obj.Merge(appliedObject)`, and `fieldset := obj.ToFieldSet()`
3. `assert(obj.Extract(fieldset) == appliedObject)`

kevindelgado:
Okay, I think I’ve got the reversibleExtract tests, but there’s one little quirk I need you to sign off on:

`ExtractItems` and `ToFieldSet` are reversible, just not directly. Going form `ToFieldSet` to `ExtractItems` requires filtering out non-leaf nodes.

The issue here is that the fieldset returned by `ToFieldSet` contains all the intermediate nodes in the fieldpath.Set (tree). Thus if you try to extract the set directly, it will return everything under the first path in the set.  

As a result, I think both `ToFieldSet` and `ExtractItems` are behaving correctly; they are just not exactly reversible. They are reversible though if you apply a filter on the output of `ToFieldSet` to remove all the intermediate nodes. This is what I have done here in the tests, and I imagine this is how we want to do it in the clients (probably via some helper we provide).

The alternative is to try to extract only the leafs of the provided fieldpath.Set. I believe this would be incorrect behavior, because you would be unable to extract a top level element and receive all its nested elements (which I believe is the desired behavior).

You might be able to find a way to may `ExtractItems` and `ToFieldSet` reversible if you required `ExtractItems` to only accept fieldpath.Sets that contain only leaf nodes, but this is highly inconsistent with how `RemoveItems` behaves and would be super complicated (maybe impossible) with the walker and resolveSchema paradigm. Furthermore, I think we can confidently say that `ExtractItems` and `ToFieldSet` are deterministically reversible, albeit just not as simple as we hoped it would be by directly piping them into each other.

typed/remove_test.go
```diff
@@ -0,0 +1,791 @@
...
+		`{"listOfMaps": [{"name": "a", "value": {"b":"x", "c":"y"}}, {"name": "d", "value": {"e":"z"}}]}`,
+		`{"mapOfLists": {"b":["y","z"]}}`,
+	}, {
+		// misc: recursive deeply nested leaves
+		`{"mapOfMapsRecursive": {"a":{"b":{"c":null}, "d":{"e":{"f":null}, "g":null}}}}`,
+		`{"mapOfMapsRecursive": {"a":{"d":{"e":{"f":{"q":null}, "p":null}}}}}`,
+	}},
+}}
+
+func filterNonLeaves(set *fieldpath.Set) *fieldpath.Set {
```
apelisse, line 715: This sounds like a reasonable implementation to have directly on the `fieldpath.Set` object

typed/remove_test.go
kevindelgado: This is exactly how Remove works currently. If you pass remove a fieldset with only the root element, it will remove everything and return an empty value, so I would Extract to work the same way in reverse, where passing a the root element will "extract" (i.e. keep) everything and return a copy of the original value.

fieldpath/set.go
```diff
@@ -206,6 +206,32 @@ func (s *Set) WithPrefix(pe PathElement) *Set {
 	return subset
 }
 
+// Leaves returns a set containing only the leaf paths
+// of a set.
+func (s *Set) Leaves() *Set {
```
apelisse, line 211: That's probably not optimal, but I would also suggest against spending time optimizing for something we may not need to optimize.

typed/remove_test.go
kevindelgado: Yea, that's a good question. I guess my thinking for keeping it this way is that the caller can use the fieldpath.Set they passed in to know that the value did not have a path to the element they expected. If we really don't want it to keep b, we could do some post-processing in `ExtractItems` to get rid of it, that way we don't have to modify the walker, because I think doing it in the walker gets tricky.

kevindelgado (commented):
Addressed feedback and added `Leaves()` as a method on the `fieldpath.Set` object (along with testing).

PTAL @apelisse

fieldpath/set.go
```diff
@@ -206,6 +206,32 @@ func (s *Set) WithPrefix(pe PathElement) *Set {
...
+			// then the parent must not be a leaf and should be
+			// overwritten by the current path
+			if !parentPath.Equals(path) {
+				paths[n] = path
+				n++
+			}
+		}
+		paths = paths[:n]
+		// save a copy of current path or else it can be overwritten
+		pathCopy := Path(make([]PathElement, len(p)))
```
apelisse, line 228: Any reason why you're not inserting directly in the set?

fieldpath/set.go
apelisse: On the other hand, the "optimal" solution might be easier to read.

fieldpath/set.go
```diff
@@ -206,6 +206,32 @@ func (s *Set) WithPrefix(pe PathElement) *Set {
...
 }
 
+// Leaves returns a set containing only the leaf paths
+// of a set.
+func (s *Set) Leaves() *Set {
+	paths := []Path{}
+	s.Iterate(func(p Path) {
+		parentPath := Path(p[0 : len(p)-1])
+		n := 0
+		for _, path := range paths {
```
apelisse, line 216: isn't this O(n^2) over the number of paths?

fieldpath/set_test.go
```diff
@@ -140,6 +140,12 @@ func BenchmarkFieldSet(b *testing.B) {
 				randOperand().RecursiveDifference(randOperand())
 			}
 		})
+		b.Run(fmt.Sprintf("leaves-%v", here.size), func(b *testing.B) {
+			b.ReportAllocs()
+			for i := 0; i < b.N; i++ {
+				randOperand().Leaves()
```
apelisse, line 146: I'm glad we have a benchmark!

fieldpath/set_test.go
```diff
@@ -456,6 +462,73 @@ func TestSetIntersectionDifference(t *testing.T) {
...
 }
 
+func TestSetLeaves(t *testing.T) {
+	table := []struct {
+		name     string
+		input    *Set
+		expected *Set
+	}{
+		{
+			name:     "empty set",
```
apelisse, line 472: Tests look good, thanks!

kevindelgado:
Added a simple ExtractApply test suite based on our conversation yesterday.
On second look, I don't think the big, hairy, multiple appliers tests added any value that isn't captured by these tests so I removed them, but I can add them/some of them back if you think there's value. Let me know if you think there are any cases I'm missing. 

Also, I didn't do any APIVersion checking, because it doesn't seem like there's anything unique to ExtractApply that needs to be tested w.r.t. to APIVersion, but I could be misunderstanding what we want out of an APIVersion check. I know you said we should fail if APIVersion is inconsitent, but it didnt' seem like merge_tests is the place to test that. If it's not already being tested somewhere else (like at a higher level), let me know where I should add tests for it.

I'm working on optimizing set.Leaves() now, but wanted to get the tests up for review in the meantime.

PTAL @apelisse

fieldpath/set.go
kevindelgado: Attempted a more performant solution.

kevindelgado (commented):
I've attempted a more performant implementation of `fieldpath.Set#Leaves`. It uses the Set's builtin `Children` and `Members` fields as discussed and inserts directly into the Set. I'm open to any suggestions to making it better.

Compared to the old way it is much faster and uses up a bit less space.

old:
```
BenchmarkFieldSet/leaves-20-16                             76585             15869 ns/op            7256 B/op         95 allocs/op
BenchmarkFieldSet/leaves-50-16                             18986             63825 ns/op           22061 B/op        268 allocs/op
BenchmarkFieldSet/leaves-100-16                             6392            209026 ns/op           75540 B/op        828 allocs/op
BenchmarkFieldSet/leaves-500-16                              565           2200738 ns/op          381671 B/op       3949 allocs/op
BenchmarkFieldSet/leaves-1000-16                             171           7137739 ns/op          822994 B/op       8319 allocs/op
```

new:
```
BenchmarkFieldSet/leaves-20-16                            111402             10611 ns/op            4486 B/op         70 allocs/op
BenchmarkFieldSet/leaves-50-16                             31767             38146 ns/op           15083 B/op        213 allocs/op
BenchmarkFieldSet/leaves-100-16                            10000            123891 ns/op           56641 B/op        720 allocs/op
BenchmarkFieldSet/leaves-500-16                             1759            720563 ns/op          284741 B/op       3434 allocs/op
BenchmarkFieldSet/leaves-1000-16                             777           1579806 ns/op          613902 B/op       7302 allocs/op
```

PTAL @apelisse

internal/fixture/state.go
```diff
@@ -341,6 +341,84 @@ func (f ForceApplyObject) preprocess(parser Parser) (Operation, error) {
...
+		set := fieldpath.NewSet()
+		mgr := state.Managers[e.Manager]
+		if mgr != nil {
+			// trying to extract the fieldSet directly will return everything
+			// under the first path in the set, so we must filter out all
+			// the non-leaf nodes from the fieldSet
+			set = mgr.Set().Leaves()
+		}
+		// ExtractFields from the state object based on the set
+		extracted := state.Live.ExtractItems(set)
```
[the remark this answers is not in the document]
kevindelgado: I think I've done this, now. Take a look and let me know if that's what you're expecting

kevindelgado:
Addressed comments relating to `ExtractApply` operation and further optimized the `Leaves` call.
New benchmark is slightly faster:
```
BenchmarkFieldSet/leaves-20-16                            224038              5368 ns/op            3269 B/op         51 allocs/op
BenchmarkFieldSet/leaves-50-16                             61814             19330 ns/op            9574 B/op        164 allocs/op
BenchmarkFieldSet/leaves-100-16                            15668             75564 ns/op           35945 B/op        690 allocs/op
BenchmarkFieldSet/leaves-500-16                             2814            421019 ns/op          172249 B/op       3432 allocs/op
BenchmarkFieldSet/leaves-1000-16                            1197            999801 ns/op          388709 B/op       7713 allocs/op
```

For posterity, an attempt was made to optimize even further by using two pointers to iterate through the members list once instead of using nested for loops. That had an almost identical benchmark profile (actually, ever so slightly slower) and was much less readable.

That attempt lives here https://github.com/kevindelgado/structured-merge-diff/blob/double-pointer-leaves/fieldpath/set.go#L209-L246

fieldpath/set.go
```diff
@@ -455,3 +478,17 @@ func (s *SetNodeMap) iteratePrefix(prefix Path, f func(Path)) {
...
+// Leaves returns a SetNodeMap containing
+// only setNodes with leaf PathElements.
+func (s *SetNodeMap) Leaves() *SetNodeMap {
+	out := &SetNodeMap{}
+	for _, n := range s.members {
+		next := setNode{
+			pathElement: n.pathElement,
+			set:         n.set.Leaves(),
+		}
+		out.members = append(out.members, next)
```
apelisse, line 491: Since we know how many members we have, I think we can pre-allocate the list beforehand. That probably has more impact that the double loop above :-)

fieldpath/set.go
```diff
@@ -206,6 +206,29 @@ func (s *Set) WithPrefix(pe PathElement) *Set {
 	return subset
 }
 
+// Leaves returns a set containing only the leaf paths
+// of a set.
+func (s *Set) Leaves() *Set {
+	members := PathElementSet{}
+	for _, member := range s.Members.members {
+		isChild := false
+		for _, child := range s.Children.members {
```
apelisse, line 215: @lavalamp Do you have an opinion about the O(n^2) loop here? There's no significant different in benchmarks, but I'm wondering why.

fieldpath/set.go
kevindelgado: I've added preallocation both here and with the `PathElementSet` [here](https://github.com/kubernetes-sigs/structured-merge-diff/pull/185/files#diff-cbe655e4df83a27b44de60b4d4008cbcf550462c360f95d36aeb9826a2af525fR212)

makes it ever so slightly faster
```
BenchmarkFieldSet/leaves-20-16                            317458              3785 ns/op            2250 B/op         41 allocs/op
BenchmarkFieldSet/leaves-50-16                             82166             14462 ns/op            6881 B/op        141 allocs/op
BenchmarkFieldSet/leaves-100-16                            20643             60226 ns/op           29160 B/op        632 allocs/op
BenchmarkFieldSet/leaves-500-16                             2984            363610 ns/op          146550 B/op       3212 allocs/op
BenchmarkFieldSet/leaves-1000-16                            1407            851304 ns/op          325102 B/op       7189 allocs/op
```

fieldpath/set.go
jpbetz: Can you binary search s.Children.members instead of scanning? Looks like they are [ordered](https://github.com/kubernetes-sigs/structured-merge-diff/blob/a4e00e9840c22df6cc47c60dcf65984228558387/fieldpath/set.go#L218
). I doubt it will matter for the typical case, but maybe help for the worst case?

jpbetz: I.e. just do `s.Children.members.Descend(child.pathElement)`

apelisse: I can't quite tell if Descend is the right thing, but search is as easy as something like this
```
loc := sort.Search(len(s.members), func(i int) bool {
		return !s.members[i].pathElement.Less(pe)
	})
```

apelisse: But even that is `O(n * log p)` with `n` the number of members and `p` the number of children. Moving the two pointers should be `O(n + p)`

kevindelgado: This results in an almost identical benchmark to both the nested for loop and O(n) implementation from [here](https://github.com/kevindelgado/structured-merge-diff/blob/double-pointer-leaves/fieldpath/set.go#L209-L246).

You can see the code I tested this with [here](https://github.com/kevindelgado/structured-merge-diff/blob/binary-search-leaves/fieldpath/set.go#L209-L231) (in case it's not optimal or what you expected).

The (almost) identical benchmark is:
```
BenchmarkFieldSet/leaves-20-16                            319598              3893 ns/op            2250 B/op         41 allocs/op
BenchmarkFieldSet/leaves-50-16                             81356             14551 ns/op            6851 B/op        140 allocs/op
BenchmarkFieldSet/leaves-100-16                            21072             58362 ns/op           29149 B/op        632 allocs/op
BenchmarkFieldSet/leaves-500-16                             3243            365927 ns/op          146657 B/op       3215 allocs/op
BenchmarkFieldSet/leaves-1000-16                            1417            844380 ns/op          325139 B/op       7190 allocs/op
```

It seems like the random set generator might not ever be testing the worst case, unless there is some under the hood optimizations going on that make the O(n), O(n log n) and O(n^2) approaches all actually do the same thing.

kevindelgado: In summary,

benchmark results:
O(n) == O(n log n) == O(n^2)

so I went with O(n^2) because for readability because:
O(n^2) > [O(n log n)](https://github.com/kevindelgado/structured-merge-diff/blob/binary-search-leaves/fieldpath/set.go#L209-L231) > [O(n)](https://github.com/kevindelgado/structured-merge-diff/blob/double-pointer-leaves/fieldpath/set.go#L209-L246)

I can look into why the random set generation used by the benchmark test is not testing the worst case, but if we're short on time and need to land this asap, we might be better just picking one for now and optimizing later?

kevindelgado: @jpbetz @apelisse thoughts^

apelisse: I'll defer to @lavalamp on this.

apelisse: This file uses the double pointer logic about everywhere (and for good reasons), so I'd still go with that if possible. I've tried to write it, it's not that bad.

fieldpath/set.go
```diff
@@ -206,6 +206,29 @@ func (s *Set) WithPrefix(pe PathElement) *Set {
 	return subset
 }
 
+// Leaves returns a set containing only the leaf paths
+// of a set.
+func (s *Set) Leaves() *Set {
+	members := MakePathElementSet(len(s.Members.members))
```
apelisse, line 212: This one is more likely to over-allocate since members are going to be filtered.

fieldpath/set.go
kevindelgado: done, ptal @apelisse

fieldpath/set.go
kevindelgado: reverted

fieldpath/set.go
```diff
@@ -206,6 +206,45 @@ func (s *Set) WithPrefix(pe PathElement) *Set {
...
+	for im < len(s.Members.members) {
+		member := s.Members.members[im]
+		if ic >= len(s.Children.members) {
+			leaves.members = append(leaves.members, member)
+			im++
+			continue
+		}
+
+		diff := member.Compare(s.Children.members[ic].pathElement)
+		if diff-prevDiff != 0 {
```
lavalamp, line 227: This subtraction is really confusing to me, I guess I expected something like:

```go
outer:
for im < len(s.Members.members) {
  member := s.Members.members[im]

  for ic < len(s.Children.members) {
    d := member.Compare(s.Children.members[ic].pathElement)
    if d == 0 {
      ic++
      im++
      continue outer
    } else if d < 0 {
      break
    } else /* if d > 0 */ {
      ic++
    }
  }
  leaves.members = append(leaves.members, member)
}
```

kevindelgado: Ya, that's way cleaner

kevindelgado:
All feedback addressed, O(n) Leaves method cleaned up.

PTAL @apelisse

apelisse (approved):
/lgtm
/approve

apelisse:
This adds a new method `Leaves()` in `Set` which is usually not considered a non-backward compatible change. Let's update the minor version.

PR #188, opened 2021-03-24
Extract items should only extract the exact paths it is supplied

### LINKED ISSUE #186: Modify the behavior of ExtractItems to return only exact paths
Currently when extracting a path such we extract everything recursively below a given path.

For example:
I apply with `“metadata”:{}`
Someone else adds `“metadata”: { “annotations”: {“foo”:”bar”}}`
When I extract with Fieldset `.metadata` we are going to return `metadata:{“annotations”:{“foo”:”bar”}}` but we want to instead return `metadata:{}`

So what this means for `ExtractItems` is that every path should be treated as a leaf and should empty out the contents of the final path element regardless of what is there.

Another example,
 if we have `”metadata”: {“annotations”: {“foo”: “bar”, “caz”: “dak”}}`
The only way to receive all the annotations should be to Extract a fieldset with both `.metadata.annotations.foo` and `.metadata.annotations.caz` because if you tried to extract `.metadata.annotations` we want this to return `”metadata”:{“annotations”:{}}}`

Atomic fields are an exception to this whereby extracting a path to an atomic field should return everything in that field.

kevindelgado:
## What
Passing a field to ExtractItems will now only extract that exact field, rather than cascade and extract the subfields owned by that field

## Why
There are situations where we apply a field with no values (such as `"metadata":{}`), when another fieldmanager comes along and applies `"metadata":{"annotations":{"foo":"bar"}}` we want the first manager to still only own the field `"metadata":{}` however it will currently own everything added under metadata.

## How
When the extract walker reaches a path that exactly matches a path in the fieldset it wants to extract, it will extract the field and its contents only in the case of a scalar field or an atomic list/map, in the case of a map or list, it will keep the field key but zero out any non key data.

## Testing
`typed/remove_test.go` updated to reflect current behavior
`merge/extract_apply_test.go` confirms no regressions and adds empty structure and atomic tests.

fixes #186


typed/helpers.go
```diff
@@ -221,6 +221,18 @@ func keyedAssociativeListItemToPathElement(a value.Allocator, s *schema.Schema,
 	return pe, nil
 }
 
+func retainOnlyListKeys(keys []string, child value.Map) {
```
kevindelgado, line 224: I'm not a huge fan of this helper, but left it as is until I could get feedback on it.

The alternative I was thinking of was to add a `RetainAll` method to the `value.Map` interface that would do the filtering in place on the map. 

The issue there is that there are a ton of implementations of that Map interface and adding to all of them seemed to get a bit hairy. If we do it that way, it feels like we should be able to do better than O(mapKeys * listKeys), maybe by iterating through list keys, creating a new map with just them and changing the old map to be the new map? Kind of just thinking out loud here. 

Not sure if that would work though because I don't think you can just replace a receiver with a new map you create, right?

Anyways, rambling aside, wanted to call this out to get any thoughts you have here.

typed/remove.go
```diff
@@ -118,7 +126,17 @@ func (w *removingWalker) doMap(t *schema.Map) ValidationErrors {
 		// but ignore them when we are removing (i.e. !w.shouldExtract)
 		if w.toRemove.Has(path) {
 			if w.shouldExtract {
-				newMap[k] = val.Unstructured()
+				valIsAtomic, err := isAtomic(val, w.schema, fieldType)
```
kevindelgado, line 129: Let me know if we can just ignore this err and don't have to do the whole errors collection rigamarole. I'm thinking we can because [`resolveSchema()`](https://github.com/kubernetes-sigs/structured-merge-diff/pull/188/files#diff-9bfd5d4edd67f21a522024f4f5725b30dffdceada9af3fee0310ab699755401cR44) does the `schema.Resolve(typeRef)` call that has the potential to error out, and we just ignore it in that case.

kevindelgado:
PR updated, PTAL @apelisse

typed/remove.go
```diff
@@ -92,12 +104,24 @@ func (w *removingWalker) doList(t *schema.List) (errs ValidationErrors) {
 }
 
 func (w *removingWalker) doMap(t *schema.Map) ValidationErrors {
+	if !w.value.IsMap() {
```
apelisse, line 107: I'm curious what use-case this covers? I'm guessing if we have for example. In that case both remove or extract can't do anything beside return nil . Is that right?

apelisse:
Looks good to me!

typed/remove.go
kevindelgado: Ya, the use case we see is any time an extract path matches to a a top-level struct that maps to a nil value.

In t[his example](https://github.com/kubernetes-sigs/structured-merge-diff/pull/188/files#diff-7f11c34fef592894fdf658dda9b43b5ee8f036a06620052c41b5144805cfd9f2R692-R699), we have a match on path `.c` so in `m.Iterate` we hit the [new recursion](https://github.com/kubernetes-sigs/structured-merge-diff/pull/188/files#diff-9bfd5d4edd67f21a522024f4f5725b30dffdceada9af3fee0310ab699755401cR145) we added to call `removeItemsWithSchema()` for the key `.c` and the val `&{Value:<nil>}` 

When we recurse into `doMap` now [the line](https://github.com/kubernetes-sigs/structured-merge-diff/pull/188/files#diff-9bfd5d4edd67f21a522024f4f5725b30dffdceada9af3fee0310ab699755401cR110) `m := w.value.AsMapUsing(w.allocator)` will panic if we don't do the isMap check first.

PR #192, opened 2021-03-31
Unsupport atomic to granular schema reconciliation

### LINKED ISSUE #191: Schema Reconciler inaccurately determines whether the node is atomic or not
Currently the `reconcileWithSchemaWalker` will believe that a list/map node has changed from atomic to granular if the node was previously atomic (and the current element relationship is not atomic).

The way it [determines whether a node is atomic](https://github.com/kubernetes-sigs/structured-merge-diff/blob/master/typed/reconcile_schema.go#L173) is by naively assuming that if the node’s fieldSet contains an path element that is a member, but that path element has no children, than the node must be atomic.

This is incorrect because there are situations where a node might be an empty list/map, but not be atomic. For example, if a manager wants to own just the `spec.metadata` field to ensure it doesn't get removed, but does not want to own any of its contents.

The guiding example that exposes this issue is the following proposed test to be added to the merge package

```
                // BROKEN
                "no_change_list": {
                        Ops: []Operation{
                                // Apply an empty list with the first manager
                                // Meaning we want the manager to own the list itself
                                // but none of its elements
                                Apply{
                                        Manager: "one",
                                        Object: `
                                                list:
                                        `,
                                        APIVersion: "v1",
                                },
                                // Apply specific elements to the list created by manager one.
                                // Manager two should own these individual elements in the list.
                                // Manager one should continue to own the list itself but none of the elements.
                                Apply{
                                        Manager: "two",
                                        Object: `
                                                list:
                                                 - a
                                                 - b
                                        `,
                                        APIVersion: "v1",
                                },
                                // Prior to this apply, manager one SHOULD own the list but none of the individual
                                // elements, but when we use manager two to remove element "a" from the list,
                                // we see that actually manager one becomes an owner of all the elements in the list.
                                //
                                // This occurs because `reconcileWithSchemaWalker` thinks the list's schema has
                                // changed from atomic to granular.
                                //
                                // It was never atomic in the first place, but the reconciler determines node
                                // atomicity by checking if a node (in this case the list itself) exists as a member in the
                                // fieldSet but this node has no children (because the list is empty).
                                Apply{
                                        Manager: "two",
                                        Object: `
                                                list:
                                                 - b
                                        `,
                                        APIVersion: "v1",
                                },
                        },
                        // BROKEN: expected:
                        //Object: `
                        //      list:
                        //      - b
                        //`,
                        // but actually got:
                        Object: `
                                list:
                                - a
                                - b
                        `,
                        APIVersion: "v1",
                        Managed: fieldpath.ManagedFields{
                                "apply-one": fieldpath.NewVersionedSet(
                                        // BROKEN expected:
                                        //_NS(
                                        //      _P("list"),
                                        //),
                                        // but actually got:
                                        _NS(
                                                _P("list"),
                                                _P("list", _V("a")),
                                                _P("list", _V("b")),
                                        ),
                                        "v1",
                                        false,
                                ),
                                "apply-two": fieldpath.NewVersionedSet(
                                        _NS(
                                                _P("list", _V("b")),
                                        ),
                                        "v1",
                                        false,
                                ),
                        },
                },
```

A perfectly correct solution may not be possible because there is no way of determining what the previous schema was. 

One proposal is to maintain a holistic view of things we know not to be atomic (because some other applier was able to apply individual elements) and then for our atomicity check we look at that holistic view to see whether or not any node was previously not atomic.

While not perfect, this is more correct than the current situation.

kevindelgado:
Fixes #191 by removing support for atomic to granular schema changes.

Now, when a field on a schema is changed from atomic to granular, the schema will be treated as if it was always granular, potentially causing existing managers to own less fields than they did previously. Previously, they would own all the contents of an atomic list/map field, now they will only own the top level field itself and none of its contents.

This was determined to be less worse than incorrectly assuming empty fields were previously atomic (and causing managers to own more fields than they should).

Following this we plan to:
1. Document this behavior in upstream kubernetes documentation.
2. Potentially revisit larger sweeping changes that could enable this support in the future (i.e. making the fieldset aware of atomicity, creating a separate "defaulting" manager to handle defaulted fields atomicity, etc)


kevindelgado:
I went ahead and removed the atomic-to-granular testing that we had. Let me know if you think there's any testing we need to add with this reversion. 

It seemed odd to have tests to explicitly prove we don't support this flow, but let me know if you think anything like that should be added to this change.

FWIW testing to confirm no new fields are added to a top level apply manager will be captured in the testing for the Extract changes (where that behavior is needed), but don't seem super relevant here or in any of the schema reconciliation tests.

typed/reconcile_schema.go
```diff
@@ -231,21 +226,16 @@ func (v *reconcileWithSchemaWalker) visitMapItems(t *schema.Map, element *fieldp
...
 	if !v.isAtomic && t.ElementRelationship == schema.Atomic {
 		if v.fieldSet != nil && v.fieldSet.Size() > 0 {
 			v.toRemove = fieldpath.NewSet(v.path) // remove all root and all children fields
 			v.toAdd = fieldpath.NewSet(v.path)    // add the root of the atomic
 		}
 		return errs
 	}
-	// reconcile maps changed from atomic to granular
-	if v.isAtomic && (t.ElementRelationship == schema.Separable || t.ElementRelationship == "") {
-		v.toAdd, errs = buildGranularFieldSet(v.path, v.value)
```
jpbetz, line 244: Is `buildGranularFieldSet` still used anywhere?

merge/schema_change_test.go
```diff
@@ -178,75 +178,3 @@ func TestGranularToAtomicSchemaChanges(t *testing.T) {
 		})
 	}
 }
-
-func TestAtomicToGranularSchemaChanges(t *testing.T) {
```
apelisse, line 182: Can we actually keep the test and update it with the new behavior? If we document what happens when someone goes from atomic to granular, then we should have a test to confirm that behavior.

typed/reconcile_schema.go
```diff
@@ -187,19 +187,14 @@ func (v *reconcileWithSchemaWalker) visitListItems(t *schema.List, element *fiel
 }
 
 func (v *reconcileWithSchemaWalker) doList(t *schema.List) (errs ValidationErrors) {
-	// reconcile lists changed from granular to atomic
+	// reconcile lists changed from granular to atomic.
+	// Note that migrations from atomic to granular are unsuported and will
```
apelisse, line 191: I'd change from "unsported" to "unrecommended", and I'm not sure I understand the consequences from that sentence. (comment also applies to below for maps)

typed/reconcile_schema.go
kevindelgado: nope, good catch

typed/reconcile_schema_test.go
```diff
@@ -207,39 +207,6 @@ var reconcileCases = []reconcileTestCase{{
 		_P("stringMap"),
 		_P("unchanged", "numeric"),
 	),
-}, {
```
apelisse, line 210: removal of this test looks good

merge/schema_change_test.go
jpbetz: Seems useful to me as well. Might also be worth adding a comment in the tests explaining the behavior further.

kevindelgado: done, let me know if the comments are sufficient

typed/reconcile_schema.go
kevindelgado: done

merge/schema_change_test.go
```diff
@@ -231,9 +221,17 @@ func TestAtomicToGranularSchemaChanges(t *testing.T) {
...
 			Managed: fieldpath.ManagedFields{
+				// Note that manager one previously owned
+				// the top level _P("struct")
+				// which included all of its subfields
+				// when the struct field was atomic.
+				//
+				// Upon changing the schema of struct from
+				// atomic to granular, manager one continues
+				// to own the same fielset as before,
+				// but does not retain ownership of any of the subfields.
```
jpbetz, line 232: Also make it clear that this is a known limitation and not the ideal beahvior?

kevindelgado: added

kevindelgado:
lgtm removed b/c I fixed a typo

typed/reconcile_schema.go
```diff
@@ -187,19 +187,17 @@ func (v *reconcileWithSchemaWalker) visitListItems(t *schema.List, element *fiel
 }
 
 func (v *reconcileWithSchemaWalker) doList(t *schema.List) (errs ValidationErrors) {
-	// reconcile lists changed from granular to atomic
+	// reconcile lists changed from granular to atomic.
+	// Note that migrations from atomic to granular are not recommended and will
+	// be treated as if they were always granular.
+	//
+	// In this case, the manager that owned the previously atomic field (and all subfields),
+	// will now own just the top-level field and none of the subfields.
```
apelisse, line 195: Perfect

apelisse:
/approve

Do we have other plans for SMD or should I cut a release once this is merged?

PR #194, opened 2021-04-13
Fix typo iff

zhijianli88:
structured-merge-diff$ sed -i 's/ iff / if /g' $(git grep -lw iff)

Signed-off-by: Li Zhijian <lizhijian@fujitsu.com>

### DIFF
--- a/schema/equals.go
+++ b/schema/equals.go
@@ -18,7 +18,7 @@ package schema
 
 import "reflect"
 
-// Equals returns true iff the two Schemas are equal.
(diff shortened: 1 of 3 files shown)


apelisse:
Thanks, I don't think these are typos, c.f. https://en.wikipedia.org/wiki/If_and_only_if

/close

lavalamp:
Yeah iff == if and only if, it's at least somewhat common in texts that might need to use that concept...

PR #200, opened 2021-06-25
Don't process schemaless fields with ReconcileFieldSetWithSchema

jpbetz:
Mitigates https://github.com/kubernetes/kubernetes/issues/102749 by avoiding traversal of:
- CRD schemas that are deduced
- Branches of schemas where x-kubernetes-preserve-unknown-fields=true and there are no fields defined

For example, this will skip traversal for:

```yaml
apiVersion: apiextensions.k8s.io/v1
kind: CustomResourceDefinition
metadata:
  name: example.k8s.io
spec:
  group: k8s.io
  # see https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#field-pruning
  preserveUnknownFields: true
  versions:
  - name: v1
    served: true
    storage: true
```

and

```yaml
apiVersion: apiextensions.k8s.io/v1
kind: CustomResourceDefinition
...
schema:
  openAPIV3Schema:
    properties:
      spec:
        type: object
        x-kubernetes-preserve-unknown-fields: true
      status:
        type: object
        x-kubernetes-preserve-unknown-fields: true
```

But still do traversal for:

```yaml
apiVersion: apiextensions.k8s.io/v1
kind: CustomResourceDefinition
...
schema:
  openAPIV3Schema:
    properties:
      spec:
        type: object
        x-kubernetes-preserve-unknown-fields: true
        properties:
          deadline:
             format: int64
             type: integer
...
```

Additional details for reviewers
-----------------------------------------

Discovery OpenAPI schema for CRD with no schema:
- https://gist.github.com/jpbetz/9172e4620e39f9ca40b19f928af8576f

How the spec.preserveUnknownFields flags is used when converting from CRDs to discoverable openapi schemas:
- https://github.com/kubernetes/kube-openapi/blob/3c818078ee3de6569a8f02b6345ea3c4cc8b0998/pkg/schemaconv/smd.go#L289-L292

How openAPI schema's map to SMD schemas, for the case of x-kubernetes-preserve-unknown-fields:
- https://github.com/kubernetes/kube-openapi/pull/239

Manual test: https://gist.github.com/jpbetz/c06ee61492d70e47d9455af5f85dd9de

### DIFF
--- a/typed/reconcile_schema.go
+++ b/typed/reconcile_schema.go
@@ -124,13 +124,6 @@ func ReconcileFieldSetWithSchema(fieldset *fieldpath.Set, tv *TypedValue) (*fiel
 	v.schema = tv.schema
 	v.typeRef = tv.typeRef
 
-	// We don't reconcile deduced types, which are primarily for use by unstructured CRDs. Deduced
-	// types do not support atomic or granular tags. Nor does the dynamic schema deduction
-	// interact well with the reconcile logic.
-	if v.schema == DeducedParseableType.Schema {
-		return nil, nil
-	}
-
 	defer v.finished()
 	errs := v.reconcile()
 
@@ -229,6 +222,12 @@ func (v *reconcileWithSchemaWalker) visitMapItems(t *schema.Map, element *fieldp
 }
 
 func (v *reconcileWithSchemaWalker) doMap(t *schema.Map) (errs ValidationErrors) {
+	// We don't currently reconcile deduced types (unstructured CRDs) or maps that contain only unknown
+	// fields since deduced types do not yet support atomic or granular tags.
+	if isUntypedDeducedMap(t) {
+		return errs
+	}
+
 	// reconcile maps and structs changed from granular to atomic.
 	// Note that migrations from atomic to granular are not recommended and will
 	// be treated as if they were always granular.
@@ -273,4 +272,19 @@ func typeRefAtPath(t *schema.Map, pe fieldpath.PathElement) (schema.TypeRef, boo
 		}
 	}
 	return tr, tr != schema.TypeRef{}
+}
+
+// isUntypedDeducedMap returns true if m has no fields defined, but allows untyped elements.
+// This is equivalent to a openAPI object that has x-kubernetes-preserve-unknown-fields=true
+// but does not have any properties defined on the object.
+func isUntypedDeducedMap(m *schema.Map) bool {
+	return isUntypedDeducedRef(m.ElementType) && m.Fields == nil
+}
+
+func isUntypedDeducedRef(t schema.TypeRef) bool {
+	if t.NamedType != nil {
+		return *t.NamedType == "__untyped_deduced_"
+	}
+	atom := t.Inlined
+	return atom.Scalar != nil && *atom.Scalar == "untyped"
 }
--- a/typed/reconcile_schema_test.go
+++ b/typed/reconcile_schema_test.go
@@ -98,6 +98,66 @@ func granularSchema(version string) typed.YAMLObject {
     - name: numeric
       type:
         scalar: numeric
+- name: empty
+  map:
+    elementType:
+      scalar: untyped
+      list:
+        elementType:
+          namedType: __untyped_atomic_
+        elementRelationship: atomic
+      map:
+        elementType:
+          namedType: __untyped_deduced_
+        elementRelationship: separable
+- name: emptyWithPreserveUnknown
+  map:
+    fields:
+    - name: preserveField
+      type:
+        map:
+          elementType:
+            scalar: untyped
+            list:
+              elementType:
+                namedType: __untyped_atomic_
+              elementRelationship: atomic
+            map:
+              elementType:
+                namedType: __untyped_deduced_
+              elementRelationship: separable
+- name: populatedWithPreserveUnknown
+  map:
+    fields:
+    - name: preserveField
+      type:
+        map:
+          fields:
+          - name: list
+            type:
+              namedType: list
+          elementType:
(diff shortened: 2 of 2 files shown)


typed/reconcile_schema.go
```diff
@@ -229,6 +222,13 @@ func (v *reconcileWithSchemaWalker) visitMapItems(t *schema.Map, element *fieldp
 }
 
 func (v *reconcileWithSchemaWalker) doMap(t *schema.Map) (errs ValidationErrors) {
+	// We don't reconcile deduced types (unstructured CRDs) or maps that contain only unknown
+	// fields. Deduced types do not support atomic or granular tags. Nor does the dynamic
+	// schema deduction interact well with the reconcile logic.
+	if isUntypedDeducedMap(t) { // We don't traverse into maps that contain only unknown fields.
+		return
```
nickgerace, line 229: Should this return `nil`?

nickgerace (commented):
Reviewed for syntax only. Looks good from here.

lavalamp:
The gist says it's still slow, approximately what percentage of the problem does this fix?

jpbetz:
> The gist says it's still slow, approximately what percentage of the problem does this fix?

It resolves the regression introduced in 1.19.10. Requests still take 8.5s locally, which is on par with what they took on 1.19.9 without this fix:

```
kubectl apply -f workflow.json --validate=false  2.71s user 0.17s system 34% cpu 8.411 total
```

lavalamp:
8.5s is too long. Eventually I actually would like make schema deduction behave reasonably. Do you know what the next biggest culprit is?

jpbetz:
> 8.5s is too long. 

I agree. For this PR I've been trying to focus on the regression.

> Eventually I actually would like make schema deduction behave reasonably. Do you know what the next biggest culprit is?

I'm fairly confident it's the validation pass. I've got a benchmark from yesterday that suggests that. I can spend some more time with that and try to break down the latency in more detail. One warning: we've already done a couple performance tuning passes on validation (circa late 2019), so all the low hanging fruit may have already been picked. Regardless, I can start digging in.

lavalamp:
Perhaps we should merge this and just do a part 2 when we solve the other problem?

typed/reconcile_schema.go
```diff
@@ -229,6 +222,13 @@ func (v *reconcileWithSchemaWalker) visitMapItems(t *schema.Map, element *fieldp
 }
 
 func (v *reconcileWithSchemaWalker) doMap(t *schema.Map) (errs ValidationErrors) {
+	// We don't reconcile deduced types (unstructured CRDs) or maps that contain only unknown
+	// fields. Deduced types do not support atomic or granular tags. Nor does the dynamic
```
lavalamp, line 226: Can you change this wording to make it clear it's not necessarily the end of the story for deduced types, like maybe we'll fix it? (I know this was a copy-paste, sorry)

typed/reconcile_schema.go
```diff
@@ -274,3 +274,18 @@ func typeRefAtPath(t *schema.Map, pe fieldpath.PathElement) (schema.TypeRef, boo
...
+// isUntypedDeducedMap returns true if m has no fields defined, but allows untyped elements.
+// This is equivalent to a openAPI object that has x-kubernetes-preserve-unknown-fields=true
+// but does not have any properties defined on the object.
+func isUntypedDeducedMap(m *schema.Map) bool {
+	return isUntypedDeducedRef(m.ElementType) && m.Fields == nil
+}
+
+func isUntypedDeducedRef(t schema.TypeRef) bool {
+	if t.NamedType != nil {
+		return *t.NamedType == "__untyped_deduced_"
```
lavalamp, line 287: There are two underscores at the beginning but only one at the end. Is that intentional?

typed/reconcile_schema_test.go
```diff
@@ -98,6 +98,66 @@ func granularSchema(version string) typed.YAMLObject {
...
+  map:
+    elementType:
+      scalar: untyped
+      list:
+        elementType:
+          namedType: __untyped_atomic_
+        elementRelationship: atomic
+      map:
+        elementType:
+          namedType: __untyped_deduced_
```
lavalamp, line 111: Does this already exist in our real schemas?

jpbetz: Yes, here's a generated SMD schema (and it's corresponding openAPI schema) to show exactly type of schemas we're expecting: https://github.com/kubernetes/kube-openapi/pull/239/files#diff-6ba5d14994ba13458a380d3cf2e3b55bc5e15a823e34dc648bbe5ae442679254R52

typed/reconcile_schema.go
jpbetz: Yes, this is now it is defined.

xref: https://github.com/kubernetes/kube-openapi/blob/3c818078ee3de6569a8f02b6345ea3c4cc8b0998/pkg/schemaconv/smd.go#L138, https://github.com/kubernetes-sigs/structured-merge-diff/blob/ea1021dbc0f242313159d5dd4801ff29304712fe/typed/parser.go#L130

typed/reconcile_schema.go
jpbetz: Sure, I've rephrased it.

typed/reconcile_schema.go
lavalamp: *eye twitch*

nickgerace:
Thank you very much! @jpbetz

apelisse:
I'm surprised that we do see any output of `go bench` in the chat of this. We're supposed to fix a performance issue and we haven't tested that this was solved anywhere?

PR #204, opened 2021-10-15
Fix potential conversion failure in addBackOwnedItems

zqzten:
Currently `addBackOwnedItems` adds back managed fields by API version in a random order (for range a map), consider the following scenario:

We have an object with two API versions: v1 and v2, and use a conversion webhook to convert between the two versions. When converting v2 to v1, the conversion webhook rely on a required field **foo** of v2.

Now there're several users use server-side apply to manage this object in different versions, when some user initiate a server-side apply to the object in v2, `addBackOwnedItems` would have to convert and add back fields to the pruned object in both v1 and v2. At this time, the order matters:

If `addBackOwnedItems` converts and adds back fields to v2 first then v1, everything will go smoothly, however, if the order is reversed, a `failed to convert pruned object at version v1` error will arise and the server-side apply will fail. This is because when `addBackOwnedItems` convert the pruned v1 object, the required field **foo** of v2 hasn't been added back and will cause the conversion from v2 to v1 to fail.

This PR indroduce a retry loop to `addBackOwnedItems` to ensure that when it convert the pruned object to a specific version, the object already has all the required fields back.

### DIFF
--- a/merge/update.go
+++ b/merge/update.go
@@ -250,33 +250,54 @@ func (s *Updater) addBackOwnedItems(merged, pruned *typed.TypedValue, managedFie
 		}
 		managedAtVersion[managerSet.APIVersion()] = managedAtVersion[managerSet.APIVersion()].Union(managerSet.Set())
 	}
+	// Add back owned items at pruned version first to avoid conversion failure
+	// caused by pruned fields which are required for conversion.
+	prunedVersion := fieldpath.APIVersion(*pruned.TypeRef().NamedType)
+	if managed, ok := managedAtVersion[prunedVersion]; ok {
+		merged, pruned, err = s.addBackOwnedItemsForVersion(merged, pruned, prunedVersion, managed)
+		if err != nil {
+			return nil, err
+		}
+		delete(managedAtVersion, prunedVersion)
+	}
 	for version, managed := range managedAtVersion {
-		merged, err = s.Converter.Convert(merged, version)
+		merged, pruned, err = s.addBackOwnedItemsForVersion(merged, pruned, version, managed)
 		if err != nil {
-			if s.Converter.IsMissingVersionError(err) {
-				continue
-			}
-			return nil, fmt.Errorf("failed to convert merged object at version %v: %v", version, err)
+			return nil, err
 		}
-		pruned, err = s.Converter.Convert(pruned, version)
-		if err != nil {
-			if s.Converter.IsMissingVersionError(err) {
-				continue
-			}
-			return nil, fmt.Errorf("failed to convert pruned object at version %v: %v", version, err)
-		}
-		mergedSet, err := merged.ToFieldSet()
-		if err != nil {
-			return nil, fmt.Errorf("failed to create field set from merged object at version %v: %v", version, err)
-		}
-		prunedSet, err := pruned.ToFieldSet()
-		if err != nil {
-			return nil, fmt.Errorf("failed to create field set from pruned object at version %v: %v", version, err)
-		}
-		sc, tr := merged.Schema(), merged.TypeRef()
-		pruned = merged.RemoveItems(mergedSet.EnsureNamedFieldsAreMembers(sc, tr).Difference(prunedSet.EnsureNamedFieldsAreMembers(sc, tr).Union(managed.EnsureNamedFieldsAreMembers(sc, tr))))
 	}
 	return pruned, nil
+}
+
+// addBackOwnedItemsForVersion adds back any fields, list and map items that were removed by prune with specific managed field path at a version.
+// It is an extracted sub-function from addBackOwnedItems for code reuse.
+func (s *Updater) addBackOwnedItemsForVersion(merged, pruned *typed.TypedValue, version fieldpath.APIVersion, managed *fieldpath.Set) (*typed.TypedValue, *typed.TypedValue, error) {
+	var err error
+	merged, err = s.Converter.Convert(merged, version)
+	if err != nil {
+		if s.Converter.IsMissingVersionError(err) {
+			return merged, pruned, nil
+		}
+		return nil, nil, fmt.Errorf("failed to convert merged object at version %v: %v", version, err)
+	}
+	pruned, err = s.Converter.Convert(pruned, version)
+	if err != nil {
+		if s.Converter.IsMissingVersionError(err) {
+			return merged, pruned, nil
+		}
+		return nil, nil, fmt.Errorf("failed to convert pruned object at version %v: %v", version, err)
+	}
+	mergedSet, err := merged.ToFieldSet()
+	if err != nil {
+		return nil, nil, fmt.Errorf("failed to create field set from merged object at version %v: %v", version, err)
+	}
+	prunedSet, err := pruned.ToFieldSet()
+	if err != nil {
+		return nil, nil, fmt.Errorf("failed to create field set from pruned object at version %v: %v", version, err)
+	}
+	sc, tr := merged.Schema(), merged.TypeRef()
+	pruned = merged.RemoveItems(mergedSet.EnsureNamedFieldsAreMembers(sc, tr).Difference(prunedSet.EnsureNamedFieldsAreMembers(sc, tr).Union(managed.EnsureNamedFieldsAreMembers(sc, tr))))
+	return merged, pruned, nil
 }
 
 // addBackDanglingItems makes sure that the fields list and map items removed by prune were
--- a/merge/multiple_appliers_test.go
+++ b/merge/multiple_appliers_test.go
@@ -1814,3 +1814,134 @@ func BenchmarkMultipleApplierRecursiveRealConversion(b *testing.B) {
 		}
 	}
 }
+
+var multiversionWithReliantFieldsParser = func() Parser {
+	parser, err := typed.NewParser(`types:
+- name: v1
+  map:
+    fields:
+      - name: field_foo_rely_on_bar
+        type:
+          scalar: string
+      - name: common_field
+        type:
+          scalar: string
+- name: v2
+  map:
+    fields:
+      - name: required_field_bar
+        type:
+          scalar: string
+      - name: common_field
+        type:
+          scalar: string
+`)
+	if err != nil {
+		panic(err)
+	}
+	return parser
+}()
+
+// reliantFieldsConverter converts v2 obj to v1 relying on the required_field_bar,
+// if required_field_bar is empty, the conversion shall fail.
+// This converter can only be used with multiversionWithReliantFieldsParser.
+type reliantFieldsConverter struct {
+}
+
+var _ merge.Converter = reliantFieldsConverter{}
+
+func (r reliantFieldsConverter) Convert(v *typed.TypedValue, version fieldpath.APIVersion) (*typed.TypedValue, error) {
+	inVersion := fieldpath.APIVersion(*v.TypeRef().NamedType)
+	if inVersion == version {
+		return v, nil
+	}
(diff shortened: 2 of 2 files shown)


apelisse:
I haven't look yet, but a test would be super helpful here, otherwise it's hard to know that it works. Especially since this code path is very complex.

apelisse:
Also, a good example of how to reproduce would be helpful to understand what's going on! Thanks :-)

zqzten:
@apelisse Thanks for the suggestion! I have written a minimal [real example](https://github.com/zqzten/examples/tree/main/conversion-webhook) to reproduce it. Here's the reproduce procedure:
1. Install the above example to your cluster.
2. Run `kubectl apply --server-side --field-manager=user2 -f config/samples/conversion-webhook_v2_networkaddress.yaml`. It simulates that user2 apply a v2 CR into the cluster and become its sole manager.
3. Run `kubectl apply --server-side --field-manager=user1 -f config/samples/conversion-webhook_v1_networkaddress_patch.yaml`. It simulates that user1 use version v1 to apply the above CR and become another manager of it.
4. Now the CR have two managers using different versions (v1 and v2), run `kubectl apply --server-side --field-manager=user2 -f config/samples/conversion-webhook_v2_networkaddress_patch.yaml` to simulate that user2 want to update the CR it applied in step 2.
There're no conflicts so we expect this apply will be done as usual. However, here's the bug comes... If you're lucky, this apply will succeed, then you can rerun it several times and, at last, you'll definitely fail with an error `Error from server: failed to prune fields: failed add back owned items: failed to convert pruned object at version conversion-webhook.example.zqzten.io/v1: conversion webhook for conversion-webhook.example.zqzten.io/v2, Kind=NetworkAddress failed: invalid address value:`
After apply this PR's fix, the above error can be eliminated so that the apply above will always work well as expected.

Hope the above example in addition to the PR's description can help you understand what's going on :)

Speaking of the test, since the random nature of this bug, I'm not sure what's the best way to test a case like this, so I solicit your further suggestion, thanks again.

apelisse:
Thanks so much for the detailed answer!

I think the next step is to create a unit-test that reproduces exactly what you encountered without the randomness, I'm pretty sure that should be doable (or at least that reproduces more consistently). Once this does, then we can be sure that you're code fix is correct. Thanks a lot for taking the time to look at this1

apelisse:
After step 3, you should have:

user2 owns address and user1 owns domain and port, right?
In step 4, user2 changes the address (and the domain for user1) which should be a conflict for user1 (they don't expect domain to change), so

> There're no conflicts so we expect this apply will be done as usual. 

This seems to be the bug in my opinion?

jiahuif:
`pruned` is an empty object. If `pruned `is first converted to v1, then nothing happens (becuase v1 conversion is no-op), but
if pruned is converted to v2, the conversion webhook cannot handle empty object.

The error depends on which of the two versions is first used.

The "root" problem is that the conversion webhook could not handle empty object, or to be exact, an empty address field.
Technically, "pruned" object may be invalid because it lacks required fields. In our case, "address" field is required but not in the "pruned" object.

zqzten:
> After step 3, you should have: user2 owns address and user1 owns domain and port, right?
@apelisse I'm afraid it's not the case. In fact, user1 does not own `domain` after step3 (because it does not apply for it, you can check `config/samples/conversion-webhook_v1_networkaddress_patch.yaml` for sure), so there's no conflicts actually.

@jiahuif 's comment is absolutely right 👍 The root cause of this bug is that some conversion webhook could not handle empty object so that it might fail on converting the pruned object.

zqzten:
Sorry for holding this PR for long. I have added a unit-test for this to prove that my change is working as expected (the test will fail without my change), so the code is ready for review now, PTAL @apelisse cc @jiahuif 

P.S. I cannot 100% avoid the randomness in test, so I decided to let it run enough times to get as consistent results as possible. I've run the test for many times and found that it is able to produce consistent result in pratical.

zqzten:
Friendly ping @apelisse to preceed with the code review process. Any feedback is appreciated!

apelisse:
We're planning on addressing this very soon, but no time right now. Thanks for your contribution!

merge/update.go
```diff
@@ -250,31 +250,47 @@ func (s *Updater) addBackOwnedItems(merged, pruned *typed.TypedValue, managedFie
 		}
 		managedAtVersion[managerSet.APIVersion()] = managedAtVersion[managerSet.APIVersion()].Union(managerSet.Set())
 	}
-	for version, managed := range managedAtVersion {
-		merged, err = s.Converter.Convert(merged, version)
```
lavalamp, line 254: Can this be more simply fixed by just not modifying `merged` and starting from the original form of `merged` each iteration? There should really not be path-dependencies sneaking in from the random iteration order of a loop like this.

lavalamp: (answering myself: no)

merge/update.go
```diff
@@ -250,31 +250,47 @@ func (s *Updater) addBackOwnedItems(merged, pruned *typed.TypedValue, managedFie
...
-		if err != nil {
-			if s.Converter.IsMissingVersionError(err) {
+	// We need a retry loop here to ensure that when we convert the pruned object
+	// to a specific version, it already has all the required fields back.
+	for {
```
lavalamp, line 255: In the very worst case, where there's exactly one conversion sequence that works, this retry loop's runtime complexity appears to be N! for N versions -- I think we can't afford that even though the average case is not that bad.

I *do* think we can afford to sort by version name so that it's at least deterministic.

I guess we don't hit this problem on built in types because we know how to default them? This is the part I'm still a bit confused about.

If determinism + asking conversion webhooks to first default isn't enough, then we need to cap the number of tries this loop can make, and ideally-- cache the successful order so that we only have to figure it out once. But I hope we don't have to do that.

lavalamp:
Many thanks for noticing this and for the test, that's super helpful.

I don't feel like this is quite the root cause though yet? My leading candidates are:
* we forgot to tell conversion webhook authors that they need to default first (we could actually have the extension-apiserver test this about them and put on a status condition if they don't handle converting a blank object)
* we forgot to implement defaulting based on the CRD schema
* Could be user error if the CRD schema lacks defaulting that it should have

I *know* conversion functions for built-in objects have similar requirements sometimes and we got around that with some clever ordering IIRC.

merge/update.go
zqzten: About the complexity, I think the worst case is O(N^2) rather than N!. The retry loop does not traverse the permutation of versions, instead, it does at least one successful conversion in each iteration (will fail fast if not) and then removes that converted version and does the next iteration. So the actual iteration count for the worst case would be 1+2+3+...+N which is N(N+1)/2.

It's hard to imagine there would be a resource type with more than 10 versions, let alone more. So it seems that the solution's performance is not that bad?

zqzten:
@lavalamp Thanks for the analysis!

Defaulting might be an applicabe solution in simple cases, but when the conversion logic becomes complicated, e.g, we use a field to decide the follwing conversion logic of other fields, if we apply a default value which dose not match the "pruned" one, it would change the other converted fields and may lead to some unexpected result. So maybe this solution is still more perferable than restrict the ability of conversion webhooks?

Also, I'm curious about below sentence:

> conversion functions for built-in objects have similar requirements sometimes and we got around that with some clever ordering

What's _clever ordering_ mean here? Where does the ordering happen? Can it also be applied to CRD conversions? Further explanation is appreciated.

Thanks again for the response!

merge/update.go
lavalamp: OK, you're right about the complexity, sorry, I missed where it removes the version on success. 

However, I think that doing that leaves cases that fail. Suppose there are four versions, and it's safe to convert in the order A->B->C->D, but most other conversions fail. Suppose the first order we try is B->C->A->D. B->C is successful, triggering removal of C, but now we're left with versions A, B, and D, and there's no possible order which will be successful (you can't get from B to D, nor can you get from D to A). So, the proposed fix only solves a subset of path-dependent conversion cases.

The current assumption enforced on built-ins is that we default before converting, and conversions must not ever fail (i.e., they are never path-dependent). I'm not sure we should relax this constraint at all.

zqzten: I'd like to clarify that the root cause of this failure is not the conversion order but the lack of **required** fields in a specific version. As long as an object is integral (each required field has a value) at a specific version, it can be converted to any other version successfully. So the conversion failure here is not really caused by a wrong order but by an imcomplete object (which is the **pruned** object) when converting. And what the fix do is not finding some order but making the pruned object integral at its current version (by adding back managed fields at this version).

Now we can return to the ABCD example, the problem is not caused by the A->B->C->D order (actually there's no conversion order limit) but by: there's a required non-empty field `fa` in A, and similarly `fb`, `fc`, `fd`. Now a pruned object comes to our loop, say it's at version A with `fa` pruned, and the loop order is B->C->A->D, here's what'll happen:
1. try convert to B, C, both fail because A is not integral (lack required field `fa`)
2. convert to A, success because there's actually no convert
3. add back owned items at version A (which include the required field `fa`), now the pruned obj is integral at version A and is ready to convert to any other version
4. convert to D, success because obj at version A is integral
5. add back owned items at version D (which include the required field `fd`), now the pruned obj is also integral at version D
6. then every following conversion will success because the obj will always be integral at its current version from now on

This bug is indeed a little tricky that it misled both of us lol. After think it twice and list the above step by step, I found I'm also wrong at the complexity, it's not O(N^2) but O(N)! The fix is not finding an order but to make sure the pruned obj is integral at first version (the version where the `pruned` parameter of the `addBackOwnedItems` function at) before it's converted to any other version. So I found an even better solution that we just get the version of `pruned` then do the add back logic on that version first and everything will go smoothly then. This makes the complexity of the fix become O(1). wdyt @lavalamp

lavalamp: Thanks for the great analysis! Your described fix sounds great, I can't object to an O(1) fix ;)

zqzten:
@lavalamp I have re-implemented the fix using the proposed O(1) solution, PTAL, thx.

merge/update.go
```diff
@@ -250,33 +250,54 @@ func (s *Updater) addBackOwnedItems(merged, pruned *typed.TypedValue, managedFie
 		}
 		managedAtVersion[managerSet.APIVersion()] = managedAtVersion[managerSet.APIVersion()].Union(managerSet.Set())
 	}
-	for version, managed := range managedAtVersion {
-		merged, err = s.Converter.Convert(merged, version)
+	// Add back owned items at pruned version first to avoid conversion failure
+	// caused by pruned fields which are required for conversion.
+	prunedVersion := fieldpath.APIVersion(*pruned.TypeRef().NamedType)
```
apelisse, line 255: I don't think that's correct, that's only an assumption we make in tests (where we use the name of the type as it's version), but once in kubernetes, the names of the type don't have a one-to-one mapping with their version.

apelisse: You could pass-in the version as a parameter I guess, since you have it in the `prune` function: `lastSet.APIVersion()`

lavalamp: Can we add a test that will catch this?

apelisse: We can, but the "internal/fixture/state" strongly expects the type to be equivalent to the version (it does so as a convenience so that the test framework doesn't have to know how to extract the version from the object, /shrug).

We could write a test, at least one, that doesn't depend on the "internal/fixture/state" and uses incompatible version/typename to detect these issues.

lavalamp: > write a test, at least one, that doesn't depend on the "internal/fixture/state" and uses incompatible version/typename to detect these issues

SGTM!

apelisse: I'm giving it a quick try right now

PR #206, opened 2021-12-04
Add new ApplyTwice operation benchmark

apelisse:
We don't have the best coverage for this.


apelisse:
Drawback: it'll merge the same lists
Positive: we don't really have good tests for that yet.

PR #210, opened 2022-01-09
Add test to verify that apply after update should work with no conflict

redbaron:
While investigating https://github.com/kubernetes/kubernetes/issues/107417 , I found that at least at the level of structured merge diff, `Apply` doesn't cause conflict with `Update` if manager name is the same.  This PR adds test for it.

### DIFF
--- a/merge/leaf_test.go
+++ b/merge/leaf_test.go
@@ -47,6 +47,37 @@ var leafFieldsParser = func() Parser {
 
 func TestUpdateLeaf(t *testing.T) {
 	tests := map[string]TestCase{
+		"update_apply": {
+			Ops: []Operation{
+				Update{
+					Manager: "default",
+					Object: `
+						numeric: 1
+					`,
+					APIVersion: "v1",
+				},
+				Apply{
+					Manager: "default",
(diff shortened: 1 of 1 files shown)


apelisse:
/ok-to-test

That test should fail right?

redbaron:
> That test should fail right?

No, it should succeed

apelisse:
Sorry, I was in a meeting when I wrote that. Apply and Update have been implemented/designed as being incompatible because they don't have the same mechanisms of owning fields, which means that they can't be converted from one to another. I don't understand how this test passes.

redbaron:
As described in the  https://github.com/kubernetes/kubernetes/issues/107417 Apply and Update are deliberately made incompatible  in kube apiserver, which I didn't agree with. This test case demonstrates that, for the structured merge purposes these operations are not in conflict same field manager can use both without introducing conflict with it's own doings.

apelisse:
I don't know what to do with this, I think we know this is how it works, but we don't rely on this behavior so I don't want to enforce that we keep it forever.
/close
We can re-open if we have stronger arguments.

PR #211, opened 2022-02-23
Skip comparison when value or object itself is null

ardaguclu:
When there is null field in left hand side(lhs) or right hand side(rhs),
comparison of TypedValue always thinks that these fields are added or removed.
That causes that time field is always updated as if some fields are changed.
However, no field is changed.

This PR adds extra check for both `lhs` and `rhs`. If both sides are
null, compare function will think that there is no change.

Fixes: https://github.com/kubernetes/kubernetes/issues/108234

cc: @apelisse

### DIFF
--- a/typed/typed.go
+++ b/typed/typed.go
@@ -126,8 +126,14 @@ func (tv TypedValue) Compare(rhs *TypedValue) (c *Comparison, err error) {
 	}
 	_, err = merge(&tv, rhs, func(w *mergingWalker) {
 		if w.lhs == nil {
+			if w.rhs == nil || w.rhs.IsNull() {
+				return
+			}
 			c.Added.Insert(w.path)
 		} else if w.rhs == nil {
+			if w.lhs == nil || w.lhs.IsNull() {
+				return
+			}
 			c.Removed.Insert(w.path)
 		} else if !value.Equals(w.rhs, w.lhs) {
 			// TODO: Equality is not sufficient for this.


apelisse:
I would have been surprised if that had worked. What is the actual problem you're trying to solve? Where is the timestamp being a problem?

ardaguclu:
Thanks @apelisse. I need some guidance then.

Execute below command and create service resource;

```bash
cat <<EOF | cluster/kubectl.sh apply -f - --server-side
apiVersion: v1
kind: Service
metadata:
  name: test
  namespace: default
spec:
  clusterIP: None
  ports:
    - name: http
      port: 443
  selector: null # Causes the issue
EOF
``` 
and expectedly it creates(**selector** field does not exist);

```json
{
  "kind": "Service",
  "apiVersion": "v1",
  "metadata": {
    "name": "test",
    "namespace": "default",
    "uid": "99bb7fe4-13bc-4026-8bff-3c3e8c5eac80",
    "resourceVersion": "377",
    "creationTimestamp": "2022-02-24T06:16:54Z",
    "managedFields": [
      {
        "manager": "kubectl",
        "operation": "Apply",
        "apiVersion": "v1",
        "time": "2022-02-24T06:16:54Z",
        "fieldsType": "FieldsV1",
        "fieldsV1": {
          "f:spec": {
            "f:clusterIP": {
              
            },
            "f:ports": {
              "k:{\"port\":443,\"protocol\":\"TCP\"}": {
                ".": {
                  
                },
                "f:name": {
                  
                },
                "f:port": {
                  
                }
              }
            },
            "f:selector": {
              
            }
          }
        }
      }
    ]
  },
  "spec": {
    "ports": [
      {
        "name": "http",
        "protocol": "TCP",
        "port": 443,
        "targetPort": 443
      }
    ],
    "clusterIP": "None",
    "clusterIPs": [
      "None"
    ],
    "type": "ClusterIP",
    "sessionAffinity": "None",
    "ipFamilies": [
      "IPv4",
      "IPv6"
    ],
    "ipFamilyPolicy": "RequireDualStack",
    "internalTrafficPolicy": "Cluster"
  },
  "status": {
    "loadBalancer": {
      
    }
  }
}
```
If you run same command multiple times, it always returns `service/test serverside-applied`(although there is no change).

In addition to that `diff --server-side` for same resource also returns time field has changed as stated in referenced issue.

After investigating the problem;

https://github.com/kubernetes/kubernetes/blob/016b96ca3896d27b37c5b2d6e223fb3320a0fdec/staging/src/k8s.io/apiserver/pkg/endpoints/handlers/fieldmanager/structuredmerge.go#L150 field returns not null `newObjTyped`.

Because https://github.com/kubernetes-sigs/structured-merge-diff/blob/65611e5f7fc60e909529889c29faa9cac17cb485/merge/update.go#L207 always returns `false`.

Because this line of code is always executed https://github.com/kubernetes-sigs/structured-merge-diff/blob/65611e5f7fc60e909529889c29faa9cac17cb485/typed/typed.go#L129. It is correct because `selector` field is null, problem is we are not checking that right hand side's `selector` field is also null.

(lhs is nil object, but rhs is unstructuredValue type whose value is nil, not itself).

If this is designed as is(just a trivial side effect of other things), I'll close my PR.

ardaguclu:
I investigated the problem further and noticed that there is no way to solve that problem(at least according to my current understanding). Since we are not storing null values, left hand side does not contain that null values. That results in there is no clue to distinguish whether user applies same null object in right hand side(as happened in referenced issue) or _actually_ applies new null field(as unit tests are failing in that PR).

Thereby, I'm closing this PR. Maybe someone who know better than me could take a look at.

/close

ardaguclu:
I don't know it is feasible but maybe we can check new and old objects equality in here https://github.com/kubernetes-sigs/structured-merge-diff/blob/65611e5f7fc60e909529889c29faa9cac17cb485/typed/typed.go#L293 besides `Removed`, `Modified` and `Added` sets.

apelisse:
Thanks, if we want to fix the original bug, we might want to do it a different way.

PR #215, opened 2022-03-31
add CopyInto method to schema.Schema

alexzielenski:
Now that schema.Schema has sync.Once attached, it makes no sense to encourage others to treat it as a value. This PR adds a `NewSchema` function as well as changes some usages of Schemas to use pointers rather than values.

### DIFF
--- a/schema/elements.go
+++ b/schema/elements.go
@@ -259,3 +259,25 @@ func (s *Schema) Resolve(tr TypeRef) (Atom, bool) {
 	}
 	return tr.Inlined, true
 }
+
+// Clones this instance of Schema into the other
+// If other is nil this method does nothing.
+// If other is already initialized, overwrites it with this instance
+// Warning: Not thread safe
+func (s Schema) CopyInto(dst *Schema) {
+	if dst == nil {
+		return
+	}
+
+	// Schema type is considered immutable so sharing references
+	dst.Types = s.Types
+
+	if s.m != nil {
+		// If cache is non-nil then the once token had been consumed.
+		// Must reset token and use it again to ensure same semantics.
+		dst.once = sync.Once{}
+		dst.once.Do(func() {
+			dst.m = s.m
+		})
+	}
+}
--- a/schema/elements_test.go
+++ b/schema/elements_test.go
@@ -120,3 +120,44 @@ func TestResolve(t *testing.T) {
 		})
 	}
 }
+
+func TestCopyInto(t *testing.T) {
+	existing := "existing"
+	notExisting := "not-existing"
+	a := Atom{List: &List{}}
+
+	tests := []struct {
+		testName       string
+		schemaTypeDefs []TypeDef
+		typeRef        TypeRef
+	}{
+		{"noNamedType", nil, TypeRef{Inlined: a}},
+		{"notExistingNamedType", nil, TypeRef{NamedType: &notExisting}},
+		{"existingNamedType", []TypeDef{{Name: existing, Atom: a}}, TypeRef{NamedType: &existing}},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.testName, func(t *testing.T) {
+			t.Parallel()
+			s := Schema{
+				Types: tt.schemaTypeDefs,
+			}
+
+			theCopy := Schema{}
+			s.CopyInto(&theCopy)
+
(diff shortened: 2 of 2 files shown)


schema/elements.go
```diff
@@ -259,3 +259,24 @@ func (s *Schema) Resolve(tr TypeRef) (Atom, bool) {
...
+// If other is already initialized, overwrites it with this instance
+func (s Schema) CopyInto(dst *Schema) {
+	if dst == nil {
+		return
+	}
+
+	// Schema type is considered immutable so sharing references
+	dst.Types = s.Types
+
+	if s.m != nil {
```
apelisse, line 274: Can you check `s.m` like this? What if the once is running on that type somewhere else?

alexzielenski: My intent was not to make this thread-safe with respect to `dest`. But you are right it should be thread-safe with respect to `s`. Will fix

alexzielenski: My intent was not to make this thread-safe with respect to `dest`. But you are right it should be thread-safe with respect to `s`. Will fix

alexzielenski: @apelisse  try again now

schema/elements.go
```diff
@@ -259,3 +263,26 @@ func (s *Schema) Resolve(tr TypeRef) (Atom, bool) {
 	}
 	return tr.Inlined, true
 }
+
+// Clones this isntance of Schema into the other
+// If other is nil this method does nothing.
+// If other is already initialized, overwrites it with this instance
+//
+// `dst` must not be shared across threads
+func (s Schema) CopyInto(dst *Schema) {
```
liggitt, line 272: is CopyInto really what we want, or is all of this a side-effect of having to dereference Schema in order to store a non-pointer in Parser? xref https://github.com/kubernetes/kubernetes/pull/109212#discussion_r840200723

apelisse:
Yeah the pointer is probably better but causes a massive change?

apelisse:
Sounds like a plan to me!

schema/elements.go
alexzielenski: PR is updated to change schema to use pointers rather than adding `CopyInto`

typed/parser.go
```diff
@@ -29,7 +29,7 @@ type YAMLObject string
 
 // Parser implements YAMLParser and allows introspecting the schema.
 type Parser struct {
-	Schema schema.Schema
+	Schema *schema.Schema
```
apelisse, line 32: Where do we create that parser?

alexzielenski: This library provides one constructor here which uses YAML unmarshalling to function, so it does not appear to require modification:

https://github.com/kubernetes-sigs/structured-merge-diff/blob/73412afd729b9cdaa8d4f87341b04e2877cdff97/typed/parser.go#L53


None of the other usages are in this library, so they will have to be updated separately when they update their dependencies. Here are a few I found:

https://github.com/kubernetes/apimachinery/blob/master/pkg/util/managedfields/gvkparser.go#L62
https://github.com/kubernetes/client-go/blob/28ccde769fc5519dd84e5512ebf303ac86ef9d7c/util/jsonpath/parser.go#L63

apelisse: Sounds good, that second link is unrelated though.

apelisse:
This is technically an API change so we should have to bump the major version :-(

alexzielenski:
> This is technically an API change so we should have to bump the major version :-(

After talking off-line about this we decided not to change the API (which would cause a major bump) since this change is too minor. Instead, @apelisse and I determined that a sufficient compromise is to allow `go vet` to find issues when `Schema` is copied, and to provide a `CopyInto` function.

schema/elements.go
```diff
@@ -259,3 +259,25 @@ func (s *Schema) Resolve(tr TypeRef) (Atom, bool) {
 	}
 	return tr.Inlined, true
 }
+
+// Clones this instance of Schema into the other
+// If other is nil this method does nothing.
+// If other is already initialized, overwrites it with this instance
+// Warning: Not thread safe
+func (s Schema) CopyInto(dst *Schema) {
```
apelisse, line 267: I know it's sort of trivial, but can we get a test?

apelisse:
I think that's good, thanks
/lgtm
/approve

PR #216, opened 2022-04-19
add CompareFields

alexzielenski:
/hold
/assign apelisse

adds ability to get a `Comparison` of an object while ignoring certain subpaths using a callback function

### DIFF
--- a/typed/merge.go
+++ b/typed/merge.go
@@ -34,6 +34,11 @@ type mergingWalker struct {
 	// How to merge. Called after schema validation for all leaf fields.
 	rule mergeRule
 
+	// If set, called before both non-leaf and leaf items are merged.
+	// Return `true` if the items should be merged
+	// and `false` if they should not merged or children be traversed
+	preItemHook traversalRule
+
 	// If set, called after non-leaf items have been merged. (`out` is
 	// probably already set.)
 	postItemHook mergeRule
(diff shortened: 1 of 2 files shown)


typed/typed.go
```diff
@@ -119,6 +119,10 @@ func (tv TypedValue) Merge(pso *TypedValue) (*TypedValue, error) {
 // match), or an error will be returned. Validation errors will be returned if
 // the objects don't conform to the schema.
 func (tv TypedValue) Compare(rhs *TypedValue) (c *Comparison, err error) {
+	return tv.CompareFields(rhs, nil)
+}
+
+func (tv TypedValue) CompareFields(rhs *TypedValue, callback func(prefix *fieldpath.Path) bool) (c *Comparison, err error) {
```
apelisse, line 125: I don't know if I'd make this an arbitrary callback rather than give it a set of fields?

apelisse:
We'll need tests and benchmarks before we can merge that, but go ahead and see if it solves our problem in k/k in the mean time.

typed/merge.go
```diff
@@ -55,6 +60,10 @@ type mergingWalker struct {
 // comparable type.
 type mergeRule func(w *mergingWalker)
 
+// Same as mergeRule but returns a boolean to decide whether the merge/traversal
+// will proceed
+type traversalRule func(w *mergingWalker) bool
```
lavalamp, line 65: Probably should call it a predicate, not a rule.

typed/merge.go
```diff
@@ -34,6 +34,11 @@ type mergingWalker struct {
 	// How to merge. Called after schema validation for all leaf fields.
 	rule mergeRule
 
+	// If set, called before both non-leaf and leaf items are merged.
+	// Return `true` if the items should be merged
+	// and `false` if they should not merged or children be traversed
+	preItemHook traversalRule
```
lavalamp, line 40: `shouldMerge`?

lavalamp: `proceedOrSkip`?

typed/typed.go
lavalamp: Yeah I am inclined to agree that passing in a set of fields to skip is probably best?

PR #219, opened 2022-06-10
Add test to show that typename and version don't always match

apelisse:
The test is flaky since the behavior depends on the order of which the
versions are processed, and the current code that prevents that from
happening depends on an invalid name for the version.

To make sure you get the failure:
```sh
$ go test -run=TestVersionDoesntMatchTypename -count=10 ./merge/
```

### DIFF
--- a/merge/multiple_appliers_test.go
+++ b/merge/multiple_appliers_test.go
@@ -1945,3 +1945,158 @@ func TestMultipleAppliersReliantFieldsConversions(t *testing.T) {
 		})
 	}
 }
+
+var versionDoesntMatchTypeName = func() Parser {
+	parser, err := typed.NewParser(`types:
+- name: TypeV1
+  map:
+    fields:
+      - name: field_foo_rely_on_bar
+        type:
+          scalar: string
+      - name: common_field
+        type:
+          scalar: string
+- name: TypeV2
+  map:
+    fields:
+      - name: required_field_bar
+        type:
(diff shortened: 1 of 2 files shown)


apelisse:
The bot failed, that's a good thing, it's supposed to :-)

apelisse:
Unfortunately I couldn't really make the test non-flaky, maybe I should run it in a loop internally so it fails more often ... suggestions?

lavalamp:
Yeah you could internally do 10 tries or something. Or you could fix the code 😅

apelisse:
> Yeah you could internally do 10 tries or something. Or you could fix the code 😅

I'll just do both ... :-)

apelisse:
I've run the test a 100 times and it didn't flake!

zqzten:
I've been busy this week, sorry😥. Thanks for fixing it for me, looks good!

zqzten (approved):

zqzten:
@apelisse Apologize for bothering, I'm wondering why did this PR get closed? Is the version fix no longer needed?

apelisse:
Wasn't this merged into another PR? I'll try to take a look.

zqzten:
Nope, at least the fix is not in the upstream code so far.

apelisse:
OK, this was a clean rebase on top of master, so none of this was indeed merge. Thanks!

PR #220, opened 2022-06-22
add field-level element relationship which overrides referred type

alexzielenski:
Part of our solution for https://github.com/kubernetes/kubernetes/pull/110495

Adds necessary functionality to support field-level annotations which may override the annotation of the referred type.

### DIFF
--- a/schema/elements.go
+++ b/schema/elements.go
@@ -30,11 +30,7 @@ type Schema struct {
 	once sync.Once
 	m    map[string]TypeDef
 
-	// Once used to protect the initialization of `lock` field.
-	lockOnce sync.Once
-	// Lock which protects writes to resolvedTypes. Used as pointer so that
-	// schema may be used as a value type
-	lock *sync.Mutex
+	lock sync.Mutex
 	// Cached results of resolving type references to atoms. Only stores
 	// type references which require fields of Atom to be overriden.
 	resolvedTypes map[TypeRef]Atom
@@ -43,7 +39,7 @@ type Schema struct {
 // A TypeSpecifier references a particular type in a schema.
 type TypeSpecifier struct {
 	Type   TypeRef `yaml:"type,omitempty"`
-	Schema Schema  `yaml:"schema,omitempty"`
+	Schema *Schema `yaml:"schema,omitempty"`
 }
 
 // TypeDef represents a named type in a schema.
@@ -290,13 +286,9 @@ func (s *Schema) Resolve(tr TypeRef) (Atom, bool) {
 	}
 
 	// Check to see if we have a cached version of this type
-	s.lockOnce.Do(func() {
-		s.lock = &sync.Mutex{}
-		s.resolvedTypes = make(map[TypeRef]Atom)
-	})
-
 	s.lock.Lock()
 	defer s.lock.Unlock()
+	s.resolvedTypes = make(map[TypeRef]Atom)
 
 	var result Atom
 	var exists bool
@@ -335,7 +327,7 @@ func (s *Schema) Resolve(tr TypeRef) (Atom, bool) {
 // If other is nil this method does nothing.
 // If other is already initialized, overwrites it with this instance
 // Warning: Not thread safe
-func (s Schema) CopyInto(dst *Schema) {
+func (s *Schema) CopyInto(dst *Schema) {
 	if dst == nil {
 		return
 	}
--- a/schema/elements.go
+++ b/schema/elements.go
@@ -16,7 +16,9 @@ limitations under the License.
 
 package schema
 
-import "sync"
+import (
+	"sync"
+)
 
 // Schema is a list of named types.
 //
@@ -27,6 +29,15 @@ type Schema struct {
 
 	once sync.Once
 	m    map[string]TypeDef
+
+	// Once used to protect the initialization of `lock` field.
+	lockOnce sync.Once
+	// Lock which protects writes to resolvedTypes. Used as pointer so that
+	// schema may be used as a value type
+	lock *sync.Mutex
+	// Cached results of resolving type references to atoms. Only stores
+	// type references which require fields of Atom to be overriden.
+	resolvedTypes map[TypeRef]Atom
 }
 
 // A TypeSpecifier references a particular type in a schema.
@@ -48,6 +59,12 @@ type TypeRef struct {
 	// Either the name or one member of Atom should be set.
 	NamedType *string `yaml:"namedType,omitempty"`
 	Inlined   Atom    `yaml:",inline,omitempty"`
+
+	// If this reference refers to a map-type or list-type, this field overrides
+	// the `ElementRelationship` of the referred type when resolved.
+	// If this field is nil, then it has no effect.
+	// See `Map` and `List` for more information about `ElementRelationship`
+	ElementRelationship *ElementRelationship `yaml:"elementRelationship,omitempty"`
 }
 
 // Atom represents the smallest possible pieces of the type system.
@@ -244,20 +261,74 @@ func (s *Schema) FindNamedType(name string) (TypeDef, bool) {
 	return t, ok
 }
 
+func (s *Schema) resolveNoOverrides(tr TypeRef) (Atom, bool) {
+	result := Atom{}
+
+	if tr.NamedType != nil {
+		t, ok := s.FindNamedType(*tr.NamedType)
+		if !ok {
+			return Atom{}, false
+		}
+
+		result = t.Atom
+	} else {
+		result = tr.Inlined
+	}
+
+	return result, true
+}
+
 // Resolve is a convenience function which returns the atom referenced, whether
 // it is inline or named. Returns (Atom{}, false) if the type can't be resolved.
 //
 // This allows callers to not care about the difference between a (possibly
 // inlined) reference and a definition.
 func (s *Schema) Resolve(tr TypeRef) (Atom, bool) {
-	if tr.NamedType != nil {
-		t, ok := s.FindNamedType(*tr.NamedType)
-		if !ok {
+	// If this is a plain reference with no overrides, just return the type
+	if tr.ElementRelationship == nil {
+		return s.resolveNoOverrides(tr)
+	}
+
+	// Check to see if we have a cached version of this type
+	s.lockOnce.Do(func() {
+		s.lock = &sync.Mutex{}
+		s.resolvedTypes = make(map[TypeRef]Atom)
+	})
+
+	s.lock.Lock()
+	defer s.lock.Unlock()
+
+	var result Atom
+	var exists bool
+
+	// Return cached result if available
+	// If not, calculate result and cache it
+	if result, exists = s.resolvedTypes[tr]; !exists {
+		if result, exists = s.resolveNoOverrides(tr); exists {
+			// Allow field-level electives to override the referred type's modifiers
+			switch {
+			case result.Map != nil:
+				mapCopy := *result.Map
+				mapCopy.ElementRelationship = *tr.ElementRelationship
+				result.Map = &mapCopy
+			case result.List != nil:
+				listCopy := *result.List
+				listCopy.ElementRelationship = *tr.ElementRelationship
+				result.List = &listCopy
+			case result.Scalar != nil:
+				return Atom{}, false
+			default:
+				return Atom{}, false
+			}
+		} else {
 			return Atom{}, false
 		}
-		return t.Atom, true
+
+		// Save result. If it is nil, that is also recorded as not existing.
+		s.resolvedTypes[tr] = result
 	}
-	return tr.Inlined, true
+
+	return result, true
 }
 
 // Clones this instance of Schema into the other
--- a/schema/elements.go
+++ b/schema/elements.go
@@ -101,11 +101,11 @@ const (
 
 // Map is a key-value pair. Its default semantics are the same as an
 // associative list, but:
-// * It is serialized differently:
+//   - It is serialized differently:
 //     map:  {"k": {"value": "v"}}
 //     list: [{"key": "k", "value": "v"}]
-// * Keys must be string typed.
-// * Keys can't have multiple components.
+//   - Keys must be string typed.
+//   - Keys can't have multiple components.
 //
 // Optionally, maps may be atomic (for example, imagine representing an RGB
 // color value--it doesn't make sense to have different actors own the R and G
--- a/typed/reconcile_schema.go
+++ b/typed/reconcile_schema.go
@@ -110,7 +110,7 @@ func (v *reconcileWithSchemaWalker) finishDescent(v2 *reconcileWithSchemaWalker)
 }
 
 // ReconcileFieldSetWithSchema reconciles the a field set with any changes to the
-//// object's schema since the field set was written. Returns the reconciled field set, or nil of
+// // object's schema since the field set was written. Returns the reconciled field set, or nil of
 // no changes were made to the field set.
 //
 // Supports:
--- a/typed/reconcile_schema.go
+++ b/typed/reconcile_schema.go
@@ -110,7 +110,7 @@ func (v *reconcileWithSchemaWalker) finishDescent(v2 *reconcileWithSchemaWalker)
 }
 
 // ReconcileFieldSetWithSchema reconciles the a field set with any changes to the
-// // object's schema since the field set was written. Returns the reconciled field set, or nil of
+// object's schema since the field set was written. Returns the reconciled field set, or nil of
 // no changes were made to the field set.
 //
 // Supports:
--- a/typed/typed.go
+++ b/typed/typed.go
@@ -99,12 +99,13 @@ func (tv TypedValue) ToFieldSet() (*fieldpath.Set, error) {
 
 // Merge returns the result of merging tv and pso ("partially specified
 // object") together. Of note:
-//  * No fields can be removed by this operation.
-//  * If both tv and pso specify a given leaf field, the result will keep pso's
-//    value.
-//  * Container typed elements will have their items ordered:
-//    * like tv, if pso doesn't change anything in the container
-//    * like pso, if pso does change something in the container.
+//   - No fields can be removed by this operation.
+//   - If both tv and pso specify a given leaf field, the result will keep pso's
+//     value.
+//   - Container typed elements will have their items ordered:
+//   - like tv, if pso doesn't change anything in the container
+//   - like pso, if pso does change something in the container.
+//
 // tv and pso must both be of the same type (their Schema and TypeRef must
 // match), or an error will be returned. Validation errors will be returned if
 // the objects don't conform to the schema.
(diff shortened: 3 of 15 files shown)


schema/elements.go
```diff
@@ -48,6 +50,10 @@ type TypeRef struct {
 	// Either the name or one member of Atom should be set.
 	NamedType *string `yaml:"namedType,omitempty"`
 	Inlined   Atom    `yaml:",inline,omitempty"`
+
+	// If `NamedType` is non-nil and refers to a map-type, this field overrides
```
apelisse, line 54: First, I don't think we should limite this to "NamedType".

Second of all, how much have you though of the "override" vs not-override? My knee-jerk reaction would also be that override is the way to go.

schema/elements.go
```diff
@@ -255,6 +261,26 @@ func (s *Schema) Resolve(tr TypeRef) (Atom, bool) {
 		if !ok {
 			return Atom{}, false
 		}
+
+		// Allow field-level electives to override the referred type
+		switch {
```
apelisse, line 266: Should we fail if the sub-type is not a map?

apelisse (commented):
We need tests for that, thanks

schema/elements.go
alexzielenski: I am leaning towards it being an override sort of behavior, rather than only applying when the initial is unset. I think that  to not override the original option when a field-level annotation is explicitly applied would be surprising.

schema/elements.go
alexzielenski: Yes and probably also should add that to schema validation

alexzielenski:
Added tests, addressed comments.

schema/elements.go
```diff
@@ -250,14 +257,49 @@ func (s *Schema) FindNamedType(name string) (TypeDef, bool) {
...
+
+		result.List = &listCopy
+
+	case result.Scalar != nil:
+		if tr.ElementRelationship != nil {
+			return Atom{}, false
+		}
+
+	default:
+		// Atom violates invariant that one of Map/List/Scalar is set
```
apelisse, line 298: This empty Atom return value is just terrible. I don't think we handle it properly anywhere and usually leads to hard to debug errors. Anyways, just ranting :-)

schema/elements.go
```diff
@@ -250,14 +257,49 @@ func (s *Schema) FindNamedType(name string) (TypeDef, bool) {
...
+		result.Map = &mapCopy
+	case result.List != nil:
+		listCopy := *result.List
+
+		if tr.ElementRelationship != nil {
+			listCopy.ElementRelationship = *tr.ElementRelationship
+		}
+
+		result.List = &listCopy
+
```
apelisse, line 291: empty line ;-)

apelisse (commented):
I think I'd like to see a more sophisticated test to show that this works correctly, take an example of a test in `merge/` where you can specific the schema and the operations.

schema/elements.go
alexzielenski: I think the contract of this function is to not inspect the first return value of the second is `false`.

schema/elements.go
```diff
@@ -48,6 +50,11 @@ type TypeRef struct {
 	// Either the name or one member of Atom should be set.
 	NamedType *string `yaml:"namedType,omitempty"`
 	Inlined   Atom    `yaml:",inline,omitempty"`
+
+	// If this reference refers to a map-type or list-type, this field overrides
```
lavalamp, line 54: only if the field is set, presumably?

schema/elements.go
```diff
@@ -250,14 +257,47 @@ func (s *Schema) FindNamedType(name string) (TypeDef, bool) {
...
+		result = t.Atom
+	} else {
+		result = tr.Inlined
 	}
-	return tr.Inlined, true
+
+	// Allow field-level electives to override the referred type's modifiers
+	switch {
+	case result.Map != nil:
+		mapCopy := *result.Map
```
lavalamp, line 276: Won't this make a lot of garbage? Can we at least do the if statement first and only do that if needed?

lavalamp (commented):
I guess you probably considered duplicating+renaming types as needed instead of this?

schema/elements.go
alexzielenski: yup. updated doc

schema/elements.go
alexzielenski: yes I was conflicted about this vs caching the resolved responses in a map as a field of `Schema` vs something else.

I have changed the code to have the allocations be conditional

lavalamp: Looks good now, can you run the benchmarks to see if this is too bad? (might have to edit something to make sure it's exercising this?)

alexzielenski: I wrote up a benchmark for this implementation, which showed to me a 100% cost in time and allocations when the `elementRelationship` override was used for ALL fields. This to me was too large a cost, so I refactored this PR to cache the result when `TypeRef` included overrides.

I thought it would be a reasonable conservative benchmark to randomly instrument 10% of ALL usages of `namedType` within fields with `elementRelationship: atomic`, and also show with 100% of ALL usages. Below is the results with the current state of the PR:

<details>
<summary>10% Fields Exercising Feature</summary>

```
name                                                  time/op
Operations/Pod/Create-20                              50.2µs ± 0%
Operations/Pod/Apply-20                                123µs ± 1%
Operations/Pod/ApplyTwice-20                           281µs ± 1%
Operations/Pod/Update-20                               118µs ± 0%
Operations/Pod/UpdateVersion-20                        166µs ± 1%
Operations/Node/Create-20                             71.7µs ± 1%
Operations/Node/Apply-20                               183µs ± 1%
Operations/Node/ApplyTwice-20                          434µs ± 1%
Operations/Node/Update-20                              185µs ± 1%
Operations/Node/UpdateVersion-20                       268µs ± 1%
Operations/Endpoints/Create-20                        5.90µs ± 1%
Operations/Endpoints/Apply-20                         12.5µs ± 0%
Operations/Endpoints/ApplyTwice-20                     649µs ± 0%
Operations/Endpoints/Update-20                         630µs ± 1%
Operations/Endpoints/UpdateVersion-20                 1.24ms ± 1%
Operations/CustomResourceDefinition/Create-20          598µs ± 0%
Operations/CustomResourceDefinition/Apply-20          1.41ms ± 0%
Operations/CustomResourceDefinition/ApplyTwice-20     3.00ms ± 0%
Operations/CustomResourceDefinition/Update-20         1.30ms ± 0%
Operations/CustomResourceDefinition/UpdateVersion-20  1.94ms ± 0%

name                                                  alloc/op
Operations/Pod/Create-20                              21.0kB ± 0%
Operations/Pod/Apply-20                               52.5kB ± 0%
Operations/Pod/ApplyTwice-20                           106kB ± 0%
Operations/Pod/Update-20                              39.4kB ± 0%
Operations/Pod/UpdateVersion-20                       51.7kB ± 0%
Operations/Node/Create-20                             30.0kB ± 0%
Operations/Node/Apply-20                              73.6kB ± 0%
Operations/Node/ApplyTwice-20                          152kB ± 0%
Operations/Node/Update-20                             60.3kB ± 0%
Operations/Node/UpdateVersion-20                      82.0kB ± 0%
Operations/Endpoints/Create-20                        3.55kB ± 0%
Operations/Endpoints/Apply-20                         6.41kB ± 0%
Operations/Endpoints/ApplyTwice-20                     126kB ± 0%
Operations/Endpoints/Update-20                         119kB ± 0%
Operations/Endpoints/UpdateVersion-20                  233kB ± 0%
Operations/CustomResourceDefinition/Create-20          170kB ± 0%
Operations/CustomResourceDefinition/Apply-20           483kB ± 0%
Operations/CustomResourceDefinition/ApplyTwice-20      962kB ± 0%
Operations/CustomResourceDefinition/Update-20          335kB ± 0%
Operations/CustomResourceDefinition/UpdateVersion-20   455kB ± 0%

name                                                  allocs/op
Operations/Pod/Create-20                                 473 ± 0%
Operations/Pod/Apply-20                                1.25k ± 0%
Operations/Pod/ApplyTwice-20                           2.61k ± 0%
Operations/Pod/Update-20                                 970 ± 0%
Operations/Pod/UpdateVersion-20                        1.40k ± 0%
Operations/Node/Create-20                                557 ± 0%
Operations/Node/Apply-20                               1.56k ± 0%
Operations/Node/ApplyTwice-20                          3.51k ± 0%
Operations/Node/Update-20                              1.41k ± 0%
Operations/Node/UpdateVersion-20                       2.16k ± 0%
Operations/Endpoints/Create-20                          76.0 ± 0%
Operations/Endpoints/Apply-20                            148 ± 0%
Operations/Endpoints/ApplyTwice-20                     6.33k ± 0%
Operations/Endpoints/Update-20                         6.17k ± 0%
Operations/Endpoints/UpdateVersion-20                  12.2k ± 0%
Operations/CustomResourceDefinition/Create-20          3.71k ± 0%
Operations/CustomResourceDefinition/Apply-20           10.3k ± 0%
Operations/CustomResourceDefinition/ApplyTwice-20      21.4k ± 0%
Operations/CustomResourceDefinition/Update-20          8.26k ± 0%
Operations/CustomResourceDefinition/UpdateVersion-20   12.7k ± 0%
```
</details>

<details>
<summary>100% of fields exercising feature</summary>

```
name                                                  time/op
Operations/Pod/Create-20                              50.3µs ± 0%
Operations/Pod/Apply-20                                123µs ± 1%
Operations/Pod/ApplyTwice-20                           281µs ± 1%
Operations/Pod/Update-20                               118µs ± 1%
Operations/Pod/UpdateVersion-20                        166µs ± 1%
Operations/Node/Create-20                             72.0µs ± 1%
Operations/Node/Apply-20                               184µs ± 1%
Operations/Node/ApplyTwice-20                          437µs ± 1%
Operations/Node/Update-20                              186µs ± 0%
Operations/Node/UpdateVersion-20                       268µs ± 0%
Operations/Endpoints/Create-20                        5.88µs ± 2%
Operations/Endpoints/Apply-20                         12.5µs ± 0%
Operations/Endpoints/ApplyTwice-20                     648µs ± 0%
Operations/Endpoints/Update-20                         628µs ± 1%
Operations/Endpoints/UpdateVersion-20                 1.24ms ± 1%
Operations/CustomResourceDefinition/Create-20          698µs ±14%
Operations/CustomResourceDefinition/Apply-20          1.62ms ±13%
Operations/CustomResourceDefinition/ApplyTwice-20     3.30ms ± 9%
Operations/CustomResourceDefinition/Update-20         1.43ms ± 9%
Operations/CustomResourceDefinition/UpdateVersion-20  2.05ms ± 5%

name                                                  alloc/op
Operations/Pod/Create-20                              21.0kB ± 0%
Operations/Pod/Apply-20                               52.5kB ± 0%
Operations/Pod/ApplyTwice-20                           106kB ± 0%
Operations/Pod/Update-20                              39.4kB ± 0%
Operations/Pod/UpdateVersion-20                       51.7kB ± 0%
Operations/Node/Create-20                             30.0kB ± 0%
Operations/Node/Apply-20                              73.6kB ± 0%
Operations/Node/ApplyTwice-20                          152kB ± 0%
Operations/Node/Update-20                             60.3kB ± 0%
Operations/Node/UpdateVersion-20                      82.0kB ± 0%
Operations/Endpoints/Create-20                        3.55kB ± 0%
Operations/Endpoints/Apply-20                         6.41kB ± 0%
Operations/Endpoints/ApplyTwice-20                     126kB ± 0%
Operations/Endpoints/Update-20                         119kB ± 0%
Operations/Endpoints/UpdateVersion-20                  233kB ± 0%
Operations/CustomResourceDefinition/Create-20          210kB ±19%
Operations/CustomResourceDefinition/Apply-20           574kB ±16%
Operations/CustomResourceDefinition/ApplyTwice-20     1.08MB ±11%
Operations/CustomResourceDefinition/Update-20          386kB ±13%
Operations/CustomResourceDefinition/UpdateVersion-20   509kB ±11%

name                                                  allocs/op
Operations/Pod/Create-20                                 473 ± 0%
Operations/Pod/Apply-20                                1.25k ± 0%
Operations/Pod/ApplyTwice-20                           2.61k ± 0%
Operations/Pod/Update-20                                 970 ± 0%
Operations/Pod/UpdateVersion-20                        1.40k ± 0%
Operations/Node/Create-20                                557 ± 0%
Operations/Node/Apply-20                               1.56k ± 0%
Operations/Node/ApplyTwice-20                          3.51k ± 0%
Operations/Node/Update-20                              1.41k ± 0%
Operations/Node/UpdateVersion-20                       2.16k ± 0%
Operations/Endpoints/Create-20                          76.0 ± 0%
Operations/Endpoints/Apply-20                            148 ± 0%
Operations/Endpoints/ApplyTwice-20                     6.33k ± 0%
Operations/Endpoints/Update-20                         6.17k ± 0%
Operations/Endpoints/UpdateVersion-20                  12.2k ± 0%
Operations/CustomResourceDefinition/Create-20          4.43k ±16%
Operations/CustomResourceDefinition/Apply-20           12.0k ±14%
Operations/CustomResourceDefinition/ApplyTwice-20      23.5k ± 9%
Operations/CustomResourceDefinition/Update-20          9.09k ± 9%
Operations/CustomResourceDefinition/UpdateVersion-20   13.6k ± 7%
```

The CRDs example seems especially impacted with 100% fields annotated because the `CustomResourceDefinition` type and its nested subtypes like (JSONSchemaProps) have a large amount of `namedType` usages which would recursively create extra instances in this example.
</details>

apelisse: You've instrumented the code in two different ways to run these two separate benchmark, can you add a separate test instead so that we can commit both versions?

schema/elements.go
```diff
@@ -27,6 +29,15 @@ type Schema struct {
 
 	once sync.Once
 	m    map[string]TypeDef
+
+	// Once used to protect the initialization of `lock` field.
+	lockOnce sync.Once
+	// Lock which protects writes to resolvedTypes. Used as pointer so that
+	// schema may be used as a value type
+	lock *sync.Mutex
```
lavalamp, line 37: Do you need to make this a pointer? What bad thing happens if you omit the Once? If the problem is that it's bad to copy Mutexes, isn't it also bad to copy Onces?

schema/elements.go
```diff
@@ -244,20 +261,74 @@ func (s *Schema) FindNamedType(name string) (TypeDef, bool) {
...
+			case result.Map != nil:
+				mapCopy := *result.Map
+				mapCopy.ElementRelationship = *tr.ElementRelationship
+				result.Map = &mapCopy
+			case result.List != nil:
+				listCopy := *result.List
+				listCopy.ElementRelationship = *tr.ElementRelationship
+				result.List = &listCopy
+			case result.Scalar != nil:
+				return Atom{}, false
```
lavalamp, line 319: hm, should it cache these too? Probably doesn't matter that much.

schema/elements.go
```diff
@@ -27,6 +29,15 @@ type Schema struct {
...
 	m    map[string]TypeDef
+
+	// Once used to protect the initialization of `lock` field.
+	lockOnce sync.Once
+	// Lock which protects writes to resolvedTypes. Used as pointer so that
+	// schema may be used as a value type
+	lock *sync.Mutex
+	// Cached results of resolving type references to atoms. Only stores
+	// type references which require fields of Atom to be overriden.
+	resolvedTypes map[TypeRef]Atom
```
lavalamp, line 40: Is `m` above ever invalidated? If so we should probably invalidate this at the same time.

schema/elements.go
alexzielenski: I trusted `go vet` too much it looks like here. Originally I had `sync.Mutex` as a value and `go vet` was throwing errors. 

I changed to this method since I saw `sync.Once` was already successfully used a few lines above in the original code, so I assumed its implementation would be safe. However, looking into the source for `Once`, I see it just stores a `Mutex` by value.

Now I am questioning whether the original code is safe. And wondering the correct approach: 

1. to use a `sync.Mutex` here and manually check if any copies are safe. Silencing go vet errors?
2. change usages of `Schema` to be as a pointer everywhere. From a brief review `Schema` is usually used as a pointer, but there are a few places where its usage as a value might be problematic:

https://github.com/kubernetes-sigs/structured-merge-diff/blob/daa3a199edcc48f138daf17ff9ec1ea875f563b8/typed/parser.go#L30-L33


(I am unsure of the consequences of changing the below to use pointers, it might be ok)
https://github.com/kubernetes-sigs/structured-merge-diff/blob/ded43451896507b64bce2978534fc6a534d3042e/schema/equals_test.go#L34-L39

https://github.com/kubernetes-sigs/structured-merge-diff/blob/f0a394c4104958b8b907bbe5a496cd55406f9d32/schema/elements.go#L39-L43

lavalamp: We should probably change all uses to be a pointer, yeah...

schema/elements.go
alexzielenski: I dont think so. The cost of not caching them is 1 extra cache lookup.

schema/elements.go
alexzielenski: `m` is not ever reset/recomputed

schema/elements.go
alexzielenski: latest changes refactor usages of `Schema` to use a pointer

schema/elements.go
alexzielenski: latest commits add these benchmarks to `real_test.go`

alexzielenski:
/retest

Had typo in last attempt.

typed/typed.go
```diff
@@ -99,12 +99,13 @@ func (tv TypedValue) ToFieldSet() (*fieldpath.Set, error) {
...
-//    value.
-//  * Container typed elements will have their items ordered:
-//    * like tv, if pso doesn't change anything in the container
-//    * like pso, if pso does change something in the container.
+//   - No fields can be removed by this operation.
+//   - If both tv and pso specify a given leaf field, the result will keep pso's
+//     value.
+//   - Container typed elements will have their items ordered:
+//   - like tv, if pso doesn't change anything in the container
+//   - like pso, if pso does change something in the container.
```
lavalamp, line 107: The extra indenting in the original was deliberate...

typed/reconcile_schema.go
```diff
@@ -110,7 +110,7 @@ func (v *reconcileWithSchemaWalker) finishDescent(v2 *reconcileWithSchemaWalker)
 }
 
 // ReconcileFieldSetWithSchema reconciles the a field set with any changes to the
-//// object's schema since the field set was written. Returns the reconciled field set, or nil of
+// // object's schema since the field set was written. Returns the reconciled field set, or nil of
```
lavalamp, line 113:
```suggestion
// object's schema since the field set was written. Returns the reconciled field set, or nil of
```

typed/typed.go
alexzielenski: unfortunately gofmt won't allow this anymore so I was forced to change it to an ordered list.

apelisse:
I tagged the merged commit with [`v4.2.2`](https://github.com/kubernetes-sigs/structured-merge-diff/tree/v4.2.2).

apelisse:
I suspect the schema change to pointer is backwards incompatible, I'll have to fix the tag and change the release branch to v5.

apelisse:
I removed the v4.2.2 tag since I think it was not correct semver.

PR #221, opened 2022-07-25
Improve operations parsing

apelisse:
Improve the way we invoke the parser in the Operations tests, so that we can later test against different schemas, and also have more than one test for the same type.

cc @alexzielenski


alexzielenski (approved):
code looks cleaner, and necessary for my pending PR

/lgtm

PR #222, opened 2022-07-25
Fix exponential merge recursion

alexzielenski:
fixes an exponential recursion bug reproduced by the introduced benchmark 

occurs when attempting to merge `__untyped_deduced_` map with a nil value `lhs` and non-nil `rhs`. `deduceAtom` will just return the same input, so merge will create a new branch for recursion. Every nested level of fields ends up exponentially creating more merge branches. This PR adds logic so that the extra nil subtree is not traversed needlessly, and only the other operand's type is used for traversal.

### DIFF
--- a/internal/testdata/apiresourceimport-cr.yaml
+++ b/internal/testdata/apiresourceimport-cr.yaml
@@ -1,0 +1,5992 @@
+apiVersion: apiresource.kcp.dev/v1alpha1
+kind: APIResourceImport
+metadata:
+  annotations:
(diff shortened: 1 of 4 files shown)


apelisse:
Looks good to me, that's really awesome. We can't really show the performance improvement since the benchmark doesn't finish in the main branch.

apelisse:
Thanks,
/lgtm
@lavalamp, do you want to have a look? I'll approve if needed.

lavalamp:
/lgtm
/approve

Thanks!

PR #223, opened 2022-08-01
Create package v5 for backward incompatible change

apelisse:
The latest PR has changed the schema to a pointer which is a backward incompatible change so we need to bump the version fo the package.


apelisse:
cc @alexzielenski

apelisse:
OK, tagged with [`v5.0.0`](https://github.com/kubernetes-sigs/structured-merge-diff/tree/v5.0.0)

PR #224, opened 2022-08-01
Revert backwards incompatible changes

apelisse:
Can we live with that? Everything still passes, but at least we don't have to bump the major, so any backport/cherry-pick will be much easier to sell.

### DIFF
--- a/schema/elements.go
+++ b/schema/elements.go
@@ -30,7 +30,11 @@ type Schema struct {
 	once sync.Once
 	m    map[string]TypeDef
 
-	lock sync.Mutex
+	// Once used to protect the initialization of `lock` field.
+	lockOnce sync.Once
+	// Lock which protects writes to resolvedTypes. Used as pointer so that
+	// schema may be used as a value type
+	lock *sync.Mutex
 	// Cached results of resolving type references to atoms. Only stores
 	// type references which require fields of Atom to be overriden.
 	resolvedTypes map[TypeRef]Atom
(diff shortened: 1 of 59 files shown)


schema/elements.go
```diff
@@ -30,7 +30,11 @@ type Schema struct {
 	once sync.Once
 	m    map[string]TypeDef
 
-	lock sync.Mutex
+	// Once used to protect the initialization of `lock` field.
+	lockOnce sync.Once
+	// Lock which protects writes to resolvedTypes. Used as pointer so that
+	// schema may be used as a value type
+	lock *sync.Mutex
```
apelisse, line 37: Do we need all these locks here? Can we make the lock by value since we're not supposed to copy a schema?

alexzielenski: yeah there were no unsafe schema usages in the ptr semantics refactor

alexzielenski: can remove the once. just need the lock

alexzielenski:
/hold

tentative lgtm

apelisse:
OK I've tried to fix the tags, done this:
- Created [`v4.2.3`](https://github.com/kubernetes-sigs/structured-merge-diff/tree/v4.2.3) with this merge commit
- Removed `release-5.0` branch and created [`release-5.0-wrong`](https://github.com/kubernetes-sigs/structured-merge-diff/tree/v5.0.0-wrong) instead
- Created [`v4.2.2-wrong`](https://github.com/kubernetes-sigs/structured-merge-diff/tree/v4.2.2-wrong) to clarify why we skipped that version

PR #233, opened 2023-03-06
Minor cleanups

apelisse:
A couple of minor fixes, including one to fix `go vet` which broke CI.


schema/schemaschema.go
```diff
@@ -110,7 +110,7 @@ var SchemaSchemaYAML = `types:
         scalar: string
     - name: deduceInvalidDiscriminator
       type:
-        scalar: bool
```
lavalamp, line 113: can you make a test adjustment that would catch this?

PR #238, opened 2023-06-23
fixture: Show diff when managers don't match

apelisse:
Very minor nit. We have the diff when the managers don't match, but we don't print it which makes debugging harder than it has to be. Let's print it  to help.


alexzielenski (approved):
/lgtm
/approve

PR #241, opened 2023-06-28
fieldpath: Remove unused fromvalue

apelisse:


apelisse:
This is an API breaking change. I don't think it's used anywhere but we might want to bump major version when we do that.

PR #242, opened 2023-07-07
deep copy slice object during append

kmala:
Since p is slice, its value is changed during the tree traversal(iterate). Hence use deep copy instead.

fixes https://github.com/kubernetes/kubernetes/issues/119141

### DIFF
--- a/merge/conflict_test.go
+++ b/merge/conflict_test.go
@@ -17,6 +17,7 @@ limitations under the License.
 package merge_test
 
 import (
+	"reflect"
 	"testing"
 
 	"sigs.k8s.io/structured-merge-diff/v4/fieldpath"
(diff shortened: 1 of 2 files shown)


merge/conflict_test.go
```diff
@@ -92,3 +93,44 @@ func TestToSet(t *testing.T) {
...
+
+func TestConflictsFromManagers(t *testing.T) {
+	type args struct {
+		sets fieldpath.ManagedFields
+	}
+	tests := []struct {
+		name string
+		args args
+		want string
+	}{
```
apelisse, line 105: Since you have only on test I think it's fine to remove the table-test. Also this repo generally doesn't use that format for table-tests.

merge/conflict_test.go
```diff
@@ -92,3 +93,44 @@ func TestToSet(t *testing.T) {
...
+		{
+			name: "test with common prefix",
+			args: args{
+				sets: fieldpath.ManagedFields{
+					"Bob": fieldpath.NewVersionedSet(
+						_NS(
+							_P("spec", "template", "spec", "containers", _KBF("name", "probe"), "livenessProbe", "exec", "command"),
+							_P("spec", "template", "spec", "containers", _KBF("name", "probe"), "livenessProbe", "periodSeconds"),
+							_P("spec", "template", "spec", "containers", _KBF("name", "probe"), "readinessProbe", "exec", "command"),
+							_P("spec", "template", "spec", "containers", _KBF("name", "probe"), "readinessProbe", "periodSeconds"),
```
apelisse, line 115: Can we make these shorter? The simplest test to reproduce is usually best.

merge/conflict_test.go
```diff
@@ -92,3 +93,44 @@ func TestToSet(t *testing.T) {
...
+			want: `conflicts with "Bob":
+- .spec.template.spec.containers[name="probe"].livenessProbe.periodSeconds
+- .spec.template.spec.containers[name="probe"].livenessProbe.exec.command
+- .spec.template.spec.containers[name="probe"].readinessProbe.periodSeconds
+- .spec.template.spec.containers[name="probe"].readinessProbe.exec.command`,
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			if got := merge.ConflictsFromManagers(tt.args.sets); !reflect.DeepEqual(got.Error(), tt.want) {
```
apelisse, line 131: You don't need reflect.DeepEqual to compare strings.

apelisse (commented):
Thanks, that looks good. A few comments for simplifying the test!

kmala:
> A few comments for simplifying the test
@apelisse  i had simplified the test, can you please review it again.

merge/conflict_test.go
```diff
@@ -92,3 +92,23 @@ func TestToSet(t *testing.T) {
 		t.Fatalf("expected\n%v\n, but got\n%v\n", expected, actual)
 	}
 }
+
+func TestConflictsFromManagers(t *testing.T) {
+
```
apelisse, line 97:
```suggestion
```

merge/conflict_test.go
```diff
@@ -92,3 +92,23 @@ func TestToSet(t *testing.T) {
...
+	got := merge.ConflictsFromManagers(fieldpath.ManagedFields{
+		"Bob": fieldpath.NewVersionedSet(
+			_NS(
+				_P("obj", "template", "obj", "list", _KBF("name", "a"), "id"),
+				_P("obj", "template", "obj", "list", _KBF("name", "a"), "key"),
```
apelisse, line 102: Are these the simplest form that trigger the bug?!

kmala: yes, this is the simplest which triggers the bug.

merge/conflict_test.go
kmala: addressed

Tusenka:
Could you explain why this PR will fix the issue. What is the cause of the issue?

kmala:
> Could you explain why this PR will fix the issue. What is the cause of the issue?

This will fix the issue because on apply [this function](https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apimachinery/pkg/util/managedfields/internal/structuredmerge.go#L152) gets called and it returns the wrong result. The cause as i mentioned in the description is because of doing shallow copy slice during iterate.

merge/conflict_test.go
kmala: @apelisse can you please review again when you get a chance

apelisse (commented):
/lgtm
/approve

PR #244, opened 2023-07-12
merge: Add tests for merging with lists that have duplicates

apelisse:
Tried to capture some of the discussions we've had in the form of unit-tests to make sure we properly capture the behavior we want. Let me know if there are use-cases that I'm missing.

I have basically found three cases:
1. Merging with an object that has a list with duplicates, but the change is unrelated to the list
2. Merging with an object that has a list with duplicates, the change modifies the list, but not a duplicated item
3. Merging with an object that has a list with duplicates, the change modifies a duplicated item

I think unarguable the first two should apply the change without de-dupping, since that's very clear that's what people would expect (their intent is not about the items they haven't specified). The last one is a little more controversial, but since i don't think we're going to allow them to apply duplicates, I think removing all the duplicates with the applied value is the "saner" option. Happy to have a discussion about it.

cc @alexzielenski @thockin

### DIFF
--- a/merge/duplicates_test.go
+++ b/merge/duplicates_test.go
@@ -1,0 +1,707 @@
+/*
+Copyright 2019 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package merge_test
+
+import (
+	"testing"
+
+	"sigs.k8s.io/structured-merge-diff/v4/fieldpath"
+	. "sigs.k8s.io/structured-merge-diff/v4/internal/fixture"
+	"sigs.k8s.io/structured-merge-diff/v4/merge"
+	"sigs.k8s.io/structured-merge-diff/v4/typed"
+)
+
+var duplicatesParser = func() Parser {
+	parser, err := typed.NewParser(`types:
+- name: type
+  map:
+    fields:
+      - name: list
+        type:
+          namedType: associativeList
+      - name: unrelated
+        type:
+          scalar: numeric
+      - name: set
+        type:
+          namedType: set
+- name: associativeList
+  list:
+    elementType:
+      namedType: myElement
+    elementRelationship: associative
+    keys:
+    - name
+- name: myElement
+  map:
+    fields:
+    - name: name
+      type:
+        scalar: string
+    - name: value1
+      type:
+        scalar: numeric
+    - name: value2
+      type:
+        scalar: numeric
+- name: set
+  list:
+    elementType:
+      scalar: numeric
+    elementRelationship: set
+`)
+	if err != nil {
+		panic(err)
+	}
+	return SameVersionParser{T: parser.Type("type")}
+}()
+
+func TestDuplicates(t *testing.T) {
+	tests := map[string]TestCase{
+		"sets/ownership/duplicates": {
+			Ops: []Operation{
+				Update{
+					Manager: "updater-one",
+					Object: `
+						set: [1, 1, 3, 4]
+					`,
+					APIVersion: "v1",
+				},
+			},
+			Managed: fieldpath.ManagedFields{
+				"updater-one": fieldpath.NewVersionedSet(
+					_NS(
+						_P("set", _V(1)),
+						_P("set", _V(3)),
+						_P("set", _V(4)),
+					),
+					"v1",
+					false,
+				),
+				"updater-two": fieldpath.NewVersionedSet(
+					_NS(
+						_P("set", _V(1)),
+					),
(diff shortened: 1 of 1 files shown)


jpbetz:
> The last one is a little more controversial, but since i don't think we're going to allow them to apply duplicates, I think removing all the duplicates with the applied value is the "saner" option. Happy to have a discussion about it.

The last one can be further broken down into these cases:

1. The duplicates are owned by a single field manager
  a. the apply uses the same field manager
  b. the apply uses a different field manager
2.  The duplicates are owned by a different field managers
  a. the apply uses one of the field managers
  b. the apply uses a different field manager (I'll treat this the same as 1.b)

(1.a) clearly works with the proposed approach. I don't see that as being at all contentious.

For (1.b), I don't see any better alternative than the proposed solution. This is less clean cut because it would be possible to append instead of replace, but when I try to reason though what operations are possible, append seems less flexible.

(2.a) is the one that I'm least confident about with because the applier already owns one of the duplicates but not the other, so clobbering both makes me more uncomfortable. In the case of `envs` it would probably be safer to only replace the owned fields, but that wouldn't be congruent with the replacement approach taken for (1.b)..

apelisse:
OK, I guess I missed some context in here, but:
> The last one can be further broken down into these cases:

There's only ever going to be 1 owner for all the duplicated fields (since the managedFields are sets, we can't really have more than one of them). Co-ownership can happen when someone applies the existing value, but I don't know about that.
Also note that I'm pretty convince we should NOT allow people to apply duplicated fields, so we can't really have the horizontal dimension 1.a/2.a  (the applier can't be one of them because since one can't apply duplicates, one of the duplicate can't be own by an applier). So it's hard to reason about your cases.

One thing that is fundamentally related though, and I'll update the tests for that is:
What if what I apply is already set as one of the multiple fields. e.g. current is `[{key: a, value: 1},{key: a, value: 2}]`, I apply `[{key: a, value: 1}]`, should we keep `[{key: a, value: 2}]`? The corollary is: if the object is `[{key: a, value: 2}]`, and I *update* to `[{key: a, value: 1},{key: a, value: 2}]`, should we keep the previous owners?

> For (1.b), I don't see any better alternative than the proposed solution. This is less clean cut because it would be possible to append instead of replace, but when I try to reason though what operations are possible, append seems less flexible.

"Append" is the current behavior so I think we have to keep this. if we wanted to replace, then yeah, we should make this atomic, but again, I don't think that's the right answer here.

apelisse:
I'm adding cases for what we should have started with ... who owns what after "simple" changes 😅

apelisse:
Added these tests because basically, we'll have to make two changes:
1. The "Compare" function, that decides what has changed, and thus who owns what and what conflicts.
2. The "Merge" function, that decides what the final outcome of applying looks like.

We could probably treat these two discussions as separate to make things simpler.

jpbetz:
> There's only ever going to be 1 owner for all the duplicated fields

This simplifies things in my mind. Since we don't have multiple owners of the fields, there are less cases to consider.  I don't know if this makes the outcome better, but it means we really don't have that many options on how to approach this. I'm onboard with replacing duplicates when merging for all cases given this.

apelisse:
Cool, I think I can make the code work to have these tests pass.

thockin (commented):
If I follow these properly, I am OK with it.  Capturing some thoughts from conversation today, specifically focused on `env.

In the face of dups, SSA today is broken, so clearly nobody is using it.

So, what does CSA do?

If there are dups in the current object and I apply with only new keys, the dups are preserved and the new key is inserted (not necessarily at the end).  This implies that anyone who is doing apply (client-side, since SSA fails) to co-own `env` while dups are present *can't* be relying on ordering (since CSA doesn't guarantee it) and are therefore almost certainly NOT adding values which xref other values (because that would need stable ordering).

If there are dups in the current object and I apply with that same key, my new value will replace the first instance of the key, and the other dups are preserved.  This is a little surprising, but if it were a problem SOMEONE would have complained.  Therefore I assume that this is not a real problem in practice.

If I apply with dups in my patch, they are silently de-duped.  This is CSA's usually MO, so not surprising.

Reminder: dups are allowed via create and update (non-patch).

I don't generally accept "that's dumb, nobody would do that" as an excuse to potentially break people.  In this case, it's either already weirdly broken in CSA or totally disallowed in SSA.  

So the gold-standard for making it not a hard-fail in SSA is not "as good as create" but "not worse than CSA".  !!

IIUC this PR proposes that:

1) Dups in the patch are an error.  They would be allowed (but warned) in create/update and disallowed by apply.  CSA silently corrupts the input by picking one, so I'll say this is net better than CSA.  Good.

2) Dups in unmodified fields or in a modified field but with different keys in the patch will be preserved.  This is the same as CSA.  Good.

3) Dups in a modified fields with the dup key in the patch will be de-duped and replaced with the new value.  This is different from CSA, but arguably better - more deterministic and understandable.  This is (IMO) the only arguable aspect of the design and given that it has never come up in 9 years (TTBOMK), not something that is actually a problem.

So, I am cautiously optimisitic on this.  I spent all my time thinking about `env` and not at all about the other fields.  I think there are further changes needed to make things like `ports` work (e.g. expanding the key to include hostPort and hostIP).

apelisse:
OK I think I have a pretty exhaustive list of what I think the behavior should be. Some details are still partly unknown, I guess I'll reason about this as I write the code

apelisse:
This one can be closed now, we've added the test.

Issue #234, opened 2023-06-05
Lenient validation and merging for non-strict `listType=map`


apelisse:
Right now, we have some lists that are marked as `listType=map` that are not strictly maps in Kubernetes. In that case, everything goes wrong in smd because the list is invalid (and can't be handled), so it's fully rejected. We can't even track the fields, which causes us to fail and return an error which is ignored by Kubernetes, and the managed fields are entirely reset.

Unfortunately, it means that objects that have such a list can't be server-side applied at all. We think it'd be simpler to merely ignore these (fairly rare) occurences and try to do the best thing possibe rather than fail.

This means:
1. When we validate an object with such a list, we should probably not fail altogether (though we should probably reject a request to server-side apply an object like this).
1. When tracking the fields, we should give duplicate fields to the same owner.
1. When server-side applying without the field, we need to be careful about how the field should be removed.
1. When we merge a list that has such a field, we need to be careful, especially if the server-side apply contains the field (should we remove both other occurences? Probably).

apelisse:
cc @thockin @jpbetz

jpbetz:
Do we have the list of fields that fall into this "marked as listType=map but not validated as having unique keys" category?

apelisse:
Not really but, that's a great question:
1. I'm pretty sure  CRDs don't allow that and never will, so there's no chance people start abusing that mechanism there (IOW, actual validation matches SSA validation)
2. Since it's only in built-ins, and because a lot of these were not designed for that initially (unfortunately), we have:
environment variables and a few containerports.

I considered adding a new, separate tag, but that would require a KEP and make things impossible to backport. I think we need to make it clear that this is a work-around and shouldn't be used for any new list (again, only for built-ins so we have a lot more control over that).

thockin:
There are two distinct failure modes, the one we have usually run into is client-side apply vs. `Service.spec.ports`, where the keys are looser than validation, and SSA is the SOLUTION (the keys match validation).

The more recent one is where SSA's keys are more strict than validation.  So far we've found:
* `Pod.spec.containers[].ports`
* `Pod.spec.containers[].env`
* `Pod.spec.containers[].envFrom`

I admit that I haven't scoured the API for this case, and I don't honestly know how I would in any reasonable amount of time. :(  IOW, there's ikely to be more

jpbetz:
I did a scan. Here's what I found:

[...]

PR #247, opened 2023-09-26
validation: Allow loose validation for sets

apelisse:
If a set (or associative lists keys) has duplicated elements, then this will not fail the validation. Everything else is still the same.

/assign @alexzielenski

### DIFF
--- a/typed/typed.go
+++ b/typed/typed.go
@@ -24,16 +24,24 @@ import (
 	"sigs.k8s.io/structured-merge-diff/v4/value"
 )
 
+// ValidationOptions is the list of all the options available when running the validation.
+type ValidationOptions int
+
+const (
+	// AllowDuplicates means that sets and associative lists can have duplicate similar items.
+	AllowDuplicates ValidationOptions = iota
+)
+
 // AsTyped accepts a value and a type and returns a TypedValue. 'v' must have
 // type 'typeName' in the schema. An error is returned if the v doesn't conform
 // to the schema.
-func AsTyped(v value.Value, s *schema.Schema, typeRef schema.TypeRef) (*TypedValue, error) {
+func AsTyped(v value.Value, s *schema.Schema, typeRef schema.TypeRef, opts ...ValidationOptions) (*TypedValue, error) {
 	tv := &TypedValue{
 		value:   v,
 		typeRef: typeRef,
 		schema:  s,
 	}
-	if err := tv.Validate(); err != nil {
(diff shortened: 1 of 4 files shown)


apelisse:
This is starting to address #234. I have all the code passing the numerous tests I've written, but I'm going to send small PRs.
cc @thockin

typed/typed.go
```diff
@@ -24,16 +24,22 @@ import (
 	"sigs.k8s.io/structured-merge-diff/v4/value"
 )
 
+type ValidationOptions int
```
apelisse, line 27: Note to self, I should document that.

typed/typed.go
```diff
@@ -24,16 +24,22 @@ import (
...
+
+const (
+	Loose ValidationOptions = iota
```
apelisse, line 30: Also I'm actually renaming this `AllowDuplicates` since .. that's a lot closer to what it actually is.

alexzielenski (commented):
easy to follow

/lgtm

typed/typed.go
```diff
@@ -24,16 +24,24 @@ import (
...
+const (
+	// AllowDuplicates means that sets and associative lists can have duplicate similar items.
+	AllowDuplicates ValidationOptions = iota
+)
+
 // AsTyped accepts a value and a type and returns a TypedValue. 'v' must have
 // type 'typeName' in the schema. An error is returned if the v doesn't conform
 // to the schema.
-func AsTyped(v value.Value, s *schema.Schema, typeRef schema.TypeRef) (*TypedValue, error) {
+func AsTyped(v value.Value, s *schema.Schema, typeRef schema.TypeRef, opts ...ValidationOptions) (*TypedValue, error) {
```
apelisse, line 38: I've decided to use that convention since we want to backport this code and I didn't want to break the compatibility with existing code (forcing me to bump the major version). Alternative could be to make a `AsTypedWithDuplicates`

apelisse:
No worries, let me know what you want me to address, I'll focus on that before continuing.

typed/typed.go
```diff
@@ -24,16 +24,24 @@ import (
 	"sigs.k8s.io/structured-merge-diff/v4/value"
 )
 
+// ValidationOptions is the list of all the options available when running the validation.
+type ValidationOptions int
```
alexzielenski, line 28: Woudl be better if this was not an int and instead something like a `func (v *validatingObjectWalker)` or something similar so that we aren't bound to very simple options in the future

PR #248, opened 2023-09-26
Validate duplicates with a new methods rather than Options

apelisse:
cc @alexzielenski 
Do you think this is better?


alexzielenski (approved):
sure
/lgtm
/hold

in case you wanted to do something else

apelisse:
OK thanks, let's not do it.

PR #249, opened 2023-09-27
tofieldset: Add tests to show it already allow duplicates

apelisse:
Continuation of #247.

This uses the pattern that I've generally used to solve the problem, which is to not try to be too smart about duplicate entries, and just record the path element for the whole thing, rather than try to track all the fields within. It's not super relevant because no one should ever be able to apply duplicates, so they wouldn't really have conflicts with sub-fields anyway, we've decided that they should have a conflict with the idea of "duplicates" (de-duplicating is a conflict).

/assign @alexzielenski

### DIFF
--- a/typed/tofieldset.go
+++ b/typed/tofieldset.go
@@ -94,9 +94,31 @@ func (v *toFieldSetWalker) doScalar(t *schema.Scalar) ValidationErrors {
 }
 
 func (v *toFieldSetWalker) visitListItems(t *schema.List, list value.List) (errs ValidationErrors) {
+	// Keeps track of the PEs we've seen
+	seen := fieldpath.MakePathElementSet(list.Length())
+	// Keeps tracks of the PEs we've counted as duplicates
+	duplicates := fieldpath.MakePathElementSet(list.Length())
 	for i := 0; i < list.Length(); i++ {
 		child := list.At(i)
 		pe, _ := listItemToPathElement(v.allocator, v.schema, t, child)
+		if seen.Has(pe) {
+			if duplicates.Has(pe) {
+				// do nothing
+			} else {
+				v.set.Insert(append(v.path, pe))
+				duplicates.Insert(pe)
+			}
+		} else {
+			seen.Insert(pe)
+		}
+	}
+
+	for i := 0; i < list.Length(); i++ {
+		child := list.At(i)
+		pe, _ := listItemToPathElement(v.allocator, v.schema, t, child)
+		if duplicates.Has(pe) {
+			continue
+		}
 		v2 := v.prepareDescent(pe, t.ElementType)
 		v2.value = child
 		errs = append(errs, v2.toFieldSet()...)
--- a/typed/toset_test.go
+++ b/typed/toset_test.go
@@ -155,7 +155,12 @@ var fieldsetCases = []fieldsetTestCase{{
 			_P("setStr", _V("b")),
 			_P("setStr", _V("c")),
 		)},
-		{`{"setBool":[true,false]}`, _NS(
+		{`{"setStr":["a","b","c","a","b","c","c"]}`, _NS(
+			_P("setStr", _V("a")),
+			_P("setStr", _V("b")),
+			_P("setStr", _V("c")),
+		)},
+		{`{"setBool":[true,false,true]}`, _NS(
 			_P("setBool", _V(true)),
 			_P("setBool", _V(false)),
 		)},
@@ -244,6 +249,16 @@ var fieldsetCases = []fieldsetTestCase{{
 			_P("list", _KBF("key", "b", "id", 1), "key"),
 			_P("list", _KBF("key", "b", "id", 1), "id"),
 		)},
+		{`{"list":[{"key":"a","id":1,"nv":2},{"key":"a","id":2,"nv":3},{"key":"b","id":1},{"key":"a","id":2,"bv":true}]}`, _NS(
+			_P("list", _KBF("key", "a", "id", 1)),
+			_P("list", _KBF("key", "a", "id", 1), "key"),
+			_P("list", _KBF("key", "a", "id", 1), "id"),
+			_P("list", _KBF("key", "a", "id", 1), "nv"),
+			_P("list", _KBF("key", "a", "id", 2)),
+			_P("list", _KBF("key", "b", "id", 1)),
+			_P("list", _KBF("key", "b", "id", 1), "key"),
+			_P("list", _KBF("key", "b", "id", 1), "id"),
+		)},
 		{`{"atomicList":["a","a","a"]}`, _NS(_P("atomicList"))},
 	},
 }}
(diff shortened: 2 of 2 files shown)


typed/toset_test.go
```diff
@@ -244,6 +249,16 @@ var fieldsetCases = []fieldsetTestCase{{
 			_P("list", _KBF("key", "b", "id", 1), "key"),
 			_P("list", _KBF("key", "b", "id", 1), "id"),
 		)},
+		{`{"list":[{"key":"a","id":1,"nv":2},{"key":"a","id":2,"nv":3},{"key":"b","id":1},{"key":"a","id":2,"bv":true}]}`, _NS(
+			_P("list", _KBF("key", "a", "id", 1)),
+			_P("list", _KBF("key", "a", "id", 1), "key"),
+			_P("list", _KBF("key", "a", "id", 1), "id"),
+			_P("list", _KBF("key", "a", "id", 1), "nv"),
+			_P("list", _KBF("key", "a", "id", 2)),
```
apelisse, line 257: cc @jpbetz 
This code is mostly irrelevant because we only use `ToFieldSet` on objects that are applied, and we should never allow apply to have duplicates (should we change that code?). I still wanted to illustrate that when someone will update with duplicates items, we will just give them ownership of the associative key and that's it, as a way to indicate that they own that key multiple times. WDYT?

typed/tofieldset.go
```diff
@@ -94,9 +94,31 @@ func (v *toFieldSetWalker) doScalar(t *schema.Scalar) ValidationErrors {
 }
 
 func (v *toFieldSetWalker) visitListItems(t *schema.List, list value.List) (errs ValidationErrors) {
+	// Keeps track of the PEs we've seen
+	seen := fieldpath.MakePathElementSet(list.Length())
+	// Keeps tracks of the PEs we've counted as duplicates
+	duplicates := fieldpath.MakePathElementSet(list.Length())
```
alexzielenski, line 100: I would've expected the implementation to use just one set and one loop; `if seen.Has(pe) { continue }`

apelisse: we're trying to add duplicates exactly once to the set. I need to know if I've already seen them to know if they are duplicates, and I need to know if I've already added them to the set.

apelisse:
/hold

I'm trying a new test with pruning fields that are removed by appliers and this doesn't work, looking into this.

typed/tofieldset.go
alexzielenski: I guess Im proposing the following pseudocode: 

```go
seen := NewSet()
for item in list {
    if seen.Has(item) { continue }
    seen.Insert(item)
    listWithoutDuplicates.append(item)
}
```

Doesn't this do as you described?

apelisse: Yeah, I think that makes sense, thanks.

typed/toset_test.go
apelisse: One drawback of this, but I don't know if we can solve it anyway, is that if someone applies one item, and then it gets updated into two items, that item won't be pruned when it's removed from the applied configuration, meaning it can no longer be removed with server-side apply. I think that's reasonable.

apelisse:
/hold cancel

Thought about this further, I think it's fine, unless I forgot the use-case I had in mind ...

typed/toset_test.go
alexzielenski: > This code is mostly irrelevant because we only use ToFieldSet on objects that are applied, and we should never allow apply to have duplicates (now, that being said, I'm curious if this whole change even makes sense?)

If apply can never have duplicates why are we making SMD tolerate duplicates?

alexzielenski, line 257: > if someone applies one item, and then it gets updated into two items, that item won't be pruned when it's removed from the applied configuration, meaning it can no longer be removed with server-side apply.

Why wouldn't it be pruned?

typed/tofieldset.go
alexzielenski: can mark this as resolved, since after discussion offline it is not what you want

alexzielenski (commented):
Just to clarify, this change makes SMD give ownership of the entire object if there are duplicates in a list map?

I'm having a hard time tracking the "before" behavior and the "after" behavior.

typed/toset_test.go
apelisse: > If apply can never have duplicates why are we making SMD ToFieldSet tolerate duplicates?

Context, we're trying to solve #234

For built-ins, the `listType=map` is used a little loosely, and we don't have a great way to address that, except for accepting some duplicates. We still can't apply duplicates, but right now you can't apply an objects that has these duplicates.

> How do Updates get the list of fields to own if not via ToFieldSet?

By using "Compare", they own the fields that they changed, since we don't know exactly what they intended to own.

apelisse: Because when you update to multiple items, it steals the ownership of the whole item and we don't prune things that you don't have ownership over.

alexzielenski: > We still can't apply duplicates, but right now you can't apply an objects that has these duplicates.

Is ToFieldSet invoked on the final merged object, or the partial patch? If it is invoked on the partial patch aren't there no duplicates and this PR has no effect?

apelisse: Correct, that's more or less what I said in the PR body:
> It's not super relevant because no one should ever be able to apply duplicates

But:
1. This is for consistency, if we agree that this is how ownership of things that are duplicates, we should probably apply it everywhere, even if it's not relevant here.
2. This is the first time we use that new pattern in the code so I want to make sure we're fine with it.

alexzielenski: That seems like correct behavior to me. What if it was a map typed list with duplicate keys but unequal objects?

Apply, "applier": [ObjWithKeyA]
Update, "updater": [ObjWithKeyAButSlighlyDifferent, ObjWithKeyAButAlsoDifferent] # Should be three copies of 'A'?
Apply, "applier": [] # is A pruned?

apelisse: Very good point.

So first of all, `Update` replaces the whole list, so if you do
> Update, "updater": [ObjWithKeyAButSlighlyDifferent, ObjWithKeyAButAlsoDifferent] # Should be three copies of 'A'?

you end-up with that exact object.

Second, the way this is meant to work, is that if you have duplicate, you own the key and the key only, so here's what happens in your scenario:

Apply, "applier": [ObjWithKeyA] -> applier owns ObjWithKeyA
Update, "updater": [ObjWithKeyAButSlighlyDifferent, ObjWithKeyAButAlsoDifferent] -> duplicates, updater owns just "KeyA" (none of the fields)
Apply, "applier": [] # is A pruned? no
Whats the final object
[ObjWithKeyAButSlighlyDifferent, ObjWithKeyAButAlsoDifferent]  with KeyA owned by updater.

alexzielenski: i think thats reasonable

alexzielenski:
LGTM from me. I think you mentioned you wanted @jpbetz input so will defer tag

alexzielenski:
whoops

PR #250, opened 2023-10-06
fieldpath: Allow map to be updated

apelisse:
Minor change that will become important and was a little misleading to me. If you add something to the PathElementValueMap twice, the second value will be dropped (as opposed to the first).

/assign @alexzielenski


alexzielenski (approved):
/lgtm

PR #251, opened 2023-10-09
typed: Update compare algorithm to handle duplicates

apelisse:
Same strategy as #249, still trying to address #234. Algorithm goes as follow:
1. Duplicates are treated as separate from non-duplicates items
2. Adding a duplicate to an existing item means we will remove the existing (counted as removed) and added as a duplicate (added).
3. Removing does the opposite
4. If any of the duplicated value changes, key will be counted as "modified".

Rewording:

In compare, treat duplicate keys in associative lists/sets as a separate atomic list, marked by owning just the duplicated key.

/assign @alexzielenski
/cc @jpbetz

### DIFF
--- a/typed/compare.go
+++ b/typed/compare.go
@@ -213,7 +213,16 @@ func (w *compareWalker) visitListItems(t *schema.List, lhs, rhs value.List) (err
 		lLen = lhs.Length()
 	}
 
-	lValues := fieldpath.MakePathElementValueMap(lLen)
+	maxLength := rLen
+	if lLen > maxLength {
+		maxLength = lLen
+	}
+	// Contains all the unique PEs between lhs and rhs, exactly once.
+	// Order doesn't matter since we're just tracking ownership in a set.
+	allPEs := make([]fieldpath.PathElement, 0, maxLength)
+
+	// Gather all the elements from lhs, indexed by PE, in a list for duplicates.
+	lValues := fieldpath.MakePathElementMap(lLen)
 	for i := 0; i < lLen; i++ {
 		child := lhs.At(i)
 		pe, err := listItemToPathElement(w.allocator, w.schema, t, child)
@@ -224,14 +233,18 @@ func (w *compareWalker) visitListItems(t *schema.List, lhs, rhs value.List) (err
 			// this element.
 			continue
 		}
-		// Ignore repeated occurences of `pe`.
-		if _, found := lValues.Get(pe); found {
-			continue
+
+		if v, found := lValues.Get(pe); found {
+			list := v.([]value.Value)
+			lValues.Insert(pe, append(list, child))
+		} else {
+			lValues.Insert(pe, []value.Value{child})
+			allPEs = append(allPEs, pe)
 		}
-		lValues.Insert(pe, child)
 	}
 
-	rValues := fieldpath.MakePathElementValueMap(rLen)
+	// Gather all the elements from rhs, indexed by PE, in a list for duplicates.
+	rValues := fieldpath.MakePathElementMap(rLen)
 	for i := 0; i < rLen; i++ {
 		rValue := rhs.At(i)
 		pe, err := listItemToPathElement(w.allocator, w.schema, t, rValue)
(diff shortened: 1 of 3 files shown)


typed/compare.go
```diff
@@ -224,11 +232,14 @@ func (w *compareWalker) visitListItems(t *schema.List, lhs, rhs value.List) (err
 			// this element.
 			continue
 		}
-		// Ignore repeated occurences of `pe`.
-		if _, found := lValues.Get(pe); found {
-			continue
+
+		if v, found := lValues.Get(pe); found {
+			list := v.Unstructured().([]value.Value)
```
alexzielenski, line 237: Would it be preferable to avoid unstructured conversion for the comparison? This kinda makes me sad

typed/compare.go
```diff
@@ -242,31 +253,67 @@ func (w *compareWalker) visitListItems(t *schema.List, lhs, rhs value.List) (err
...
+				for i := range lList {
+					if !value.Equals(lList[i], rList[i]) {
+						return false
+					}
+				}
+				return true
+			}
+			if !listEqual(lList, rList) {
+				w.comparison.Modified.Insert(append(w.path, pe))
+			}
```
alexzielenski, line 305: If I'm reading this correctly, this is the "no duplicates" case and should be semantically identical to the old code?

typed/compare.go
```diff
@@ -242,31 +253,67 @@ func (w *compareWalker) visitListItems(t *schema.List, lhs, rhs value.List) (err
...
+		case len(lList) >= 2:
+			if len(rList) != 0 {
+				errs = append(errs, w.compareListItem(t, pe, nil, rList[0])...)
```
alexzielenski, line 308: Why `nil` for old value? Maybe I'm not following these two "duplicates" cases very well

typed/compare.go
```diff
@@ -242,31 +253,67 @@ func (w *compareWalker) visitListItems(t *schema.List, lhs, rhs value.List) (err
...
+			}
+			w.comparison.Removed.Insert(append(w.path, pe))
```
alexzielenski, line 310: Would love some comments explaining why these cases are different

alexzielenski (commented):
I'm still not super familiar with this part of SMD yet so I asked a lot of questions, sorry :)

typed/compare.go
```diff
@@ -242,31 +253,67 @@ func (w *compareWalker) visitListItems(t *schema.List, lhs, rhs value.List) (err
...
+		case len(lList) <= 1 && len(rList) <= 1:
+			lValue := value.Value(nil)
+			if len(lList) != 0 {
+				lValue = lList[0]
+			}
+			rValue := value.Value(nil)
+			if len(rList) != 0 {
+				rValue = rList[0]
+			}
+			errs = append(errs, w.compareListItem(t, pe, lValue, rValue)...)
```
alexzielenski, line 290: If I'm reading this correctly, this is the "no duplicates" case and should be semantically identical to the old code?

apelisse:
OK, I've added comments to explain what happens, hopefully that helps.

typed/compare.go
alexzielenski: Yeah I dont understand how this line works. Shouldn't `Unstructured` return a []interface{}?

alexzielenski (commented):
lgtm except for that comment about unstructured

typed/compare.go
alexzielenski: after Comments and offline discussion it makes snese to me. resolved

typed/compare.go
alexzielenski: after Comments and offline discussion it makes snese to me. resolved

typed/compare.go
apelisse: Yeah, it's because I'm abusing the PathElementValueMap here. I don't insert a valid value in there, I cast a `[]value.Value` into a value and then retrieve it, and it just works. I don't know if we want to introduce generics code in here just yet, so I can't really make the map take whatever type of parameter type I want.

apelisse:
OK I've added a new commit to create a new map of PEs to `interface{}`, and then updated the other commit to use it instead of value, which simplifies the code a tiny bit since we don't have to create the new valueinterface thingy.

alexzielenski:
/lgtm
/hold

in case you wanted to change anything else

apelisse:
/hold

I'm good thanks!

apelisse:
I meant
/hold cancel

PR #252, opened 2023-10-10
add `Set` to value interface for lists

alexzielenski:
mirrors behavior of existing Set for maps

needed to write a `Default` function that applies schema defaults to a `value.Value` in Declarative Validation

/assign @apelisse

### DIFF
--- a/value/list.go
+++ b/value/list.go
@@ -30,6 +30,7 @@ type List interface {
 	AtUsing(Allocator, int) Value
 
 	// Set sets the value at the given index to the provided value
+	// It panics if the index is negative or too large to allocate the memory.
 	Set(int, Value)
 
 	// Range returns a ListRange for iterating over the items in the list.
--- a/value/list.go
+++ b/value/list.go
@@ -28,6 +28,10 @@ type List interface {
 	// The returned Value should be given back to the Allocator when no longer needed
 	// by calling Allocator.Free(Value).
 	AtUsing(Allocator, int) Value
+
+	// Set sets the value at the given index to the provided value
+	Set(int, Value)
+
 	// Range returns a ListRange for iterating over the items in the list.
 	Range() ListRange
 	// RangeUsing uses the provided allocator and returns a ListRange for
--- a/value/listreflect.go
+++ b/value/listreflect.go
@@ -74,6 +74,17 @@ func (r listReflect) EqualsUsing(a Allocator, other List) bool {
 	return ListEqualsUsing(a, &r, other)
 }
 
+func toReflectValue(v Value) reflect.Value {
+	if vr, ok := v.(*valueReflect); ok {
+		return vr.Value
+	}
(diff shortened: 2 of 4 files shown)


value/listreflect.go
```diff
@@ -74,6 +74,19 @@ func (r listReflect) EqualsUsing(a Allocator, other List) bool {
 	return ListEqualsUsing(a, &r, other)
 }
 
+func toReflectValue(v Value) reflect.Value {
+	if vr, ok := v.(*valueReflect); ok {
+		return vr.Value
```
alexzielenski, line 79: unsafe

value/listreflect.go
```diff
@@ -74,6 +74,19 @@ func (r listReflect) EqualsUsing(a Allocator, other List) bool {
...
+	}
+	return reflect.ValueOf(v.Unstructured())
```
alexzielenski, line 81: I am using toUnstructured because the complementary function for maps does the same, but it makes me feel bad.

I feel like we should be using `FromUnstructured` but unfortunately that is in upstream k8s... @apelisse  tips?

apelisse:
I think that's fine, thanks
/lgtm
/approve

apelisse:
You need to fix a gofmt thingy.

value/valuereflect_test.go
```diff
@@ -661,6 +661,33 @@ func TestReflectList(t *testing.T) {
 	}
 }
 
+func TestReflectListMutateSet(t *testing.T) {
```
apelisse, line 664: Can we have a test `Set`-ting something that is not a scalar? :-)

alexzielenski: added a test with two map types. one set using reflect which works. another is set using a value wrapping an unstructured, which doesnt

value/list.go
```diff
@@ -28,6 +28,10 @@ type List interface {
 	// The returned Value should be given back to the Allocator when no longer needed
 	// by calling Allocator.Free(Value).
 	AtUsing(Allocator, int) Value
+
+	// Set sets the value at the given index to the provided value
```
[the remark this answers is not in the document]
alexzielenski: added godoc about out of range values

value/listreflect.go
alexzielenski: Wasn't referring to the performance cost. But setting any random reflect field to `Unstructured` sounds wrong...

apelisse: Yeah, I don't remember much but we discussed that with @jpbetz I think a long time ago, there's no way to make that work. If we receive an unstructured, we don't really know what types we're supposed to create, it's not made for that. We could get it through reflection but it's a lot more work. SMD only creates and modifies unstructured objects so that's easy but updating typed object wasn't part of the design.

apelisse:
/close

I think we abandoned the idea.

PR #253, opened 2023-10-13
typed: Allow duplicates when we merge

apelisse:
The goal is to ignore duplicates if they are not included in the partial object, and to replace all of the occurences if they are in the partial object.

/assign @liggitt 
/cc @alexzielenski

### DIFF
--- a/typed/merge.go
+++ b/typed/merge.go
@@ -211,7 +211,7 @@ func (w *mergingWalker) visitListItems(t *schema.List, lhs, rhs value.List) (err
 			if pe.Equals(rhsPEs[rI]) {
 				// merge LHS & RHS items
 				mergedRHS.Insert(pe, struct{}{})
-				lChild, _ := observedLHS.Get(pe)
+				lChild, _ := observedLHS.Get(pe) // may be nil if the PE is duplicaated.
 				rChild, _ := observedRHS.Get(pe)
 				mergeOut, errs := w.mergeListItem(t, pe, lChild, rChild)
 				errs = append(errs, errs...)
@@ -237,7 +237,7 @@ func (w *mergingWalker) visitListItems(t *schema.List, lhs, rhs value.List) (err
 		if lI < lLen {
 			pe := lhsPEs[lI]
 			if _, ok := observedRHS.Get(pe); !ok {
-				// take LHS item
+				// take LHS item using At to make sure we get the right item (observed may not contain the right item).
 				lChild := lhs.AtUsing(w.allocator, lI)
 				mergeOut, errs := w.mergeListItem(t, pe, lChild, nil)
 				errs = append(errs, errs...)
@@ -255,7 +255,7 @@ func (w *mergingWalker) visitListItems(t *schema.List, lhs, rhs value.List) (err
 			// Take the RHS item, merge with matching LHS item if possible
 			pe := rhsPEs[rI]
 			mergedRHS.Insert(pe, struct{}{})
-			lChild, _ := observedLHS.Get(pe) // may be nil
+			lChild, _ := observedLHS.Get(pe) // may be nil if absent or duplicaated.
 			rChild, _ := observedRHS.Get(pe)
 			mergeOut, errs := w.mergeListItem(t, pe, lChild, rChild)
 			errs = append(errs, errs...)
@@ -305,6 +305,9 @@ func (w *mergingWalker) indexListPathElements(t *schema.List, list value.List, a
 			continue
 		} else if !found {
 			observed.Insert(pe, child)
+		} else {
+			// Duplicated items are not merged with the new value, make them nil.
+			observed.Insert(pe, value.NewValueInterface(nil))
 		}
 		pes = append(pes, pe)
 	}
--- a/typed/merge.go
+++ b/typed/merge.go
@@ -180,10 +180,14 @@ func (w *mergingWalker) visitListItems(t *schema.List, lhs, rhs value.List) (err
 	}
 	out := make([]interface{}, 0, outLen)
 
-	rhsPEs, observedRHS, rhsErrs := w.indexListPathElements(t, rhs)
+	rhsPEs, observedRHS, rhsErrs := w.indexListPathElements(t, rhs, false)
 	errs = append(errs, rhsErrs...)
-	lhsPEs, observedLHS, lhsErrs := w.indexListPathElements(t, lhs)
+	lhsPEs, observedLHS, lhsErrs := w.indexListPathElements(t, lhs, true)
 	errs = append(errs, lhsErrs...)
+
+	if len(errs) != 0 {
+		return errs
+	}
 
 	sharedOrder := make([]*fieldpath.PathElement, 0, rLen)
 	for i := range rhsPEs {
@@ -199,12 +203,14 @@ func (w *mergingWalker) visitListItems(t *schema.List, lhs, rhs value.List) (err
 		sharedOrder = sharedOrder[1:]
 	}
 
+	mergedRHS := fieldpath.MakePathElementMap(len(rhsPEs))
 	lLen, rLen = len(lhsPEs), len(rhsPEs)
 	for lI, rI := 0, 0; lI < lLen || rI < rLen; {
 		if lI < lLen && rI < rLen {
 			pe := lhsPEs[lI]
 			if pe.Equals(rhsPEs[rI]) {
 				// merge LHS & RHS items
+				mergedRHS.Insert(pe, struct{}{})
 				lChild, _ := observedLHS.Get(pe)
 				rChild, _ := observedRHS.Get(pe)
 				mergeOut, errs := w.mergeListItem(t, pe, lChild, rChild)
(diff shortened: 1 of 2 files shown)


typed/merge_test.go
```diff
@@ -423,6 +467,18 @@ var mergeCases = []mergeTestCase{{
 		`{"atomicList":["a","a","a"]}`,
 		`{"atomicList":["a","a"]}`,
 		`{"atomicList":["a","a"]}`,
+	}, {
+		`{"list":[{"key":"a","id":1,"bv":true},{"key":"b","id":2},{"key":"a","id":1,"bv":false,"nv":2}]}`,
+		`{"list":[{"key":"a","id":1,"nv":3}, {"key":"c","id":3},{"key":"b","id":2}]}`,
+		`{"list":[{"key":"a","id":1,"bv":true,"nv":3},{"key":"c","id":3},{"key":"b","id":2}]}`,
```
apelisse, line 473: So this is merging the new item with the first occurence, I'm currently thinking we should replace altogether.

typed/merge.go
```diff
@@ -222,29 +228,33 @@ func (w *mergingWalker) visitListItems(t *schema.List, lhs, rhs value.List) (err
...
 				continue
 			}
 		}
 		if lI < lLen {
-			pe := lhsOrder[lI]
+			pe := lhsPEs[lI]
 			if _, ok := observedRHS.Get(pe); !ok {
 				// take LHS item
-				lChild, _ := observedLHS.Get(pe)
+				lChild := lhs.AtUsing(w.allocator, lI)
```
alexzielenski, line 241: This seems equivalent, since we are looking LHS first. Just curious why the change? Is it a noop?

typed/merge.go
```diff
@@ -290,11 +300,9 @@ func (w *mergingWalker) indexListPathElements(t *schema.List, list value.List) (
 			// this element.
 			continue
 		}
-		if _, found := observed.Get(pe); found {
-			errs = append(errs, errorf("duplicate entries for key %v", pe.String())...)
-			continue
+		if _, found := observed.Get(pe); !found {
+			observed.Insert(pe, child)
```
alexzielenski, line 304: Is this existence check necessary? Is `observed` a set?

typed/merge.go
```diff
@@ -222,29 +228,33 @@ func (w *mergingWalker) visitListItems(t *schema.List, lhs, rhs value.List) (err
...
+				lChild := lhs.AtUsing(w.allocator, lI)
 				mergeOut, errs := w.mergeListItem(t, pe, lChild, nil)
 				errs = append(errs, errs...)
 				if mergeOut != nil {
 					out = append(out, *mergeOut)
 				}
 				lI++
 				continue
+			} else if _, ok := mergedRHS.Get(pe); ok {
+				// we've already merged it with RHS, we don't want to duplicate it, skip it.
```
alexzielenski, line 250: I'm not sure why this case is necessary. It's hard to track the invariants of `observedRHS` vs `mergedRHS` without going very deep into the code.

typed/merge.go
```diff
@@ -180,14 +180,18 @@ func (w *mergingWalker) visitListItems(t *schema.List, lhs, rhs value.List) (err
 	}
 	out := make([]interface{}, 0, outLen)
 
-	rhsOrder, observedRHS, rhsErrs := w.indexListPathElements(t, rhs)
+	rhsPEs, observedRHS, rhsErrs := w.indexListPathElements(t, rhs)
```
alexzielenski, line 183: Are the PE's no longer ordered?

typed/merge_test.go
alexzielenski: This case surprised me. From our prior conversations I would not have expected the merge with one of the duplicates. It doesn't seem like as the author of such a resource I'd be very conscious of which of the duplicates I am merging with.


Is it possible to have two instances of the same value but they are owned by different field managers? In that case I almost would expect "MY" version of the duplicate to be merged; but don't think that'd be worth the effort.

alexzielenski (commented):
Except for one tests make a lot of sense and very straightforward.

TBH I could not gain deep sense of the code changes without obtaining degree in SMD. This code could benefit a lot from some more inline comments about the invariants of this algorithm.

typed/merge.go
apelisse: They are still, but "order" didn't really tell me what they were. I could revert or make it `rhsOrderedPEs` or something.

typed/merge.go
apelisse: Actually, lhs can have multiple times the same PE, so we want to make sure we get the right one (I don't think it's equivalent).

typed/merge.go
apelisse: observed is a map from pathelement to the actual value.

the value that ends-up being in the PE is the one we're going to use to merge I think, so if we were to override here, I think we would end-up merging with the last item in the list rather than the first. But since we've decided we might not want to do that, I will probably deal with that differently.

typed/merge_test.go
apelisse: > Is it possible to have two instances of the same value but they are owned by different field managers?

No.

> This case surprised me. From our prior conversations I would not have expected the merge with one of the duplicates. It doesn't seem like as the author of such a resource I'd be very conscious of which of the duplicates I am merging with.

yeah, I think we just need to replace rather than merge, which means I'll need to know if they are duplicated and replace altogether, I'll update the algorithm.

typed/merge.go
```diff
@@ -180,14 +180,18 @@ func (w *mergingWalker) visitListItems(t *schema.List, lhs, rhs value.List) (err
 	}
```
liggitt, line 180: AI: look for issue reported where someone couldn't update because of managed fields

typed/merge.go
```diff
@@ -199,12 +203,14 @@ func (w *mergingWalker) visitListItems(t *schema.List, lhs, rhs value.List) (err
...
+	lLen, rLen = len(lhsPEs), len(rhsPEs)
 	for lI, rI := 0, 0; lI < lLen || rI < rLen; {
 		if lI < lLen && rI < rLen {
-			pe := lhsOrder[lI]
-			if pe.Equals(rhsOrder[rI]) {
+			pe := lhsPEs[lI]
+			if pe.Equals(rhsPEs[rI]) {
 				// merge LHS & RHS items
+				mergedRHS.Insert(pe, struct{}{})
 				lChild, _ := observedLHS.Get(pe)
```
liggitt, line 214: comment that `lChild` will be nil if the LHS had duplicates for pe

typed/merge.go
```diff
@@ -222,29 +228,33 @@ func (w *mergingWalker) visitListItems(t *schema.List, lhs, rhs value.List) (err
...
 				// shared item, but not the one we want in this round
 				lI++
 				continue
 			}
 		}
 		if lI < lLen {
-			pe := lhsOrder[lI]
+			pe := lhsPEs[lI]
 			if _, ok := observedRHS.Get(pe); !ok {
 				// take LHS item
```
liggitt, line 240: comment that we use lhs.AtUsing instead of observedLHS because observedLHS can contain nil if the lhs had duplicates

typed/merge.go
```diff
@@ -222,29 +228,33 @@ func (w *mergingWalker) visitListItems(t *schema.List, lhs, rhs value.List) (err
...
+				// we've already merged it with RHS, we don't want to duplicate it, skip it.
+				lI++
 			}
 		}
 		if rI < rLen {
 			// Take the RHS item, merge with matching LHS item if possible
-			pe := rhsOrder[rI]
+			pe := rhsPEs[rI]
+			mergedRHS.Insert(pe, struct{}{})
 			lChild, _ := observedLHS.Get(pe) // may be nil
```
liggitt, line 258:
```suggestion
			lChild, _ := observedLHS.Get(pe) // may be nil if not present or LHS duplicated this item
```

typed/merge.go
```diff
@@ -291,10 +301,10 @@ func (w *mergingWalker) indexListPathElements(t *schema.List, list value.List) (
 			continue
 		}
 		if _, found := observed.Get(pe); found {
-			errs = append(errs, errorf("duplicate entries for key %v", pe.String())...)
-			continue
+			observed.Insert(pe, value.NewValueInterface(nil))
```
liggitt, line 304: is it possible to *only* tolerate duplicates making the LHS and keep returning the duplicate error making the RHS?

liggitt (commented):
tests look good, the change was shockingly small.

mostly requests for clearer comments, one request to keep the duplicate-error-returning bit the same when constructing observedRHS

typed/merge.go
apelisse: Added a comment for that.

apelisse:
PTAL, I think I've addressed all the commits. Also split the rename of rhsOrder and lhsOrder in a seperate commit to keep the meat of the change isolated.

liggitt:
Thanks, open the k/k PR to update master and ping me for review.

I suspect we'll want to backport this, not sure if that means picking the fix to other smd branches or not.

PR #255, opened 2023-10-24
Remove unions logic, keep schema components

apelisse:
This removes the generally unused union code from the library in a
non-backward incompatible way since the API was not used anyways. The
schema isn't changed since we don't want to break existing schema (the
unions fields will continue to be just ignored).

cc @alexzielenski @liggitt


apelisse:
I tested this code along with #254 and the tests are now working.

liggitt:
> We keep the whole API surface intact to stay within the same major release of structured-merge-diff even though the union code will now panic if used in any place

is this to satisfy an APIDiff-type check, or do we really want to let someone using this code compile and then panic them at runtime?

apelisse:
> I can't find any uses of these symbols in kubernetes/kubernetes or on github at all (using grep.app). Would it be better to remove them?

Removed them. It's obviously better to remove them but I had a hard time convincing myself I could remove these fuctions without bumping the major version which I'm not ready to do (and makes backporting the change more difficult).

liggitt:
/lgtm
/approve

update the title/description to match where this ended up

PR #259, opened 2024-05-23
Specifying `{}` to remove nested fields should yield `{}`, not `nil`

alexzielenski:
Consider old object:

```yaml
{
  requiredField: {
      optionalField1: 1
      optionalFIeld2: 2
   }
}
```

And patch, all with a single owner
```yaml
{
    requiredField: {}
}
```

I expect this to yield:

```yaml
{
    requiredField: {}
}
```

But today it yields:

```yaml
{
    requiredField: nil
}
```

Which results in an error on k8s since `nil` is not an acceptable value for this `requiredField`

### DIFF
--- a/typed/remove.go
+++ b/typed/remove.go
@@ -19,6 +19,8 @@ import (
 	"sigs.k8s.io/structured-merge-diff/v4/value"
 )
 
+var REMOVEKEEPEMPTYCOLLECTIONS = false
+
 type removingWalker struct {
 	value         value.Value
 	out           interface{}
@@ -58,6 +60,9 @@ func (w *removingWalker) doList(t *schema.List) (errs ValidationErrors) {
 	defer w.allocator.Free(l)
 	// If list is null or empty just return
 	if l == nil || l.Length() == 0 {
+		if REMOVEKEEPEMPTYCOLLECTIONS {
+			w.out = w.value.Unstructured()
+		}
 		return nil
 	}
 
@@ -66,6 +71,10 @@ func (w *removingWalker) doList(t *schema.List) (errs ValidationErrors) {
 	if t.ElementRelationship == schema.Atomic {
 		if w.shouldExtract {
 			w.out = w.value.Unstructured()
+		} else if !w.toRemove.Has(fieldpath.Path{}) && REMOVEKEEPEMPTYCOLLECTIONS {
+			// If the atomic list itself wasn't being removed, then it
+			// is being set to empty list (we only get here if a prefix of this fieldPath is in toRemove)
+			w.out = []interface{}{}
 		}
 		return nil
 	}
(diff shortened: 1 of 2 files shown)


typed/remove.go
```diff
@@ -58,6 +60,9 @@ func (w *removingWalker) doList(t *schema.List) (errs ValidationErrors) {
 	defer w.allocator.Free(l)
 	// If list is null or empty just return
 	if l == nil || l.Length() == 0 {
+		if REMOVEKEEPEMPTYCOLLECTIONS {
```
apelisse, line 63: Should we treat separately empty and nil here? Is that part of the problem?

typed/update_test.go
```diff
@@ -0,0 +1,141 @@
+package typed_test
+
+import (
+	"testing"
+
+	"sigs.k8s.io/structured-merge-diff/v4/internal/fixture"
```
apelisse, line 6: Aren't the other tests usually importing everything from fixture?

typed/update_test.go
```diff
@@ -0,0 +1,141 @@
...
+					Manager:    "default",
+					APIVersion: "v1",
+					Object: `
+                        nestedMap: {}
+                    `,
+				},
+			},
+			APIVersion: `v1`,
+			Object:     `{nestedMap: {}}`,
+		},
```
apelisse, line 121: Yeah, so I guess now I'm confused about:
1. What happens if we apply null instead.
2. What happens if you have further level of nesting (i.e. you're apply `{}` when you had `{'nested': {'numeric':1}}` (I'm guessing the same but would love to see)
3. How do you entirely remove the field? If null keeps null and {} keep it to empty, then how do you remove it altogether?

apelisse:
One last thing, I suspect lists and `maps` should behave the same. Struct, I'm not entirely sure. maybe they should?

typed/update_test.go
alexzielenski: Will add cases for 1 & 2

Re: 3. If you wanted to make `{nestedMap: { myField: value }}` into `{}` you would just apply `{}`. If you applied `{}` then `nestedMap` field is included in the `toRemove` set. If you had applied `{nestedMap: {}}` then only `nestedMap.myField` is in the `toRemove` set, so `nestedMap` should be retained

alexzielenski: For the null case naively I would expect it to keep null if you specified null.

alexzielenski:
> One last thing, I suspect lists and maps should behave the same. Struct, I'm not entirely sure. maybe they should?

This bug also affects structs. There are valid CRDs that can be constructed with defined object fields where `{}` is valid value users should be able to set. Today they cannot

typed/remove.go
alexzielenski: This change wasn't directly part of the problem I was trying to solve but I think something like this will be needed to have a consistent solution. The behavior of this right now always results in a `nil` list even if you specify `[]`

ybettan:
Hello.

Is this fix planned to be merged soon? We currently have to bump our API cause we need to add the `+nullable` kubebuilder annotation to a bunch of API object and adding a new mutating webhook is just breaking any pointers/slices in the API objects.

Ref: https://github.com/kubernetes-sigs/kernel-module-management/pull/826

PR #260, opened 2024-06-05
Error on conversion to unstructured for invalid json.Marshalers.

benluddy:
If a type's implementation of json.Marshaler returns bytes representing a valid JSON object or array followed by anything other than trailing whitespace, return an error rather than ignoring the trailing data. The documentation for the Marshaler interface indicates that implementations shouldn't do this, but it is safer to check (as json.Marshal does) than to rely on it.

### DIFF
--- a/value/reflectcache.go
+++ b/value/reflectcache.go
@@ -19,7 +19,9 @@ package value
 import (
 	"bytes"
 	"encoding/json"
+	"errors"
 	"fmt"
+	"io"
 	"reflect"
 	"sort"
 	"sync"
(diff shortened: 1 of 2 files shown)


value/reflectcache_test.go
```diff
@@ -69,6 +70,8 @@ func TestToUnstructured(t *testing.T) {
 		{Data: `{"a":1}`, Expected: map[string]interface{}{"a": int64(1)}},
```
jpbetz, line 70: Test with trailing whitespace as well, e.g. `{"a": 1} ` ?

value/reflectcache_test.go
```diff
@@ -69,6 +70,8 @@ func TestToUnstructured(t *testing.T) {
...
 		{Data: `0`, Expected: int64(0)},
 		{Data: `0.0`, Expected: float64(0)},
+		{Data: `{}}`, ExpectedErrorMessage: "error decoding object from json: unexpected trailing data after object end"},
```
jpbetz, line 73: It is unfortunate for a case like `{ "a": 1, "b": {}}, "c": 3}` that this error message does not indicate where in input string the error was found, since that could help the author of the JSON find and correct the mistake easier.  Can we use `InputOffset` somehow to improve the error?

jpbetz:
/approve

/assign @alexzielenski
Would you be willing to be an extra pair of eyes on this one?

jpbetz:
/approve
(For real now that I've been added to OWNERS)

value/reflectcache.go
```diff
@@ -389,6 +391,11 @@ func unmarshal(data []byte, v interface{}) error {
 		if err := decoder.Decode(v); err != nil {
 			return err
 		}
+		next := decoder.InputOffset()
+		if _, err := decoder.Token(); !errors.Is(err, io.EOF) {
+			tail := bytes.TrimLeft(data[next:], " \t\r\n")
+			return fmt.Errorf("unexpected trailing data at offset %d", len(data)-len(tail))
+		}
```
sttts, line 398: before adding more complexity here: any reason this function is not just like the following?
```golang
// unmarshal unmarshals the given data
// If v is a *map[string]interface{}, numbers are converted to int64 or float64
func unmarshal(data []byte, v interface{}) error {
	d := json.NewDecoder(bytes.NewReader(data))
	d.UseNumber()
	if err := d.Decode(&v); err != nil {
		return err
	}

	next := d.InputOffset()
	if _, err := d.Token(); !errors.Is(err, io.EOF) {
		tail := bytes.TrimLeft(data[next:], " \t\r\n")
		return fmt.Errorf("unexpected trailing data at offset %d", len(data)-len(tail))
	}

	// If the decode succeeds, post-process the map to convert json.Number objects to int64 or float64
	switch v := v.(type) {
	case *map[string]interface{}:
		return convertMapNumbers(*v, 0)
	case *[]interface{}:
		return convertSliceNumbers(*v, 0)
	default:
		return nil
	}
}
```

sttts: It tests green.

benluddy: That seems like it would be equivalent in all the ways `unmarshal` is currently used. If a `v` with dynamic type `*interface{}` is passed, then JSON number inputs would unmarshal to `json.Number` with this change instead of `float64`.

sttts: right, and I would claim this case was kind of an oversight. We should call convertNumber.

sttts: What I mean is that it decoded to float64, while integers should become int64.

sttts:
/lgtm

/hold

in case you want to call `convertNumber` for the `*interface{}` case. I think it would be more uniform und expected. Remove hold at will.

benluddy:
> in case you want to call `convertNumber` for the `*interface{}` case. I think it would be more uniform und expected.

Updated to avoid any surprises in the future.

PR #262, opened 2024-07-09
Bump dependencies

lucacome:
Bumps dependencies


lucacome:
Seems like `vendor` wasn't updated in a while. Happy to delete the directory, let me know.

cici37:
/ok-to-test
I checked the version in transition repo(k/k) and it will not break the transition dependency rule we maintain in k8s.

go.mod
```diff
@@ -1,11 +1,10 @@
 module sigs.k8s.io/structured-merge-diff/v4
 
+go 1.13
```
cici37, line 3: Should the go version also bumped?

PR #265, opened 2024-10-29
Accept filter instead of exclude mask for ignored fields

jpbetz:
/sig api-machinery

https://github.com/kubernetes/kubernetes/pull/128266 needs a way of expressing reset fields that can express "exclude all fields except for the container resource fields modifiable via /resize".

This provides a way to pass a filter instead of a field exclusion set.


fieldpath/set.go
```diff
@@ -136,6 +137,113 @@ func (s *Set) EnsureNamedFieldsAreMembers(sc *schema.Schema, tr schema.TypeRef)
...
+		part := parts[i]
+		var pattern PathPattern
+		switch t := part.(type) {
+		case PathPattern:
+			pattern = t
+		case PathElement:
+			pattern = PathPattern{PathElement: t}
+		case string:
+			pattern = PathPattern{PathElement: PathElement{FieldName: &t}}
+		case int:
```
AnishShah, line 165: Should we add a case for key-value fields similar to what we have for `MakePath` func:

https://github.com/kubernetes-sigs/structured-merge-diff/blob/cf09e71f4508bfc416135048e392a8a4990bf4e3/fieldpath/path.go#L94-L103

fieldpath/set.go
```diff
@@ -503,3 +638,50 @@ func (s *SetNodeMap) Leaves() *SetNodeMap {
 	}
 	return out
 }
+
+// Filter defines an interface for filtering a set.
```
thockin, line 642: It's a little confusing to someone not steeped in this area - some of these are talking about EXCLUDE and some are talking about INCLUDE - is there a way to make them clearer?  Maybe I am just not the audience for it :)

fieldpath/set.go
jpbetz: Added.  `PathElement` made it possible to handle all cases, but parity with `MakePath` is a good reason these two other cases.

fieldpath/set.go
jpbetz: I've expanded this comment out significantly. Hopefully that makes it more clear.

jpbetz:
/assign @thockin 
Do you have reviewer permissions in this repo?  @AnishShah doesn't have /lgtm in this relm.

thockin:
I don't think I do.  I am trying to understand it and figure out how much I want to complain about the structure of it vs. accepting it as-is :)

jpbetz:
> I don't think I do. I am trying to understand it and figure out how much I want to complain about the structure of it vs. accepting it as-is :)

I don't mean to rush you.  I'm thinking purely about the logistics of how to get a PR in this repo merged when it is ready.  I'll hold to merge on your review.

fieldpath/set.go
```diff
@@ -503,3 +652,50 @@ func (s *SetNodeMap) Leaves() *SetNodeMap {
...
+	excludeSet *Set
+}
+
+func (t excludeFilter) Filter(set *Set) *Set {
+	return set.RecursiveDifference(t.excludeSet)
+}
+
+// NewPatternFilter returns a filter that only includes field paths that match the pattern.
+// PrefixPattern and MustPrefixPattern can help create basic SetPatterns.
+func NewPatternFilter(pattern *SetPattern) Filter {
```
thockin, line 691: Should this be `NewIncludeFilter` or `NewIncludePatternFilter`

fieldpath/set.go
```diff
@@ -136,6 +138,126 @@ func (s *Set) EnsureNamedFieldsAreMembers(sc *schema.Schema, tr schema.TypeRef)
...
+	current := MatchAnySet() // match all field patch suffixes
+	for i := len(parts) - 1; i >= 0; i-- {
+		part := parts[i]
+		var pattern PathPattern
+		switch t := part.(type) {
+		case PathPattern:
+			pattern = t
+		case PathElement:
+			pattern = PathPattern{PathElement: t}
+		case *value.FieldList:
```
thockin, line 170: Is there an example where this is used?  I am trying to understand how it works.

fieldpath/set.go
```diff
@@ -136,6 +138,126 @@ func (s *Set) EnsureNamedFieldsAreMembers(sc *schema.Schema, tr schema.TypeRef)
...
+}
+
+// MatchAnySet returns a SetPattern that matches any set.
+func MatchAnySet() *SetPattern {
+	return &SetPattern{Wildcard: true}
+}
+
+// SetPattern defines a pattern that matches fields in a Set.
+// SetPattern is structured much like a Set but with wildcard support.
+type SetPattern struct {
```
thockin, line 206: It's a little awkward being singular here, when it's possibly plural.  This is made a little worse because `NewPatternFilter()` takes a single pattern and `PrefixPattern()` constructs a `SetPattern` with one member.

How is one supposed to construct a pattern with multiple members?  The use-case in question needs multiple, I think.

The recursive use of `SetPattern` is tricky, but maybe allows something like the following:

from the existing test:

```
filter := NewPatternFilter(MustPrefixPattern("spec", "containers", MatchAnyPathElement(), "resources"))
```

could maybe be:

```
filter := NewPatternFilter(
    MustMakePattern("spec", "containers", MatchAnyPathElement(), "resources"),
    MustMakePattern("spec", "initContainers", MatchAnyPathElement(), "resources"),
    MustMakePattern("spec", "resources"),
)
```

And we would combine those into the final `SetPattern` (if len(args) == 1, just use it, else combine them)?

fieldpath/set.go
```diff
@@ -136,6 +138,126 @@ func (s *Set) EnsureNamedFieldsAreMembers(sc *schema.Schema, tr schema.TypeRef)
...
+// The pattern parts may any of:
+//
+//   - PathPattern - for wildcards, `MatchAnyPathElement()` can be used as well.
+//   - PathElement - for any path element
+//   - value.FieldList - for associative list keys
+//   - value.Value - for scalar list elements
+//   - string - For field names
+//   - int - for array indices
+func PrefixPattern(parts ...interface{}) (*SetPattern, error) {
+	current := MatchAnySet() // match all field patch suffixes
```
thockin, line 161: s/patch/path ?

fieldpath/set.go
```diff
@@ -136,6 +138,126 @@ func (s *Set) EnsureNamedFieldsAreMembers(sc *schema.Schema, tr schema.TypeRef)
 	}
 }
 
+// MustPrefixPattern is the same as PrefixPattern except it panics if parts can't be
+// turned into a SetPattern.
+func MustPrefixPattern(parts ...interface{}) *SetPattern {
```
thockin, line 143: I'd argue for "Make" in the names of these.

thockin (commented):
I wrote and then deleted a dozen comments, as I understood it better on each reading.  I did not review all the set logic, but looking at it for the ergonomics of defining patterns.

fieldpath/set.go
```diff
@@ -503,3 +652,50 @@ func (s *SetNodeMap) Leaves() *SetNodeMap {
...
+// NewPatternFilter can be used to create a filter that removes all fields except
+// the fields that match a field path pattern. PrefixPattern and MustPrefixPattern
+// can be used to define field path patterns.
+type Filter interface {
+	// Filter returns a filtered copy of the set.
+	Filter(*Set) *Set
+}
+
+// NewExcludeFilter returns a filter that removes field paths in the exclude set.
+func NewExcludeFilter(exclude *Set) Filter {
```
thockin, line 668: Maybe `NewExcludeSetFilter` so the names fit a pattern of `New<In/Ex>clude<Source>Filter()` ?

fieldpath/set.go
AnishShah: `MakePrefixPatternOrDie` similar to `MakePathOrDie` in other file? 

https://github.com/kubernetes-sigs/structured-merge-diff/blob/f034ec0f9a297cb014e5b5f7f4b4c9e4545777eb/fieldpath/path.go#L112

fieldpath/set.go
jpbetz: :+1:  I've gone with `NewIncludePatternFilter`.

fieldpath/set.go
jpbetz: I've added some tests.  It's used to define a listMap key field path part (since this system supports dereferecing by list map keys).  Example usage:

```go
NewIncludePatternFilter(MustPrefixPattern("spec", "listMap", &value.FieldList{
  {Name: "key1", Value: value.NewValueInterface("value1")},
  {Name: "key2", Value: value.NewValueInterface("value2")},
})),
```

fieldpath/set.go
jpbetz: A plural name makes sense to me.  How about `SetMatcher`?

```go
filter := NewPatternFilter(
    MustMakePattern("spec", "containers", MatchAnyPathElement(), "resources"),
    MustMakePattern("spec", "initContainers", MatchAnyPathElement(), "resources"),
    MustMakePattern("spec", "resources"),
)
```

This seems like a reasonable way define matcher that branches off in multiple directions.  `SetPattern` does support multiple branches, but hand crafting one as a data literal is non-egronomic to the extreme.

The easiest way to support might be to accept multiple matchers and then union them together.

fieldpath/set.go
jpbetz: I've gone with `MakePrefixMatcherOrDie` to conform to the conventions of the most adjacent code.  Feel free to argue for a different convention.

fieldpath/set.go
jpbetz: Added merge support and modified `NewIncludeMatcherFilter()` to take matchers as varargs, then merge them together before filtering.

I need someone to double-check me for the  "wildcard takes precedence" test. Is this what we want?

jpbetz:
@thockin @AnishShah feedback applied.

merge/update.go
```diff
@@ -189,13 +189,9 @@ func (s *Updater) Apply(liveObject, configObject *typed.TypedValue, version fiel
 		return nil, fieldpath.ManagedFields{}, fmt.Errorf("failed to get field set: %v", err)
 	}
 
-	ignored := s.IgnoredFields[version]
-	if ignored != nil {
-		set = set.RecursiveDifference(ignored)
-		// TODO: is this correct. If we don't remove from lastSet pruning might remove the fields?
-		if lastSet != nil {
-			lastSet.Set().RecursiveDifference(ignored)
-		}
```
jpbetz, line 198: I removed these lines because they are a no-op. Note that the result of the operation is not assigned to lastSet.

fieldpath/set_test.go
```diff
@@ -755,3 +756,255 @@ func TestSetNodeMapIterate(t *testing.T) {
...
+		{
+			name: "container resize fields: filter status and metadata",
+			input: NewSet(
+				MakePathOrDie("spec"),
+				MakePathOrDie("status"),
+				MakePathOrDie("metadata"),
+			),
+			filter: NewIncludeMatcherFilter(MakePrefixMatcherOrDie("spec", "containers", MatchAnyPathElement(), "resources")),
+			expect: NewSet(
+				MakePathOrDie("spec"),
```
thockin, line 814: Semantic question - is this the right result?  Spec isn't a "leaf" and it didn't actually match the filter (it matched part of the filter).  I don't know whis codebase well enough, but that was a little surprising to me.  Yay for tests.

fieldpath/set_test.go
```diff
@@ -755,3 +756,255 @@ func TestSetNodeMapIterate(t *testing.T) {
...
+					}),
+				MakePathOrDie("spec", "listMap",
+					&value.FieldList{
+						{Name: "key1", Value: value.NewValueInterface("value1")},
+						{Name: "key2", Value: value.NewValueInterface("value2")},
+					}, "field"),
+			),
+		},
+		{
+			name: "filter listMap key",
```
thockin, line 884: I can't tell what is different between this case and the previous, with the same name?

fieldpath/set_test.go
```diff
@@ -755,3 +756,255 @@ func TestSetNodeMapIterate(t *testing.T) {
...
+		{
+			name: "filter by index",
+			input: NewSet(
+				MakePathOrDie("spec"),
+				MakePathOrDie("spec", "list", 0, "value"),
+				MakePathOrDie("spec", "list", 1, "value"),
+			),
+			filter: NewIncludeMatcherFilter(MakePrefixMatcherOrDie("spec", "list", 1, "value")),
+			expect: NewSet(
+				MakePathOrDie("spec"),
```
thockin, line 943: Unclear to me why this has "spec" and "spec.list[1].value" but not "spec.list" and "spec.list[1]" as intermediate steps.  But "wildcard takes precedence" case below does not match "spec" ?

fieldpath/set_test.go
jpbetz: The background here is that these filters are operating against SSA managed fields.  BUT, the filters are not aware of the SSA field granularity (the atomic vs. granular thing).  Worse, the SSA field granularity could change and it would probably be better if this is durable to such changes.

If we include parent fields in matches, then we're saying: "I don't care what the SSA granularity is, I logically want these fields kept, and if the managed fields cause the field set to terminate at some parent because it's atomic, that's fine."

fieldpath/set_test.go
jpbetz: Thanks for catching.

Looks like I fat fingered something while doing my copy-paste a previous test then modify it to create my next test.  I've removed.

fieldpath/set_test.go
jpbetz: I was just inconsistent about what included in the inputs.

I've done a sweep to add all parents to inputs.  Hopefully this helps clarify what exactly exact match does.

thockin (commented):
LGTM, but I am not approver here.  I understand the design, though.

For posterity, Joe and I discussed this in real time.  He tells me it's fine and does not change SSA behavior:

Given input (indent for clarity only), which has exactly two "leaf" fields:
```
{
    "foo"                         /* struct */,
        "foo.bar"                 /* struct */,
            "foo.bar.bat"         /* struct */,
                "foo.bar.bat.qux" /* LEAF string */,
        "foo.zrb",                /* LEAF string */
}
```

Consider a strategy that has `GetResetFields() => "foo.bar.bat"`.  That would produce:
```
{
    "foo", "foo.bar", "foo.zrb"
    // note: "foo.bar.bat" and its children are culled
    // note: "foo.bar" remains, though it is empty
}
```

If we inverted it to an `Include("foo.zrb")`, that would produce:
```
{
    "foo", "foo.zrb"
    // note: only path elements leading to "foo.zrb" remain
}
```

In particular, note that "foo.bar" is absent.  We didn't prune it from the old case, even though it is "empty".  It didn't match anything in the new case.  

In neither case did we match the "leaf" fields under "foo.bar.bat/*" (as expected).

fieldpath/set.go
jpbetz: Nice added.

fieldpath/set_test.go
jpbetz: Yes, that's correct.

fieldpath/set.go
```diff
@@ -17,6 +17,8 @@ limitations under the License.
 package fieldpath
 
 import (
+	"fmt"
+	"sigs.k8s.io/structured-merge-diff/v4/value"
```
cici37, line 21: nit: import sorting

cici37 (commented):
/lgtm
/hold in case others wanna take a look.

PR #270, opened 2024-11-16
Run go mod vendor

jpbetz:
Vendor is out of date, both with go.mod and with k/k.

Separately, I'll add a verify vendor presubmit.


Jefftree:
/lgtm

Dropping vendor is also an alternative we should look into: https://github.com/kubernetes-sigs/structured-merge-diff/issues/271

Issue #273, opened 2024-11-20
Merge would fail if an rhs partially specified object omits key fields despite they can be derived from paths


andrii-korotkov-verkada:
Hello, I hope your day is going well.
I'm encountering the errors like
```
associative list with keys has an element that omits key field "name" (and doesn't have default value)
```
The code in question is [here](https://github.com/kubernetes-sigs/structured-merge-diff/blob/dc61957ff9fb318c34de8e631186ad004169b124/typed/helpers.go#L212-L222).
Here are examples of manifests used:
Here's an example of a part of rhs (derived from `Compare` result, `Removed` field):
```
.spec.template.spec.containers[name="nginx"].terminationMessagePath
.spec.template.spec.containers[name="nginx"].terminationMessagePolicy
.spec.template.spec.containers[name="nginx"].livenessProbe.failureThreshold
.spec.template.spec.containers[name="nginx"].livenessProbe.successThreshold
.spec.template.spec.containers[name="nginx"].livenessProbe.timeoutSeconds
.spec.template.spec.containers[name="nginx"].ports[containerPort=8080,protocol="TCP"]
.spec.template.spec.containers[name="nginx"].ports[containerPort=8080,protocol="TCP"].containerPort
.spec.template.spec.containers[name="nginx"].ports[containerPort=8080,protocol="TCP"].protocol
```
So it doesn't have
```
.spec.template.spec.containers[name="nginx"].name
```
since it's not getting removed.
But trying to merge this kind of rhs seems to depend on having that field explicitly.
If I do post-processing using the code
```
// appendKeyFields appends the key fields like "name" to the set elements which have some entries keyed by those fields.
// This is needed to merge properly, see https://github.com/argoproj/argo-cd/issues/20792.
func appendKeyFields(set *fieldpath.Set) {
	keyFieldPaths := []fieldpath.Path{}
	set.Iterate(func(path fieldpath.Path) {
		for i := 0; i < len(path); i++ {
			if path[i].Key != nil {
				for _, keyField := range *path[i].Key {
					pathPartCopy := make([]fieldpath.PathElement, i+1)
					copy(pathPartCopy, path[:i+1])
					newPath := append(pathPartCopy, fieldpath.PathElement{FieldName: &keyField.Name})
					keyFieldPaths = append(keyFieldPaths, newPath)
				}
			}
		}
	})
	for _, path := range keyFieldPaths {
		if !set.Has(path) {
			set.Insert(path)
		}
	}
[...]

PR #275, opened 2024-12-07
feat: Adds ExtractItems option to include key fields

yongruilin:
This PR adds an option `AppendKeyFields` to the `ExtractItems` function. 

When doing extracting with a fieldpath set which doesn't include the key fields, with this option, the result would include the key field values. So that the extraction result would follow the schema without hitting key field missing errors.

Ref: #273

### DIFF
--- a/typed/remove_test.go
+++ b/typed/remove_test.go
@@ -885,3 +885,116 @@ func TestReversibleExtract(t *testing.T) {
 		})
 	}
 }
+
+type extractWithKeysTestCase struct {
+	name         string
+	rootTypeName string
+	schema       typed.YAMLObject
+	triplets     []extractTriplet
+}
+
+type extractTriplet struct {
+	object     typed.YAMLObject
+	set        *fieldpath.Set
+	wantOutput interface{}
+}
+
+var extractWithKeysCases = []extractWithKeysTestCase{{
+	name:         "associativeAndAtomicSchema",
+	rootTypeName: "myRoot",
+	schema:       typed.YAMLObject(associativeAndAtomicSchema),
+	triplets: []extractTriplet{
+		{
+			// extract with all key fields included
+			object: `{"list":[{"key":"nginx","id":1,"nv":2}]}`,
+			set: _NS(
+				_P("list", _KBF("key", "nginx", "id", 1), "key"),
+				_P("list", _KBF("key", "nginx", "id", 1), "id"),
+			),
+			wantOutput: typed.YAMLObject(`{"list":[{"key":"nginx","id":1}]}`),
+		},
+		{
+			// extract no key field included
+			object: `{"list":[{"key":"nginx","id":1,"nv":2}]}`,
+			set: _NS(
+				_P("list", _KBF("key", "nginx", "id", 1), "nv"),
+			),
+			wantOutput: typed.YAMLObject(`{"list":[{"key":"nginx","id":1, "nv":2}]}`),
+		},
+		{
+			// extract with partial keys included
+			object: `{"list":[{"key":"nginx","id":1,"nv":2}]}`,
+			set: _NS(
+				_P("list", _KBF("key", "nginx", "id", 1), "nv"),
+				_P("list", _KBF("key", "nginx", "id", 1), "id"),
+			),
+			wantOutput: typed.YAMLObject(`{"list":[{"key":"nginx","id":1, "nv":2}]}`),
+		},
+		{
+			// extract with null field value
+			object: `{"list":[{"key":"nginx","id":1,"nv":2}]}`,
+			set: _NS(
+				_P("list", _KBF("key", "nginx", "id", 1), "value"),
+			),
+			wantOutput: map[string]interface{}{
+				"list": []interface{}{nil},
+			},
+		},
+	},
+}}
+
+func (tt extractWithKeysTestCase) test(t *testing.T) {
+	parser, err := typed.NewParser(tt.schema)
+	if err != nil {
+		t.Fatalf("failed to create schema: %v", err)
+	}
+	pt := parser.Type(tt.rootTypeName)
+
+	for i, triplet := range tt.triplets {
+		triplet := triplet
+		t.Run(fmt.Sprintf("%v-valid-%v", tt.name, i), func(t *testing.T) {
+			t.Parallel()
+			// source typedValue obj
+			tv, err := pt.FromYAML(triplet.object)
+			if err != nil {
+				t.Fatal(err)
+			}
+			gotExtracted := tv.ExtractItems(triplet.set, typed.WithAppendKeyFields())
+
+			switch triplet.wantOutput.(type) {
+			case typed.YAMLObject:
+				wantOut, err := pt.FromYAML(triplet.wantOutput.(typed.YAMLObject))
+				if err != nil {
+					t.Fatalf("unable to parser/validate removeOutput yaml: %v\n%v", err, triplet.wantOutput)
+				}
+
+				if !value.Equals(gotExtracted.AsValue(), wantOut.AsValue()) {
+					t.Errorf("ExtractItems expected\n%v\nbut got\n%v\n",
+						value.ToString(wantOut.AsValue()), value.ToString(gotExtracted.AsValue()),
+					)
+				}
+			default:
+				// The extracted result
+				wantOut := value.NewValueInterface(triplet.wantOutput)
+				if !value.Equals(gotExtracted.AsValue(), wantOut) {
+					t.Errorf("ExtractItems expected\n%v\nbut got\n%v\n",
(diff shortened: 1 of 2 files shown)


typed/typed.go
```diff
@@ -187,7 +195,33 @@ func (tv TypedValue) RemoveItems(items *fieldpath.Set) *TypedValue {
...
+				if !tvPathSet.Has(path) {
+					return
+				}
+				for i, pe := range path {
+					if pe.Key == nil {
+						continue
+					}
+					for _, keyField := range *pe.Key {
+						keyName := keyField.Name
+						keyFieldPath := append(path[:i+1:i+1], fieldpath.PathElement{FieldName: &keyName})
```
andrii-korotkov-verkada, line 217: We may have to make a copy of path, at least I've run into some issues when not doing so in my workaround.

yongruilin: The `copy` doesn't make the difference here as the fields in the `PathElement` are pointers and they would still be copied to the new path, so that the new path still points to the same `PathElement`.

I think what matters is creating a new string `keyName := keyField.Name` in the for-loop, so that the appended `PathElement.FieldName` won't be overwritten on the same address.

Or, could you share the issues you encounter when not doing copy?

andrii-korotkov-verkada: Ah, maybe you are right. I have https://github.com/argoproj/gitops-engine/pull/640/commits/e881697fbaf4e3fc770eb7e0dd365606e4cffe8c, but I indeed didn't create a new string, so that might have been an issue.

typed/typed.go
```diff
@@ -32,6 +32,14 @@ const (
...
 )
 
+// ExtractItemsOptions is the list of all the options available when extracting items.
+type ExtractItemsOptions int
+
+const (
+	// AppendKeyFields means that when extracting items, the key field would also be included.
+	AppendKeyFields ExtractItemsOptions = iota
+)
+
```
jpbetz, line 42: Are we confident all future options will be flags?  Alternative form that would handle arbitrary option params: https://github.com/kubernetes/kube-openapi/pull/519/files#diff-72f8260914152fe9557b05d756ec94854184645e991ff755855628428d6d7a18R410-R421

typed/typed.go
```diff
@@ -187,7 +195,33 @@ func (tv TypedValue) RemoveItems(items *fieldpath.Set) *TypedValue {
...
 
 // ExtractItems returns a value with only the provided list or map items extracted from the value.
-func (tv TypedValue) ExtractItems(items *fieldpath.Set) *TypedValue {
+func (tv TypedValue) ExtractItems(items *fieldpath.Set, opts ...ExtractItemsOptions) *TypedValue {
+	for _, opt := range opts {
+		switch opt {
+		case AppendKeyFields:
+			tvPathSet, err := tv.ToFieldSet()
+			if err != nil {
+				continue
```
jpbetz, line 204: Can we include a comment explaining why continue is safe and correct here?

typed/typed.go
jpbetz: I'm fine with setting capacity here as a way to force an inline copy to be created.  I also wouldn't mind an extra comment for future readers stating what happens here since this is use of the language that not all readers may pick up on easily.

jpbetz (commented):
Two suggestions then LGTM

typed/typed.go
yongruilin: SG, updated using this format.

typed/typed.go
yongruilin: Added the comment.

typed/typed.go
yongruilin: I've added the comment.

typed/typed.go
```diff
@@ -187,7 +201,41 @@ func (tv TypedValue) RemoveItems(items *fieldpath.Set) *TypedValue {
...
+	for _, opt := range opts {
+		opt(options)
+	}
+	if options.appendKeyFields {
+		tvPathSet, err := tv.ToFieldSet()
+		if err != nil {
+			// It is safe to continue and skip appending the key fields, because
+			// if there is error to convert to fieldPath.Set, there is
+			// no way to find the key fields values by following the schema.
+			goto startExtrcation
```
leoluz, line 215:
```suggestion
			goto startExtraction
```

typed/typed.go
```diff
@@ -187,7 +201,41 @@ func (tv TypedValue) RemoveItems(items *fieldpath.Set) *TypedValue {
...
+					// This ensures that appending to keyFieldPath creates a new underlying array, avoiding accidental
+					// modification of the original slice (path).
+					keyFieldPath := append(path[:i+1:i+1], fieldpath.PathElement{FieldName: &keyName})
+					keyFieldPathSet.Insert(keyFieldPath)
+				}
+			}
+		})
+		items = items.Union(keyFieldPathSet)
+	}
+startExtrcation:
```
leoluz, line 238:
```suggestion
startExtraction:
```

typed/typed.go
```diff
@@ -32,6 +32,20 @@ const (
 	AllowDuplicates ValidationOptions = iota
 )
 
+// ExtractItemsOptions is the options available when extracting items.
```
leoluz, line 35:
```suggestion
// extractItemsOptions is the options available when extracting items.
```

typed/typed.go
leoluz: IMO, it would be preferable to return the error and avoid the `goto` than failing silently. I understand that it would change the signature and break consumers. Maybe a new function?

typed/typed.go
yongruilin: Updated.

typed/typed.go
yongruilin: Updated

typed/remove_test.go
```diff
@@ -906,3 +885,116 @@ func TestReversibleExtract(t *testing.T) {
...
+		{
+			// extract no key field included
+			object: `{"list":[{"key":"nginx","id":1,"nv":2}]}`,
+			set: _NS(
+				_P("list", _KBF("key", "nginx", "id", 1), "nv"),
+			),
+			wantOutput: typed.YAMLObject(`{"list":[{"key":"nginx","id":1, "nv":2}]}`),
+		},
+		{
+			// extract with patial keys included
```
Jefftree, line 925:
```suggestion
			// extract with partial keys included
```

yongruilin: Updated. thanks!

typed/typed.go
yongruilin: It is a good practice to expose the error. But I'm not a fan to add a new interface here. We should address the non-error return interface(`RemoveItems`,`ExtractItems`) in another PR. wdyt?

typed/typed.go
```diff
@@ -187,7 +201,41 @@ func (tv TypedValue) RemoveItems(items *fieldpath.Set) *TypedValue {
...
 // ExtractItems returns a value with only the provided list or map items extracted from the value.
-func (tv TypedValue) ExtractItems(items *fieldpath.Set) *TypedValue {
+func (tv TypedValue) ExtractItems(items *fieldpath.Set, opts ...extractItemsOption) *TypedValue {
+	options := &extractItemsOptions{}
+	for _, opt := range opts {
+		opt(options)
+	}
+	if options.appendKeyFields {
+		tvPathSet, err := tv.ToFieldSet()
+		if err != nil {
```
leoluz, line 211: What if we do `if err == nil { <Add the logic to set the keyname here> }` ?
This way I think we could get rid of the `goto`. 
Wdyt?

typed/typed.go
leoluz: LGTM. 👍🏻

typed/typed.go
yongruilin: Sounds good. Updated.

typed/typed.go
```diff
@@ -187,7 +201,37 @@ func (tv TypedValue) RemoveItems(items *fieldpath.Set) *TypedValue {
 }
 
 // ExtractItems returns a value with only the provided list or map items extracted from the value.
-func (tv TypedValue) ExtractItems(items *fieldpath.Set) *TypedValue {
+func (tv TypedValue) ExtractItems(items *fieldpath.Set, opts ...extractItemsOption) *TypedValue {
```
jpbetz, line 204:
```suggestion
func (tv TypedValue) ExtractItems(items *fieldpath.Set, opts ...ExtractItemsOption) *TypedValue {
```

Requiring a non-exported type as an argument of an exported type is problematic. 

One way to detect this would be to use the test package in `typed/remove_test.go`.  I don't think you need necessarily to do that here, but that's a good way to test against only the exported API.

yongruilin: My original thought was that putting it non-exported can force user to only use the exported `Withxxx()` function to construct the options.  But it indeed looks strange to put it in the exported function. 

Updated.

Issue #276, opened 2024-12-18
Add apidiff to prevent accidental breaking changes


yongruilin:
https://pkg.go.dev/golang.org/x/exp/apidiff can be used to determines whether two examples of a package or module are compatible. We can add a script and run it against this repo to prevent incompatible changes. Then potentially tune it to be a blocking presubmit job.

PR #279, opened 2025-01-21
fix: update apidiff script to install tool and set up remote repository

yongruilin:
fix error when execute apidiff.sh
1. 
```
./hack/apidiff.sh: line 58: GOBIN: unbound variable
```

2. 
```
./hack/apidiff.sh: line 98: apidiff: command not found
```
3. 
```
fatal: Not a valid object name origin/master
``` 

ref: https://github.com/kubernetes-sigs/structured-merge-diff/issues/276

### DIFF
--- a/hack/apidiff.sh
+++ b/hack/apidiff.sh
@@ -24,7 +24,7 @@ usage() {
     echo "   -t <revision>: Report changes in code up to and including this revision."
     echo "                  Default is the current working tree instead of a revision."
     echo "   -r <revision>: Report changes in code added since this revision."
-    echo "                  Default is the common base of origin/master and HEAD."
+    echo "                  Default is the common base of master and HEAD."
     exit 1
 }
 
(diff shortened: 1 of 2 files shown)


hack/apidiff.sh
```diff
@@ -55,10 +55,21 @@ fi
...
+    GOBIN=${GOBIN:-$(go env GOPATH)/bin}
     echo "Installing apidiff into ${GOBIN}."
     go install golang.org/x/exp/cmd/apidiff@latest
+    # Add GOBIN to PATH
+    export PATH=$PATH:${GOBIN}
 fi
 
+# Ensure the remote is set
+if ! git remote get-url origin &> /dev/null; then
+    git remote add origin https://github.com/kubernetes-sigs/structured-merge-diff.git
```
jpbetz, line 67: Would an error be more appropriate?  Modifying git configuration as a side effect makes me slightly uncomfortable.

jpbetz:
@BenTheElder Would you be willing to do a bash review?  I'm not amazing at scripting.  At a glance this looks like it does what we expect.  I'm mostly curious if you know if better ways to do any of these things or if you see any hidden footguns.

hack/apidiff.sh
BenTheElder: Yeah, same, the other problem is that origin is the default so it may be the user's fork.

BenTheElder: Usually we instruct people to use "upstream" (the original repo under kubernetes) and "origin" (your fork) as the remotes, but e can't guarantee it.

see also: the apidiff script in kubernetes/kubernetes

hack/apidiff.sh
```diff
@@ -55,10 +55,21 @@ fi
 
 # Check for apidiff tool, install it if not found
 if ! command -v "${API_DIFF_TOOL}" &> /dev/null; then
+    GOBIN=${GOBIN:-$(go env GOPATH)/bin}
```
BenTheElder, line 58: typically it would be more polite to use a temporary local GOBIN instead of writing binaries to the system GOBIN

in kubernetes/kubernetes build outputs go under `_output/` (which is gitignored) and we create a temp GOBIN under there.

the actual compile cache will be reused from the system but we'll avoid modifying binaries in the real user PATH

yongruilin: Updated with a local GOBIN and output to `_output/`.

BenTheElder: it looks like we also need .gitignore, https://github.com/kubernetes-sigs/structured-merge-diff/blob/master/.gitignore only has `.idea`

I guess being a library, we don't have a standard build output location yet

BenTheElder:
BTW re: bash, when in doubt I highly recommend shellcheck.net, we use this in k/k with very few options set. (We do ignore a few of the lints, but not many). There are IDE plugins.

hack/apidiff.sh
```diff
@@ -53,16 +53,39 @@ if [ "$#" -ge 1 ]; then
     TARGET_DIR="$1"
 fi
 
+# Ensure LOCAL_GOBIN is an absolute path
+LOCAL_GOBIN="$(pwd)/_output/bin"
+mkdir -p "${LOCAL_GOBIN}"
+
+# Add cleanup trap to remove the temporary binary directory
+trap 'rm -rf "${LOCAL_GOBIN}"' EXIT
```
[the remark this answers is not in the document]
yongruilin: I combined all into the cleanup function.

hack/apidiff.sh
yongruilin: I think it should be fine use delete the created directory after execution.

hack/apidiff.sh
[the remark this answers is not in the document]
BenTheElder: I meant that if we have any other scripts install tools, this script is clobbering the entire fixed gobin, instead we could probably just leave it behind and leave removing it to something like `make clean` (manually invoked), or we should use a temp path that is specific to this script run. (like mktemp -d under _output)

hack/apidiff.sh
```diff
@@ -53,30 +53,65 @@ if [ "$#" -ge 1 ]; then
...
+            exit 1
+        fi
+    fi
 fi
 
+# Verify base commit exists
+REFERENCE_REVISION="$(git rev-parse --verify "${REFERENCE_REVISION}")"
+
 # Step 1: Create a temporary directory for worktrees
 TMP_DIR=$(mktemp -d)
```
BenTheElder, line 101: my suggestion: use the shared tempdir for work scoped to this script run..

create underneath this a gopath and a worktree subdir

use a shared exit trap to remove the whole thing.

you may want to create the tempdir under something like _output, for performance reasons if you're doing worktree etc (otherwise they may be on different filesystems), you can do this with an argument to mktemp

BenTheElder: and gitignore _output (or similar), whatever paths we use within the repo that are for builds/intermediate outputs

yongruilin: Thanks for the suggestion, I updated to move things into the shared temp dir under _output

BenTheElder:
I think this looks right, thanks!

jpbetz:
/approve

Thanks @BenTheElder for reviewing this one!

PR #280, opened 2025-01-23
KEP-5051: Add opt-in support for unset markers

jpbetz:
This implements the SMD bits of https://github.com/kubernetes/enhancements/pull/5052.

All the new functionality is opt-in. No default behavior is changed. This is extensively tested.

Benchmark data suggests negligible performance impact: https://gist.github.com/jpbetz/df3806a5605a174597a3d0d6459125c5

### DIFF
--- a/typed/markers.go
+++ b/typed/markers.go
@@ -1,0 +1,321 @@
+/*
+Copyright 2025 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package typed
+
+import (
+	"fmt"
+	"sync"
+
+	"sigs.k8s.io/structured-merge-diff/v4/fieldpath"
+	"sigs.k8s.io/structured-merge-diff/v4/schema"
+	"sigs.k8s.io/structured-merge-diff/v4/value"
+)
+
+const (
+	// markerKey is the key used to store marker values in the object.
+	markerKey = "k8s_io__value"
+
+	// unsetMarkerValue is the value used to indicate that a marker is unset.
+	unsetMarkerValue = "unset"
+)
+
+var mPool = sync.Pool{
+	New: func() interface{} { return &markerExtractorWalker{} },
+}
+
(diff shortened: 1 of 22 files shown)


typed/markers.go
```diff
@@ -0,0 +1,321 @@
...
+	"sync"
+
+	"sigs.k8s.io/structured-merge-diff/v4/fieldpath"
+	"sigs.k8s.io/structured-merge-diff/v4/schema"
+	"sigs.k8s.io/structured-merge-diff/v4/value"
+)
+
+const (
+	// markerKey is the key used to store marker values in the object.
+	markerKey = "k8s_io__value"
```
jpbetz, line 30: @deads2k @sttts Suggestions welcome on how to name the key/value of the markers used in the proposal.  BTH, I'm actively looking for a better string.  Having a key that is also a valid CEL identifier is important for MutatingAdmissionPolicy, so `[a-z][A-Z][0-9]_` (non-number first char) is strongly preferred.

jpbetz:
/retest
Now that apidiff PR has merged.

typed/markers.go
```diff
@@ -0,0 +1,321 @@
...
+
+	if t.ElementRelationship == schema.Atomic {
+		// If the map is atomic, we need to check for unset markers before we descend further.
+		m.Iterate(func(key string, val value.Value) bool {
+			if isUnsetMarker(val) {
+				errs = append(errs, ValidationErrors{{ErrorMessage: "markers are not allowed in the contents of atomics"}}.WithPath(v.path.String())...)
+			}
+			return true
+		})
+		return errs
```
yongruilin, line 316: Do we descend further if this map is atomic?

typed/markers.go
```diff
@@ -0,0 +1,321 @@
...
+	}
+	return nil
+}
+
+func (v *markerExtractorWalker) nearestElementRelationship() *schema.ElementRelationship {
+	for i := len(v.parentElementRelationship) - 1; i >= 0; i-- {
+		if v.parentElementRelationship[i] != nil {
+			return v.parentElementRelationship[i]
+		}
+	}
```
yongruilin, line 128: We're only looking for the nearestElementRelationship, what if it is nested? e.g. an atomic List which contains maps(not atomic), 
```
              types:
              - name: TestType
                map:
                  fields:
                  - name: items
                    type:
                      list:
                        elementType:
                          map:
                            fields:
                            - name: key
                              type:
                                scalar: string
                            - name: value
                              type:
                                scalar: string
                        elementRelationship: atomic
```
and applyconfig is `{"items": [{"key": "k1", "value": {"k8s_io__value": "unset"}}]}`, should we throw an error for it? The parentElementRelationship would be `["", "atomic", ""]` when validating the marker location.

Or, it should be `if v.parentElementRelationship[i] != nil &&  *v.parentElementRelationship[i] != "" && *v.parentElementRelationship[i] != "separable {`  to looking for the nearest "atomic" or 'associative"?

jpbetz:
/close
The KEP for this did not get approved.

Issue #281, opened 2025-02-01
error "non-constant format string" with go 1.24


yongruilin:
When trying to run test locally but got error
```
# sigs.k8s.io/structured-merge-diff/v4/typed
typed/validate.go:160:17: non-constant format string in call to sigs.k8s.io/structured-merge-diff/v4/typed.errorf
typed/validate.go:196:17: non-constant format string in call to sigs.k8s.io/structured-merge-diff/v4/typed.errorf
FAIL    sigs.k8s.io/structured-merge-diff/v4/typed [build failed]
```
This is due to the update in go 1.24 https://github.com/golang/go/issues/60529

Need to update the error return to address.

Ref: 
- https://github.com/kubernetes/kubernetes/issues/127191
- https://github.com/kubernetes/kubernetes/pull/127422

PR #282, opened 2025-02-01
fix: update error formatting and test naming

yongruilin:
This is for addressing the change in go1.24
https://github.com/golang/go/issues/60529, which suppress the error on non-constant format and no args.

Fix: #281

### DIFF
--- a/typed/validate.go
+++ b/typed/validate.go
@@ -157,7 +157,7 @@ func (v *validatingObjectWalker) visitListItems(t *schema.List, list value.List)
 func (v *validatingObjectWalker) doList(t *schema.List) (errs ValidationErrors) {
 	list, err := listValue(v.allocator, v.value)
 	if err != nil {
-		return errorf(err.Error())
+		return errorf("%v", err)
 	}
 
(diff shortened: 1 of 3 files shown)


typed/validate.go
```diff
@@ -157,7 +157,7 @@ func (v *validatingObjectWalker) visitListItems(t *schema.List, list value.List)
 func (v *validatingObjectWalker) doList(t *schema.List) (errs ValidationErrors) {
 	list, err := listValue(v.allocator, v.value)
 	if err != nil {
-		return errorf(err.Error())
```
jpbetz, line 160: Should this be "%w" ?

yongruilin: It is following the same pattern as https://github.com/kubernetes-sigs/structured-merge-diff/blob/0ee1c6563b9c1026100560fef64164bde5fabbfd/typed/merge.go#L163

jpbetz: Ok, SGTM

jpbetz (approved):
/approve
/lgtm

PR #288, opened 2025-04-18
stop exporting jsoniter based apis

BenTheElder:
see: https://github.com/kubernetes-sigs/structured-merge-diff/issues/202

TODO: should we bump the module to v6 since this removes APIs?

I can't find evidence that anyone else was using these:

none in the org:
https://cs.k8s.io/?q=(ReadJSONIter)%7C(WriteJSONStream)&i=nope&literal=nope&files=&excludeFiles=vendor&repos=

only in `vendor/`:
https://grep.app/search?f.lang=Go&regexp=true&q=%28ReadJSONIter%29%7C%28WriteJSONStream%29

### DIFF
--- a/fieldpath/element.go
+++ b/fieldpath/element.go
@@ -21,7 +21,7 @@ import (
 	"sort"
 	"strings"
 
-	"sigs.k8s.io/structured-merge-diff/v4/value"
+	"sigs.k8s.io/structured-merge-diff/v6/value"
 )
 
 // PathElement describes how to select a child field given a containing object.
--- a/fieldpath/element_test.go
+++ b/fieldpath/element_test.go
@@ -19,7 +19,7 @@ package fieldpath
 import (
 	"testing"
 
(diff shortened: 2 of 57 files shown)


BenTheElder:
> TODO: should we bump the module to v6 since this removes APIs?

Yes, because apidiff fails :-)

BenTheElder:
NOTE: apidiff now fails because the module version was bumped.
https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_structured-merge-diff/288/pull-structured-merge-diff-apidiff/1913310145010470912

I could see an argument for either way.

/assign @jpbetz

jpbetz:
/approve
/lgtm
I agree this worth bumping major version for.

liggitt:
let's wait to release v6 until we're sure we've made all the public API changes we need?

PR #290, opened 2025-04-30
Fix CanOmit() to return true for zero valued struct

jpbetz:
This is needed to support omitempty inlined structs

### DIFF
--- a/value/jsontagutil.go
+++ b/value/jsontagutil.go
@@ -106,6 +106,8 @@ func isEmpty(v reflect.Value) bool {
 		return v.Float() == 0
 	case reflect.Interface, reflect.Ptr:
 		return v.IsNil()
+	case reflect.Struct:
+		return v.IsZero()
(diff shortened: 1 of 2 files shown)


yongruilin (commented):
presubmit test failed.
Looks like need to modify the testcase https://github.com/kubernetes-sigs/structured-merge-diff/blob/e8775feb43c2ee5a4883d791bce62d2669ac13a6/value/valuereflect_test.go#L381-L383
to
```
expectDeleted: map[string]interface{}{},
```

jpbetz:
/hold

This might not be correct.  ToUnstructured in k/k disagrees with this.  I need to dig in a bit more.

It might be that the current behavior is correct and we need to thread in `omitzero` to handle some of the edges.

jpbetz:
Looks like https://github.com/kubernetes/kubernetes/pull/130989 covers it

PR #291, opened 2025-05-01
Add copy and iterator functions to fieldpath types

jpbetz:
Sets of fields are useful throughout API machinery.  I'm currently working on extracting reachable fields from CEL expressions, and using fieldpath.Set makes a lot of sense.  But the code is a mess without these utilities.

Also bump to go 1.23 to pick up golang iterators.


yongruilin (commented):
/lgtm

PR #292, opened 2025-06-16
Migrate to encoding/json/v2

### LINKED ISSUE #202: remove use of json-iterator / reflect2
The approach taken by json-iterator and its reflect library is not long-term stable (it pins to internal details of the stdlib reflect package), and requires maintenance per go release, which does not appear to be sustainable https://github.com/golang/go/issues/48238#issuecomment-917000973

Minimum versions to work with go1.18+ are:
* github.com/json-iterator/go v1.1.12
* github.com/modern-go/reflect2 v1.0.2

In addition to the maintenance aspect, it is hard to reason about the safety of the implementation (https://github.com/golang/go/issues/48238#issuecomment-916331048)

I'm exploring what it would look like to use the stdlib json decoder as a base and add in the features we rely on (case-sensitive keys, preservation of ints/floats, etc) in https://github.com/kubernetes/kubernetes/pull/105030

I would encourage this project to consider the long-term sustainability/safety of the json-iterator dependency as well.

inteon:
Replaces the `github.com/json-iterator/go` dependency with `encoding/json/v2`.

```
goos: darwin
goarch: arm64
pkg: sigs.k8s.io/structured-merge-diff/v6/fieldpath
cpu: Apple M4 Pro
                              │ 1-master.txt │            9-final.txt             │
                              │    sec/op    │   sec/op     vs base               │
FieldSet/serialize-20-14         3.569µ ± 1%   5.757µ ± 2%  +61.31% (p=0.002 n=6)
FieldSet/serialize-pe-20-14      1.333µ ± 0%   1.366µ ± 1%   +2.44% (p=0.002 n=6)
FieldSet/deserialize-20-14       7.755µ ± 1%   9.646µ ± 1%  +24.38% (p=0.002 n=6)
FieldSet/serialize-50-14         10.11µ ± 0%   16.87µ ± 0%  +66.95% (p=0.002 n=6)
FieldSet/serialize-pe-50-14      2.444µ ± 1%   2.531µ ± 0%   +3.56% (p=0.002 n=6)
FieldSet/deserialize-50-14       20.17µ ± 0%   26.91µ ± 1%  +33.44% (p=0.002 n=6)
FieldSet/serialize-100-14        33.13µ ± 3%   54.32µ ± 1%  +63.97% (p=0.002 n=6)
FieldSet/serialize-pe-100-14     3.293µ ± 3%   3.363µ ± 0%        ~ (p=0.058 n=6)
FieldSet/deserialize-100-14      65.54µ ± 1%   87.16µ ± 2%  +32.98% (p=0.002 n=6)
FieldSet/serialize-500-14        177.5µ ± 2%   288.6µ ± 2%  +62.62% (p=0.002 n=6)
FieldSet/serialize-pe-500-14     3.675µ ± 1%   3.816µ ± 1%   +3.84% (p=0.002 n=6)
FieldSet/deserialize-500-14      337.8µ ± 8%   450.0µ ± 1%  +33.21% (p=0.002 n=6)
FieldSet/serialize-1000-14       392.0µ ± 1%   643.5µ ± 2%  +64.16% (p=0.002 n=6)
FieldSet/serialize-pe-1000-14    3.695µ ± 3%   3.835µ ± 1%   +3.79% (p=0.002 n=6)
FieldSet/deserialize-1000-14     759.5µ ± 2%   989.4µ ± 1%  +30.27% (p=0.002 n=6)
geomean                          19.96µ        26.00µ       +30.25%

                              │ 1-master.txt  │             9-final.txt             │
                              │     B/op      │     B/op      vs base               │
FieldSet/serialize-20-14          2412.0 ± 0%     954.0 ± 0%  -60.45% (p=0.002 n=6)
FieldSet/serialize-pe-20-14        935.0 ± 0%     269.0 ± 0%  -71.23% (p=0.002 n=6)
FieldSet/deserialize-20-14      11.143Ki ± 0%   7.562Ki ± 0%  -32.13% (p=0.002 n=6)
FieldSet/serialize-50-14         7.195Ki ± 0%   2.688Ki ± 0%  -62.64% (p=0.002 n=6)
FieldSet/serialize-pe-50-14       1748.0 ± 0%     503.0 ± 0%  -71.22% (p=0.002 n=6)
FieldSet/deserialize-50-14       23.60Ki ± 0%   21.04Ki ± 0%  -10.85% (p=0.002 n=6)
FieldSet/serialize-100-14       20.573Ki ± 0%   8.800Ki ± 0%  -57.23% (p=0.002 n=6)
FieldSet/serialize-pe-100-14      2321.0 ± 0%     672.0 ± 0%  -71.05% (p=0.002 n=6)
FieldSet/deserialize-100-14      71.69Ki ± 0%   72.46Ki ± 0%   +1.08% (p=0.002 n=6)
FieldSet/serialize-500-14       111.72Ki ± 0%   44.78Ki ± 0%  -59.92% (p=0.002 n=6)
FieldSet/serialize-pe-500-14      2615.0 ± 0%     751.0 ± 0%  -71.28% (p=0.002 n=6)
FieldSet/deserialize-500-14      348.8Ki ± 0%   371.0Ki ± 0%   +6.38% (p=0.002 n=6)
FieldSet/serialize-1000-14      244.52Ki ± 1%   97.04Ki ± 0%  -60.31% (p=0.002 n=6)
FieldSet/serialize-pe-1000-14     2616.0 ± 0%     752.0 ± 0%  -71.25% (p=0.002 n=6)
FieldSet/deserialize-1000-14     761.8Ki ± 0%   817.6Ki ± 0%   +7.34% (p=0.002 n=6)
geomean                          15.95Ki        7.567Ki       -52.56%

                              │ 1-master.txt │            9-final.txt             │
                              │  allocs/op   │  allocs/op   vs base               │
FieldSet/serialize-20-14          8.000 ± 0%    4.000 ± 0%  -50.00% (p=0.002 n=6)
FieldSet/serialize-pe-20-14       41.00 ± 0%    17.00 ± 0%  -58.54% (p=0.002 n=6)
FieldSet/deserialize-20-14        266.0 ± 0%    275.0 ± 0%   +3.38% (p=0.002 n=6)
FieldSet/serialize-50-14         13.000 ± 0%    9.000 ± 0%  -30.77% (p=0.002 n=6)
FieldSet/serialize-pe-50-14       76.00 ± 0%    32.00 ± 0%  -57.89% (p=0.002 n=6)
FieldSet/deserialize-50-14        777.0 ± 0%    808.0 ± 0%   +3.99% (p=0.002 n=6)
FieldSet/serialize-100-14         27.00 ± 0%    30.00 ± 0%  +11.11% (p=0.002 n=6)
FieldSet/serialize-pe-100-14     102.00 ± 0%    42.00 ± 0%  -58.82% (p=0.002 n=6)
FieldSet/deserialize-100-14      2.589k ± 0%   2.690k ± 0%   +3.90% (p=0.002 n=6)
FieldSet/serialize-500-14         120.0 ± 0%    151.0 ± 1%  +25.83% (p=0.002 n=6)
FieldSet/serialize-pe-500-14     115.00 ± 0%    47.00 ± 0%  -59.13% (p=0.002 n=6)
FieldSet/deserialize-500-14      13.27k ± 0%   13.82k ± 0%   +4.12% (p=0.002 n=6)
FieldSet/serialize-1000-14        259.0 ± 0%    337.0 ± 0%  +30.12% (p=0.002 n=6)
FieldSet/serialize-pe-1000-14    115.00 ± 0%    48.00 ± 0%  -58.26% (p=0.002 n=6)
FieldSet/deserialize-1000-14     29.41k ± 0%   30.61k ± 0%   +4.08% (p=0.002 n=6)
geomean                           211.7         155.0       -26.77%

pkg: sigs.k8s.io/structured-merge-diff/v6/value
                    │ 1-master.txt │            9-final.txt             │
                    │    sec/op    │   sec/op     vs base               │
JSON/serialize-14      4.549µ ± 1%   4.614µ ± 1%   +1.45% (p=0.041 n=6)
JSON/deserialize-14    2.999µ ± 1%   4.498µ ± 0%  +49.97% (p=0.002 n=6)
geomean                3.693µ        4.556µ       +23.35%

                    │ 1-master.txt │             9-final.txt             │
                    │     B/op     │     B/op      vs base               │
JSON/serialize-14     7.079Ki ± 0%   1.109Ki ± 0%  -84.33% (p=0.002 n=6)
JSON/deserialize-14   5.308Ki ± 0%   5.639Ki ± 0%   +6.24% (p=0.002 n=6)
geomean               6.130Ki        2.501Ki       -59.20%

                    │ 1-master.txt │            9-final.txt            │
                    │  allocs/op   │ allocs/op   vs base               │
JSON/serialize-14      148.00 ± 0%   94.00 ± 0%  -36.49% (p=0.002 n=6)
JSON/deserialize-14     138.0 ± 0%   156.0 ± 0%  +13.04% (p=0.002 n=6)
geomean                 142.9        121.1       -15.27%
```

closes https://github.com/kubernetes-sigs/structured-merge-diff/issues/202


dims:
xref: https://github.com/kubernetes/kubernetes/issues/132312

fieldpath/serialize.go
```diff
@@ -18,221 +18,228 @@ package fieldpath
...
-	if err != nil {
+	buf := bytes.Buffer{}
+	enc := jsontext.Encoder{}
+	enc.Reset(&buf)
+	if err := s.emitContentsV1(false, &enc); err != nil {
 		return err
 	}
-	stream.WriteObjectEnd()
-	return stream.Flush()
+	_, err := buf.WriteTo(w)
```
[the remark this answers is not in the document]
dims: Either we fix the test or we trim space here for the tests to pass. may be we should fix the test?

dims:
For the `pull-structured-merge-diff-test` failure, please add this to fix the error @inteon 

```
diff --git a/internal/cli/main_test.go b/internal/cli/main_test.go
index 3e409ede0673..8beed5c6cf55 100644
--- a/internal/cli/main_test.go
+++ b/internal/cli/main_test.go
@@ -21,6 +21,7 @@ import (
        "encoding/json"
        "io/ioutil"
        "path/filepath"
+       "strings"
        "testing"
 )

@@ -135,7 +136,7 @@ func (tt *testCase) checkOutput(t *testing.T, got []byte) {
                t.Fatalf("couldn't read expected output %q: %v", tt.expectedOutputPath, err)
        }

-       if a, e := string(got), string(want); a != e {
+       if a, e := strings.TrimSpace(string(got)), strings.TrimSpace(string(want)); a != e {
                t.Errorf("output didn't match expected output: got:\n%v\nwanted:\n%v\n", a, e)
        }
 }
```

liggitt:
/assign @jpbetz 
who is the primary apimachinery approver on this bit and was deeply involved in the initial performance-driven use of json-iterator in these bits

fieldpath/serialize-pe.go
```diff
@@ -127,60 +145,58 @@ func DeserializePathElement(s string) (PathElement, error) {
 	}
 }
 
-var (
-	readPool  = jsoniter.NewIterator(jsoniter.ConfigCompatibleWithStandardLibrary).Pool()
-	writePool = jsoniter.NewStream(jsoniter.ConfigCompatibleWithStandardLibrary, nil, 1024).Pool()
-)
+type PathElementSerializer struct {
+	buffer  bytes.Buffer
+	encoder jsontext.Encoder
```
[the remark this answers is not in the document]
inteon: I updated the PR to use `MarshalJSONTo`, which removed the need to do this manual pooling afaik.

fieldpath/serialize-pe.go
```diff
@@ -17,72 +17,101 @@ limitations under the License.
...
-// TODO: eliminate this https://github.com/kubernetes-sigs/structured-merge-diff/issues/202
-func readJSONIter(iter *jsoniter.Iterator) (value.Value, error) {
-	v := iter.Read()
-	if iter.Error != nil && iter.Error != io.EOF {
-		return nil, iter.Error
-	}
-	return value.NewValueInterface(v), nil
+// writeValueToEncoder writes a value to an Encoder.
+func writeValueToEncoder(v value.Value, enc *jsontext.Encoder) error {
+	return json.MarshalEncode(enc, v.Unstructured(), json.Deterministic(true))
```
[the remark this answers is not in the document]
inteon: Indeed, I found that by using `MarshalJSONTo`, we can use `json.MarshalWrite`.

liggitt:
Thanks for the updates, how are the overall benchmarks looking (not just the subset in the description)?

As you were adjusting the implementation, were there any unit tests it would make sense to add to catch edges the previous implementations handled we want to ensure the new one does as well? I'm thinking specifically of things like:
* handling of extra data in the input bytes/buffer when decoding/deserializing (e.g. `"somevalue"extrastuff` or `{"key":"value"}extrastuff`)
* handling of special characters in strings that need escaping where the raw bytes would not be the same as the decoded or encoded/escaped bytes
* handling of ignorable whitespace when decoding

inteon:
Did some further tuning and got the # allocations lower than on master.

BenTheElder:
> Golang releases a stable json/v2 (The alternative would be to add an internal copy of json-experiment to this repo like kube-openapi has, but I don't know if it's worth it given how close json/v2 is to stable

even with stable json/v2, we might need to temporarily use a fork, otherwise we kubernetes branches that aren't on that go version yet can't update SMD.

but we should encapsulate it and plan to eliminate it when we're ready to require that minimum go version

even kubernetes master isn't on 1.25 yet

liggitt:
> We just have to make sure this new library behaves the same as the existing implementation functionally, which I think existing tests should be able to do(?).

I'm not sure how detailed the existing tests are at all the edge cases of valid and invalid variants on input (handling of escaped values in keys, whitespace before/after/between tokens, valid and invalid syntax, etc), and byte-for-byte assertions about output. Since this is effectively rewriting some of the encoding/decoding paths, we need to make sure we have test coverage for those things.

lalitc375:
> > We just have to make sure this new library behaves the same as the existing implementation functionally, which I think existing tests should be able to do(?).
> 
> I'm not sure how detailed the existing tests are at all the edge cases of valid and invalid variants on input (handling of escaped values in keys, whitespace before/after/between tokens, valid and invalid syntax, etc), and byte-for-byte assertions about output. Since this needed to effectively rewrite some of the encoding/decoding paths, we need to make sure we have test coverage for those things.


There are not enough tests for unicode and escape characters . I have added those tests in #300.   Including these new tests, We should detect regression in Serialization and Deserialization code in future.

liggitt:
Excellent, #300 looks like a great step forward for test coverage of normalized encoding. We'll probably want similar additions for:
1. decoding of valid-but-non-normalized values working properly (insignificant leading / trailing / interspersed whitespace, or non-canonical escaped values, etc)
2. decoding of invalid values erroring properly

liggitt:
Looks like we can drop the "Performance is not yet great" comment from the description now... this looks ~neutral or an improvement for memory use at this point 🎉

liggitt:
> @inteon do we want to pick this up when k8s 1.36 cycle reopens

We want to pick it up as soon as possible, but we need to wait until json/v2 lands in stdlib in a non-experimental way. I'm not sure if that's happening in Go 1.26 or not (the status of https://github.com/golang/go/issues/76406 and https://github.com/orgs/golang/projects/50/views/1 makes me think not).

api-machinery folks have a detailed code review of this PR scheduled for later this week. Once we're happy with the shape of it, and do a final round of benchmarks, we'll submit this as positive community feedback on the current shape of json/v2, and then wait patiently for json/v2 to land in stdlib 🤞

value/fields.go
```diff
@@ -27,10 +32,109 @@ type Field struct {
...
+func (fl *FieldList) UnmarshalJSONFrom(parser *jsontext.Decoder) error {
+	if objStart, err := parser.ReadToken(); err != nil {
+		return fmt.Errorf("parsing JSON: %v", err)
+	} else if objStart.Kind() != jsontext.BeginObject.Kind() {
+		return fmt.Errorf("expected object")
+	}
+
+	var fields FieldList
+	for {
+		rawKey, err := parser.ReadToken()
```
liggitt: what tokens come back from this?

begin obj
[key value]*
end obj

are there other tokens we have to handle, like whitespace, comma, etc?

value/fields.go
```diff
@@ -27,10 +32,109 @@ type Field struct {
...
+		if err == io.EOF {
+			return fmt.Errorf("unexpected EOF")
+		} else if err != nil {
+			return fmt.Errorf("parsing JSON: %v", err)
+		}
+
```
liggitt: is it possible to encounter a token other than EndObject or String here? will ReadToken() error in that case? can we make sure there's a test that some other token type instead of a string key errors correctly?

value/fields.go
```diff
@@ -27,10 +32,109 @@ type Field struct {
 	Value Value
 }
 
+// Not meant to be used by an external library.
+type FastMarshalValue struct {
```
liggitt: curious why we need a wrapper type instead of adding MarshalJSONTo to Value directly?

fieldpath/serialize-pe.go
```diff
@@ -127,60 +94,58 @@ func DeserializePathElement(s string) (PathElement, error) {
...
-	readPool  = jsoniter.NewIterator(jsoniter.ConfigCompatibleWithStandardLibrary).Pool()
-	writePool = jsoniter.NewStream(jsoniter.ConfigCompatibleWithStandardLibrary, nil, 1024).Pool()
-)
-
 // SerializePathElement serializes a path element
 func SerializePathElement(pe PathElement) (string, error) {
-	buf := strings.Builder{}
-	err := serializePathElementToWriter(&buf, pe)
-	return buf.String(), err
+	serializer := pathElementSerializer{}
```
liggitt: is there a reason to construct this instance here instead of using one from the pool?

fieldpath/serialize-pe.go
```diff
@@ -127,60 +94,58 @@ func DeserializePathElement(s string) (PathElement, error) {
...
-func serializePathElementToWriter(w io.Writer, pe PathElement) error {
-	stream := writePool.BorrowStream(w)
-	defer writePool.ReturnStream(stream)
+type pathElementSerializer struct {
+	builder   bytes.Buffer
+	fastValue value.FastMarshalValue
+}
+
+func (pes *pathElementSerializer) reset() {
+	pes.builder.Reset()
```
liggitt: if the builder is "big", we want to nil it out instead of Resetting it and retaining a potentially very large buffer

make a constant, set it to something like 1kb, and if the builder is bigger than that, nil it out instead of calling reset

fieldpath/serialize-pe.go
```diff
@@ -127,60 +94,58 @@ func DeserializePathElement(s string) (PathElement, error) {
...
+		return "", err
+	}
+	return serializer.builder.String(), nil
 }
 
-func serializePathElementToWriter(w io.Writer, pe PathElement) error {
-	stream := writePool.BorrowStream(w)
-	defer writePool.ReturnStream(stream)
+type pathElementSerializer struct {
+	builder   bytes.Buffer
```
liggitt: since we ultimately want a string out of SerializePathElement, is it possible to use a strings.Builder here and get the same performance out of json/v2?

let's try each of these steps, benchmarking as we go:
1. switch to strings.Builder
2. see if we still need the pathElementSerializer pool and drop it if not

fieldpath/serialize.go
```diff
@@ -17,222 +17,230 @@ limitations under the License.
...
 		if c := mpe.Compare(cpe); c < 0 {
-			preWrite()
-			if err := serializePathElementToWriter(r.reset(), mpe); err != nil {
+			if err := writePathKey(om, mpe); err != nil {
 				return err
 			}
-			stream.WriteObjectField(r.unsafeString())
-			stream.WriteEmptyObject()
-			mi++
-		} else if c > 0 {
```
liggitt: I agree collapsing these two cases into a single block with the differences using `c == 0` should be equivalent, but it would be helpful to split that simplification out into its own commit / PR and merge ahead of the json-iterator -> json/v2 change

that would make it easier to see the diff in this PR just being the json-related changes

fieldpath/serialize.go
```diff
@@ -17,222 +17,230 @@ limitations under the License.
...
-	return iter.Error
+	return nil
 }
 
 // returns true if this subtree is also (or only) a member of parent; s is nil
 // if there are no further children.
-func readIterV1(iter *jsoniter.Iterator) (children *Set, isMember bool) {
-	iter.ReadMapCB(func(iter *jsoniter.Iterator, key string) bool {
-		if key == "." {
+func (s *setContentsV1) readIterV1(parser *jsontext.Decoder) (children *Set, isMember bool, err error) {
```
liggitt: the receiver isn't used, change this back to a function:

```suggestion
func readIterV1(parser *jsontext.Decoder) (children *Set, isMember bool, err error) {
```

liggitt:
Thanks for rebasing this.

Now that Go 1.27 is released and json/v2 is available, we're looking to pick this back up to merge and incorporate during the Kubernetes 1.38 cycle.

I think the requested test coverage items were split out and merged already. I'm sweeping through this to see what was remaining.

value/fields.go
```diff
@@ -27,10 +32,109 @@ type Field struct {
...
+		k := rawKey.String()
+
+		var v any
+		if err := json.UnmarshalDecode(parser, &v); err == io.EOF {
+			return fmt.Errorf("unexpected EOF")
+		} else if err != nil {
+			return fmt.Errorf("parsing JSON: %v", err)
+		}
+
+		fields = append(fields, Field{Name: k, Value: NewValueInterface(v)})
```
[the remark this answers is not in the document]
liggitt: test added in https://github.com/kubernetes-sigs/structured-merge-diff/pull/315, looks like current behavior accepts duplicate entries and accumulates them

fieldpath/serialize-pe.go
```diff
@@ -127,60 +94,58 @@ func DeserializePathElement(s string) (PathElement, error) {
...
+}
+
+func (pes *pathElementSerializer) serialize(pe PathElement) error {
 	switch {
 	case pe.FieldName != nil:
-		if _, err := stream.Write(peFieldSepBytes); err != nil {
+		if _, err := pes.builder.Write(peFieldSepBytes); err != nil {
+			return err
+		}
+		if _, err := pes.builder.WriteString(*pe.FieldName); err != nil {
```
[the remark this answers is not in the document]
liggitt: json-iterator code dumps to string without escaping - added test coverage in https://github.com/kubernetes-sigs/structured-merge-diff/pull/337

fieldpath/serialize-pe_test.go
```diff
@@ -159,6 +159,11 @@ func TestDeserializePathElementSuccess(t *testing.T) {
 		// Unescaped UTF-8 in key/value
 		{`k:{"name-🚀":"my-container"}`, KeyElement(value.Field{Name: "name-🚀", Value: value.NewValueInterface("my-container")})},
 		{`v:"value-🚀"`, ValueElement(value.NewValueInterface("value-🚀"))},
+
+		// A null key element is equivalent to an empty one.
```
liggitt, line 443: This test fails on master, code on master returns a non-nil zero-length list, this PR returns a nil list

Opened https://github.com/kubernetes-sigs/structured-merge-diff/pull/338 to add this test scenario to master in a way that passes, so we can make sure this PR mimics the behavior

value/fields.go
liggitt: Makes sense. If we have to export it for cross-package use, let's at least put it inside an `internal` package so  things outside the module can't use it

fieldpath/serialize.go
```diff
@@ -17,222 +17,230 @@ limitations under the License.
...
-		if childIsMember {
+
+		if isChildMember {
 			if children == nil {
 				children = &Set{}
 			}
+
+			// Append the member to the members list, we will sort it later
 			m := &children.Members.members
-			// Since we expect that most of the time these will have been
```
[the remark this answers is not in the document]
liggitt: Always appending and sorting at the end actually changed the behavior when duplicates are encountered

The unit test added in https://github.com/kubernetes-sigs/structured-merge-diff/pull/340 reveals this

Reverting to the previous logic here and below fixes that unit test

liggitt:
I pushed commits rebasing, resolving conflicts, and fixing current unit test failures to https://github.com/liggitt/structured-merge-diff/commits/use_json_v2-20260825-liggitt/ if you want to pull those in and squash. I'm still working through the remaining open comments to see what is still relevant.

edit: since you marked the PR as editable by maintainers, I went ahead and pushed my rebased branch with my commits, so this could exercise CI ... I hope you don't mind. It looks like the revert to current Set behavior spiked the benchmarks, so before squashing, let's see if we can keep compatible behavior while keeping benchmarks looking better

fieldpath/serialize-pe_test.go
liggitt: resolved in https://github.com/liggitt/structured-merge-diff/commits/use_json_v2-20260825-liggitt/

fieldpath/serialize.go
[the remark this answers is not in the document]
liggitt: resolved in https://github.com/liggitt/structured-merge-diff/commits/use_json_v2-20260825-liggitt/

fieldpath/serialize.go
liggitt: resolved in https://github.com/liggitt/structured-merge-diff/commits/use_json_v2-20260825-liggitt/

fieldpath/serialize.go
```diff
@@ -17,222 +17,230 @@ limitations under the License.
...
+
 	if children == nil {
 		isMember = true
 	}
 
-	return children, isMember
+	return children, isMember, nil
+}
+
+// FromJSON clears s and reads a JSON formatted set structure.
```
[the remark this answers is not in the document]
liggitt: resolved in https://github.com/liggitt/structured-merge-diff/commits/use_json_v2-20260825-liggitt/

fieldpath/serialize.go
```diff
@@ -17,222 +17,230 @@ limitations under the License.
...
-			} else {
-				*children.Children.Descend(pe) = *grandchildren
-			}
+			*m = append(*m, setNode{pe, grandChildren})
 		}
-		return true
-	})
+	}
+
+	// Sort the members and children
```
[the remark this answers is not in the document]
liggitt: resolved in https://github.com/liggitt/structured-merge-diff/commits/use_json_v2-20260825-liggitt/

fieldpath/serialize.go
```diff
@@ -17,222 +17,230 @@ limitations under the License.
...
-			}
+			*m = append(*m, pe)
 		}
-		if grandchildren != nil {
+
+		if grandChildren != nil {
 			if children == nil {
 				children = &Set{}
 			}
-			// Since we expect that most of the time these will have been
```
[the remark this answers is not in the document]
liggitt: resolved in https://github.com/liggitt/structured-merge-diff/commits/use_json_v2-20260825-liggitt/

value/fields.go
liggitt: covered by unit tests now

value/fields.go
```diff
@@ -27,10 +32,109 @@ type Field struct {
...
+		}
+
+		if rawKey.Kind() == jsontext.EndObject.Kind() {
+			break
+		}
+
+		k := rawKey.String()
+
+		var v any
+		if err := json.UnmarshalDecode(parser, &v); err == io.EOF {
```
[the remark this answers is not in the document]
liggitt: passes with unit tests in master

liggitt:
rebased on master after making the benchmark deterministic in https://github.com/kubernetes-sigs/structured-merge-diff/pull/344

Added a benchmark of the value.ToJSON/FromJSON functions, which we were missing, and recaptured at each layer with this:

> `go test -benchmem -run=^$ -bench '^(BenchmarkFieldSet|BenchmarkJSON)/.*serialize.*$' sigs.k8s.io/structured-merge-diff/v6/fieldpath sigs.k8s.io/structured-merge-diff/v6/value -count=6`

Found a way to trigger fastpath decodes in json/v2 into `any` objects assuming we won't have duplicate keys, and only if we do falling back to AllowDuplicateNames which is slower.

With that change, benchmark looks like this:

serialize is consistently 60-70% better on bytes, inconsistently better/worse on allocs:
```
                             │ 1-master.txt │          5-pathelement.txt          │
                             │    sec/op    │    sec/op     vs base               │
FieldSet/serialize-20-14        3.535µ ± 0%    5.751µ ± 1%  +62.67% (p=0.002 n=6)
FieldSet/serialize-50-14        9.899µ ± 1%   17.154µ ± 4%  +73.30% (p=0.002 n=6)
FieldSet/serialize-100-14       32.47µ ± 1%    55.85µ ± 3%  +72.01% (p=0.002 n=6)
FieldSet/serialize-500-14       174.6µ ± 0%    300.1µ ± 1%  +71.91% (p=0.002 n=6)
FieldSet/serialize-1000-14      386.6µ ± 3%    658.0µ ± 1%  +70.19% (p=0.002 n=6)

                             │ 1-master.txt  │          5-pathelement.txt          │
                             │     B/op      │     B/op      vs base               │
FieldSet/serialize-20-14         2411.5 ± 0%     954.0 ± 0%  -60.44% (p=0.002 n=6)
FieldSet/serialize-50-14        7.193Ki ± 0%   2.688Ki ± 0%  -62.63% (p=0.002 n=6)
FieldSet/serialize-100-14      20.577Ki ± 0%   8.802Ki ± 0%  -57.23% (p=0.002 n=6)
FieldSet/serialize-500-14      111.84Ki ± 0%   44.78Ki ± 0%  -59.96% (p=0.002 n=6)
FieldSet/serialize-1000-14     244.43Ki ± 1%   97.03Ki ± 0%  -60.30% (p=0.002 n=6)

                             │ 1-master.txt │         5-pathelement.txt          │
                             │  allocs/op   │  allocs/op   vs base               │
FieldSet/serialize-20-14         8.000 ± 0%    4.000 ± 0%  -50.00% (p=0.002 n=6)
FieldSet/serialize-50-14        13.000 ± 0%    9.000 ± 0%  -30.77% (p=0.002 n=6)
FieldSet/serialize-100-14        27.00 ± 0%    30.00 ± 0%  +11.11% (p=0.002 n=6)
FieldSet/serialize-500-14        120.0 ± 0%    151.0 ± 1%  +25.83% (p=0.002 n=6)
FieldSet/serialize-1000-14       259.0 ± 0%    337.0 ± 0%  +30.12% (p=0.002 n=6)

pkg: sigs.k8s.io/structured-merge-diff/v6/value
                    │ 1-master.txt │         5-pathelement.txt          │
                    │    sec/op    │   sec/op     vs base               │
JSON/serialize-14      3.991µ ± 3%   4.511µ ± 0%  +13.02% (p=0.002 n=6)

                    │ 1-master.txt │          5-pathelement.txt          │
                    │     B/op     │     B/op      vs base               │
JSON/serialize-14     7.079Ki ± 0%   1.109Ki ± 0%  -84.33% (p=0.002 n=6)

                    │ 1-master.txt │         5-pathelement.txt         │
                    │  allocs/op   │ allocs/op   vs base               │
JSON/serialize-14      148.00 ± 0%   94.00 ± 0%  -36.49% (p=0.002 n=6)
```

deserialize is inconsistently better/worse on bytes, consistently ~4% worse on allocs:
```
                             │ 1-master.txt │          5-pathelement.txt          │
                             │    sec/op    │    sec/op     vs base               │
FieldSet/deserialize-20-14      7.014µ ± 1%    9.585µ ± 3%  +36.66% (p=0.002 n=6)
FieldSet/deserialize-50-14      17.65µ ± 3%    26.72µ ± 1%  +51.35% (p=0.002 n=6)
FieldSet/deserialize-100-14     56.99µ ± 1%    88.10µ ± 3%  +54.60% (p=0.002 n=6)
FieldSet/deserialize-500-14     291.2µ ± 0%    459.2µ ± 0%  +57.68% (p=0.002 n=6)
FieldSet/deserialize-1000-14    652.6µ ± 1%   1018.0µ ± 1%  +55.98% (p=0.002 n=6)

                             │ 1-master.txt  │          5-pathelement.txt          │
                             │     B/op      │     B/op      vs base               │
FieldSet/deserialize-20-14     11.144Ki ± 0%   7.563Ki ± 0%  -32.13% (p=0.002 n=6)
FieldSet/deserialize-50-14      23.60Ki ± 0%   21.04Ki ± 0%  -10.85% (p=0.002 n=6)
FieldSet/deserialize-100-14     71.69Ki ± 0%   72.44Ki ± 0%   +1.06% (p=0.002 n=6)
FieldSet/deserialize-500-14     348.8Ki ± 0%   371.0Ki ± 0%   +6.38% (p=0.002 n=6)
FieldSet/deserialize-1000-14    761.6Ki ± 0%   817.6Ki ± 0%   +7.35% (p=0.002 n=6)

                             │ 1-master.txt │         5-pathelement.txt          │
                             │  allocs/op   │  allocs/op   vs base               │
FieldSet/deserialize-20-14       266.0 ± 0%    275.0 ± 0%   +3.38% (p=0.002 n=6)
FieldSet/deserialize-50-14       777.0 ± 0%    808.0 ± 0%   +3.99% (p=0.002 n=6)
FieldSet/deserialize-100-14     2.589k ± 0%   2.689k ± 0%   +3.88% (p=0.002 n=6)
FieldSet/deserialize-500-14     13.27k ± 0%   13.82k ± 0%   +4.13% (p=0.002 n=6)
FieldSet/deserialize-1000-14    29.41k ± 0%   30.61k ± 0%   +4.09% (p=0.002 n=6)

pkg: sigs.k8s.io/structured-merge-diff/v6/value
                    │ 1-master.txt │         5-pathelement.txt          │
                    │    sec/op    │   sec/op     vs base               │
JSON/deserialize-14    2.561µ ± 3%   4.636µ ± 1%  +81.02% (p=0.002 n=6)

                    │ 1-master.txt │          5-pathelement.txt          │
                    │     B/op     │     B/op      vs base               │
JSON/deserialize-14   5.308Ki ± 0%   5.639Ki ± 0%   +6.24% (p=0.002 n=6)

                    │ 1-master.txt │         5-pathelement.txt         │
                    │  allocs/op   │ allocs/op   vs base               │
JSON/deserialize-14     138.0 ± 0%   156.0 ± 0%  +13.04% (p=0.002 n=6)
```

value/fields.go
```diff
@@ -27,10 +32,109 @@ type Field struct {
...
 // have a different name.
 type FieldList []Field
 
+var _ json.MarshalerTo = (*FieldList)(nil)
+var _ json.UnmarshalerFrom = (*FieldList)(nil)
+
+func (fl *FieldList) MarshalJSONTo(enc *jsontext.Encoder) error {
+	enc.WriteToken(jsontext.BeginObject)
+	for _, f := range *fl {
+		if err := enc.WriteToken(jsontext.String(f.Name)); err != nil {
```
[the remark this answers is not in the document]
liggitt: added escaping unit tests to master, resolving

fieldpath/serialize-pe.go
liggitt: benchmarking showed no improvement with the pool, resolving

edit: oh... nothing calls this function from a benchmark... added a benchmark and pooling definitely helps :)

fieldpath/serialize-pe.go
liggitt: benchmarked with switch to strings.Builder, it made allocs worse on serialize

benchmarked without the pool, it made allocs worse on serialize

resolving

fieldpath/serialize-pe.go
liggitt: pushed a commit to do this

value/fields.go
```diff
@@ -27,10 +32,109 @@ type Field struct {
...
+		list := v.AsList()
+		for i := 0; i < list.Length(); i++ {
+			if err := valueMarshalJSONTo(enc, list.At(i)); err != nil {
+				return err
+			}
+		}
+		return enc.WriteToken(jsontext.EndArray)
+	case v.IsMap():
+		// use the json marshaller to make sure the key ordering is deterministic
+		fallthrough
```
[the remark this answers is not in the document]
liggitt: fixed in latest push

value/fields.go
```diff
@@ -27,10 +32,109 @@ type Field struct {
...
+	default:
+		return json.MarshalEncode(enc, v.Unstructured(), json.Deterministic(true))
```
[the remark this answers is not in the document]
liggitt: re-read this more thoroughly, and Unstructured() recursively converts the remaining tree to Unstructured()

the code in master always converted to Unstructured(), so this fast-path optimizing a single-level non-map is strictly better

value/fields.go
```diff
@@ -27,10 +32,109 @@ type Field struct {
...
+func valueMarshalJSONTo(enc *jsontext.Encoder, v Value) error {
+	switch {
+	case v.IsNull():
+		return enc.WriteToken(jsontext.Null)
+	case v.IsFloat():
+		return enc.WriteToken(jsontext.Float(v.AsFloat()))
+	case v.IsInt():
+		return enc.WriteToken(jsontext.Int(v.AsInt()))
+	case v.IsString():
+		return enc.WriteToken(jsontext.String(v.AsString()))
```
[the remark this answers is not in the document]
liggitt: benchmarked making this a value switch and there was no performance impact, don't worry about this

value/fields.go
liggitt: relocated to be an unexported sibling type to pathElementSerializer in latest commit, resolving

fieldpath/serialize-pe.go
```diff
@@ -127,60 +94,58 @@ func DeserializePathElement(s string) (PathElement, error) {
...
+		if err := json.MarshalWrite(&pes.builder, pe.Key, json.Deterministic(true)); err != nil {
+			return err
 		}
-		stream.WriteObjectEnd()
 	case pe.Value != nil:
-		if _, err := stream.Write(peValueSepBytes); err != nil {
+		if _, err := pes.builder.Write(peValueSepBytes); err != nil {
+			return err
+		}
+		pes.fastValue.Value = pe.Value
```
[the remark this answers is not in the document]
liggitt: I played with this for a while and couldn't beat the allocs of making this a field on the pathElementSerializer, all the custom implementation versions performed worse, resolving

value/fields.go
```diff
@@ -27,10 +32,109 @@ type Field struct {
...
+}
+
+var _ json.MarshalerTo = FastMarshalValue{}
+
+func (mv FastMarshalValue) MarshalJSONTo(enc *jsontext.Encoder) error {
+	return valueMarshalJSONTo(enc, *mv.Value)
+}
+
+func valueMarshalJSONTo(enc *jsontext.Encoder, v Value) error {
+	switch {
```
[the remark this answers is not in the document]
liggitt: reading more carefully, IsInt() does *not* return true for uint64, so I don't think we have any new risk there

fieldpath/serialize.go
liggitt: pushed in `minimize diff` commit

liggitt:
All my comments are addressed, and the performance looks good enough in my opinion to merge. Will let @inteon, @jpbetz, and @cici37 take a look at the stacked commits before squashing.

Before merge, I'd like to get kubernetes/kubernetes updated to Go 1.27 (https://github.com/kubernetes/kubernetes/pull/141663), and get a green run of this in a POC PR (https://github.com/kubernetes/kubernetes/pull/141664).

We also might want to consider making this a v7 bump of structured-merge-diff, since it's bumping the required Go version and significantly changing the internal impl

inteon:
> All my comments are addressed, and the performance looks good enough in my opinion to merge. Will let @inteon, @jpbetz, and @cici37 take a look at the stacked commits before squashing.
> 
> Before merge, I'd like to get kubernetes/kubernetes updated to Go 1.27 ([kubernetes/kubernetes#141663](https://github.com/kubernetes/kubernetes/pull/141663)), and get a green run of this in a POC PR ([kubernetes/kubernetes#141664](https://github.com/kubernetes/kubernetes/pull/141664)).
> 
> We also might want to consider making this a v7 bump of structured-merge-diff, since it's bumping the required Go version and significantly changing the internal impl

Thank you for finishing this PR, the changes you added look good to me.

value/fields.go
```diff
@@ -27,10 +33,110 @@ type Field struct {
...
+		}
+		list := v.AsList()
+		for i := 0; i < list.Length(); i++ {
+			if err := ValueMarshalJSONTo(enc, list.At(i)); err != nil {
+				return err
+			}
+		}
+		return enc.WriteToken(jsontext.EndArray)
+	case v.IsMap():
+		return json.MarshalEncode(enc, v.Unstructured())
```
jpbetz: I think this relies on the caller's encoder already carrying Deterministic(true) (MarshalEncode only joins opts when non-empty). The in-module paths inherit it from the MarshalWrite calls in serialize-pe.go, but ValueMarshalJSONTo and FieldList.MarshalJSONTo are exported, so a caller with a plain encoder — or just json.Marshal(&fieldList) — gets map values in random key order (~20 distinct orderings over 300 runs when I probed it). 328638f dropped the explicit option that used to be on this line; can we re-add it?

```suggestion
		return json.MarshalEncode(enc, v.Unstructured(), json.Deterministic(true))
```

(verified locally, suite passes)

value/fields.go
```diff
@@ -27,10 +33,110 @@ type Field struct {
...
+	default:
+		return fmt.Errorf("cannot marshal unknown value type to json")
```
jpbetz: Follow-up to the resolved IsInt/uint64 thread: this error case changes behavior for values jsoniter used to reflect-marshal — NewValueInterface(uint64(7)), structs, []byte, typed slices/maps now error here, while value.ToJSON on the same Value still succeeds via Unstructured() reflection, so the two encoders disagree about what's marshalable. Also, `uint` values > MaxInt64 aren't caught by this case: IsInt matches uint and AsInt overflows, so they silently serialize as -1 (master wrote the full value) — the "no overflow risk" conclusion holds for uint64 but not uint. None of this is reachable from Kubernetes (typed validation rejects uint64 identically on both trees before serialization), so probably fine as-is — but maybe include %T in the error, and worth a line in the squash message.

value/fields.go
```diff
@@ -27,10 +33,110 @@ type Field struct {
 	Value Value
 }
 
+func ValueMarshalJSONTo(enc *jsontext.Encoder, v Value) error {
```
jpbetz: I checked whether this could move to an internal package per the earlier thread — I don't think it can: FieldList.MarshalJSONTo below needs it, and any internal home would have to import value for the Value type, which is a cycle (I tried the refactor; the build fails). Since it's the only discretionary new export in the module, could we give it a godoc marking it for intra-module use, in the spirit of the old "DO NOT EXPORT" markers on readJSONIter/writeJSONStream? While in here: the comment on UnmarshalJSONFrom below still says "FieldListFromJSON" (stale name), and MarshalJSONTo has no comment.

value/fields.go
```diff
@@ -27,10 +33,110 @@ type Field struct {
...
+}
+
 // FieldList is a list of key-value pairs. Each field is expected to
 // have a different name.
 type FieldList []Field
 
+var _ json.MarshalerTo = (*FieldList)(nil)
+var _ json.UnmarshalerFrom = (*FieldList)(nil)
+
+func (fl *FieldList) MarshalJSONTo(enc *jsontext.Encoder) error {
```
jpbetz: Side effect of putting these methods on the exported FieldList: on go1.27, encoding/json v1 honors the v2 interfaces when the value is addressable, so plain json.Marshal/Unmarshal of a FieldList changes shape vs v6.4. Probed both trees: Marshal(&fl) flips from the reflected [{"Name":...,"Value":...}] array to {"a":1,...} (non-addressable fl still reflects, so output now depends on addressability), a nil *FieldList flips null → {}, and Unmarshal inverts (object form used to error and now works; the old array form now errors). GODEBUG=jsonv2=0 doesn't revert it. Nothing in k/k round-trips FieldList through v1 json today (only use builds it field-by-field), so I think this is acceptable — but worth a release-note mention, or if we'd rather not touch the public type's contract, the methods could move to an unexported wrapper in fieldpath like fastMarshalValue did.

fieldpath/serialize-pe.go
```diff
@@ -143,60 +109,77 @@ func DeserializePathElement(s string) (PathElement, error) {
...
+		if err := json.MarshalWrite(&pes.builder, &pes.fastValue, json.Deterministic(true)); err != nil {
```
jpbetz: One escaping difference I don't think any thread covered: jsoniter's stdlib-compat config HTML/JS-escaped string *values* (< > & → \u003c etc., plus U+2028/U+2029) in v: payloads, k: field values, and value.ToJSON; jsontext doesn't. So stored keys like `v:"https://example.com/?a=1\u0026b=2"` re-encode with a raw & — one-time byte churn, plus flip-flop while HA apiservers are version-skewed (decode treats both forms as equal path elements, so churn rather than corruption). EscapeForHTML(true)+EscapeForJS(true) on the marshal calls reproduces jsoniter's value bytes exactly, but it's position-blind: it would also escape k: member *names* jsoniter left raw. Escaped chars in values (URLs with &) seem much more likely in real data than <>& in merge-key field names, so the options look like the closer match to me — but it's a judgment call. Whichever way, could we add a round-trip case with </&/U+2028 to pin it? Nothing in the suite covers these today.

fieldpath/serialize.go
```diff
@@ -136,85 +116,114 @@ func (s *Set) emitContentsV1(includeSelf bool, stream *jsoniter.Stream, r *reusa
...
 		} else if err != nil {
-			iter.ReportError("parsing key as path element", err.Error())
-			iter.Skip()
-			return true
+			return nil, false, fmt.Errorf("parsing key as path element: %v", err)
+		}
+
+		grandchildren, childIsMember, err := readIterV1(parser)
+		if err != nil {
+			return nil, false, fmt.Errorf("parsing value as set: %v", err)
```
jpbetz: I think this wrap fires once per recursion level with constant text, so a max-depth input (jsontext caps at 10k) comes back as a ~220KB error containing 10,000 copies of "parsing value as set: ", built in ~0.7s of quadratic string copying — master returned ~200B in a few ms for the same input. The jsontext error already carries a JSON pointer + offset, so the wraps add nothing; maybe just `return nil, false, err` here, or wrap once up in UnmarshalJSONFrom. Only reachable on malformed/hostile input, so follow-up-friendly. (Related, no action needed: jsontext now applies the same 10k cap on encode, which seems strictly better — master happily wrote sets its own decoder could never re-read, and output is byte-identical below the cap.)

fieldpath/serialize.go
```diff
@@ -136,85 +116,114 @@ func (s *Set) emitContentsV1(includeSelf bool, stream *jsoniter.Stream, r *reusa
...
+		// A null is equivalent to an empty object: it contributes no
+		// children, and is a member of its parent (if any).
+		return nil, true, nil
+	default:
+		return nil, false, fmt.Errorf("expected object")
+	}
+
+	for {
+		rawKey, err := parser.ReadToken()
+		if err == io.EOF {
```
jpbetz: nit: I don't think this branch (or the two like it at value/fields.go:112,125) is reachable — jsontext returns bare io.EOF only at top level before any token, and we're already inside the object here, so truncation arrives as a syntactic error wrapping io.ErrUnexpectedEOF and lands in the branch below. Swapping all three for panics survives the full suite plus truncation probes at every byte prefix. Fine to drop, or use errors.Is(err, io.ErrUnexpectedEOF) if we want the friendlier message — fine as a follow-up either way.

fieldpath/serialize.go
```diff
@@ -136,85 +116,114 @@ func (s *Set) emitContentsV1(includeSelf bool, stream *jsoniter.Stream, r *reusa
...
+		*(*Set)(s) = *found
 	}
-	if iter.Error == io.EOF {
-		return nil
-	}
-	return iter.Error
+	return nil
+}
+
+// FromJSON clears s and reads a JSON formatted set structure.
```
jpbetz: nit: on a parse error *s is now left untouched (jsoniter overwrote it with partial/empty state before reporting the error; the trailing-data case still overwrites in both versions since the value parse succeeds first). Every caller uses a fresh Set and checks the error, so the new behavior seems strictly better — but "clears s" here is now only true on success; maybe reword.

internal/json/json.go
```diff
@@ -0,0 +1,43 @@
+package json
```
jpbetz: nit: missing the Apache license header (every other .go file has one), and a package doc line would help. The comment on allowInvalidUTF8 still calls itself "decodeOptions" from an earlier revision. Also fieldpath declares its own copies of allowInvalidUTF8/allowDuplicates — maybe worth a single home so they can't drift. Fine as a pre-squash tidy-up.

value/fields.go
```diff
@@ -27,10 +33,110 @@ type Field struct {
 	Value Value
 }
 
+func ValueMarshalJSONTo(enc *jsontext.Encoder, v Value) error {
+	switch {
+	case v.IsNull():
+		return enc.WriteToken(jsontext.Null)
+	case v.IsFloat():
+		return enc.WriteToken(jsontext.Float(v.AsFloat()))
```
jpbetz: jsontext.Float encodes NaN/±Inf as the JSON strings "NaN"/"Infinity" instead of failing, so SerializePathElement now succeeds where it previously errored, and the value round-trips as a string rather than a float. Suggest rejecting non-finite floats to keep the previous behavior:

```suggestion
	case v.IsFloat():
		f := v.AsFloat()
		if math.IsNaN(f) || math.IsInf(f, 0) {
			return fmt.Errorf("unsupported value: %v", f)
		}
		return enc.WriteToken(jsontext.Float(f))
```

Requires a `math` import. Verified locally; full suite passes.

value/fields.go
jpbetz: Behavior change: value types jsoniter reflect-marshaled (uint64, structs, []byte, typed slices/maps) now error here, while value.ToJSON still marshals them — the two encoders disagree. Separately, uint values > MaxInt64 pass IsInt and silently serialize as -1. Neither is reachable from Kubernetes objects (typed validation rejects them first). Suggest including %T in the error message and noting the change in the squash message.

value/fields.go
jpbetz: This can't move to an internal package: the helper needs the Value type, so an internal home would create an import cycle with value (verified). Since it must stay exported, suggest a doc comment marking it for intra-module use only, like the old "DO NOT EXPORT" markers. Related: the UnmarshalJSONFrom comment below refers to a nonexistent "FieldListFromJSON", and MarshalJSONTo has no comment.

value/fields.go
jpbetz: On Go 1.27, encoding/json v1 honors these v2 methods, so json.Marshal/Unmarshal of an addressable FieldList changes shape: marshal flips from an array of {Name, Value} objects to a plain JSON object, a nil *FieldList becomes {} instead of null, and unmarshal now requires the object form. No k/k code depends on the old shape. Suggest a release-note mention — or, if we'd rather not change the public type's v1 behavior, move the methods to an unexported wrapper in fieldpath like fastMarshalValue.

fieldpath/serialize.go
jpbetz: This wraps the error once per recursion level, so a depth-capped input (jsontext limit: 10k) produces a ~220KB error string built by quadratic copying. The jsontext error already includes a JSON pointer and byte offset. Suggest returning err unwrapped here, or wrapping once in UnmarshalJSONFrom. Only affects malformed input; fine as a follow-up.

fieldpath/serialize.go
jpbetz: nit: this branch (and the two like it at value/fields.go:112,125) appears unreachable — inside an object, jsontext reports truncation as a wrapped io.ErrUnexpectedEOF, never bare io.EOF. Suggest dropping them, or matching with errors.Is(err, io.ErrUnexpectedEOF) if the friendlier message is wanted. Fine as a follow-up.

value/value_test.go
```diff
@@ -181,9 +181,8 @@ func fromJSONCases(includeDuplicateCases bool) []fromJSONCase {
 		{"escapedLatin", `{"\u00c5":1}`, `{"Å":1}`},
 		{"escapedMultiBytePair", `{"\uD83D\uDE02":1}`, `{"😂":1}`},
 
-		// TODO: When we migrate to json/v2 we should use jsontext.AllowInvalidUTF8
-		// and change this test case to expect the bytes to be munged to U+FFFD.
-		{"byteOrderMark", "{\"\xff\xfe\": 1}", "{\"\xff\xfe\": 1}"},
+		// Invalid UTF-8 is tolerated; the offending bytes are munged to U+FFFD.
+		{"byteOrderMark", "{\"\xff\xfe\": 1}", "{\"��\": 1}"},
```
jpbetz: Suggest also pinning the two remaining byte-format changes: floats with exponent -7..-9 serialize as `1e-7` instead of jsoniter's `1e-07` (matches stdlib v1; no option restores the padding), and U+0008/U+000C escape as `\b`/`\f` instead of `\u0008`/`\u000c`. Both are decode-compatible in either direction; a round-trip case for each would make the new forms the documented behavior.

value/fields.go
jpbetz:
```suggestion
	case v.IsFloat():
		f := v.AsFloat()
		if math.IsNaN(f) || math.IsInf(f, 0) {
			return fmt.Errorf("unsupported value: %v", f)
		}
		return enc.WriteToken(jsontext.Float(f))
```

to preserve NaN handling?

value/fields.go
jpbetz:
```suggestion
		return json.MarshalEncode(enc, v.Unstructured(), json.Deterministic(true))
```

I think?

value/fields.go
jpbetz: These methods change how plain encoding/json handles the exported FieldList on Go 1.27: marshal output flips from an array of {Name, Value} objects to a plain JSON object, and unmarshal now expects the new shape. Nothing in k/k depends on the old shape, but it is a public behavior change — suggest a release note, or moving the methods to an unexported wrapper in fieldpath.

fieldpath/serialize-pe.go
```diff
@@ -143,60 +109,77 @@ func DeserializePathElement(s string) (PathElement, error) {
...
-		stream.WriteObjectStart()
-
-		for i, field := range *pe.Key {
-			if i > 0 {
-				stream.WriteMore()
-			}
-			stream.WriteObjectField(field.Name)
-			writeJSONStream(field.Value, stream)
+		// preserve json-iterator behavior of tolerating duplicates (duplicates output multiple times)
+		if err := json.MarshalWrite(&pes.builder, pe.Key, json.Deterministic(true), allowDuplicates); err != nil {
```
jpbetz: Shouldn't we retain allowInvalidUTF8 on all encodes? If we don't I think we have a round-trip problem.

Looks like SerializePathElement accepts an f: name with invalid bytes, Set.ToJSON rejects the same element.

fieldpath/serialize-pe.go
jpbetz: Do we care that jsoniter wrote HTML-escaped, json/v2 writes raw?  Any & < > would be impacted.

fieldpath/serialize.go
jpbetz: Deeply nested invalid input produces enormous errors here — the message grows by one "parsing value as set:" per recursion level (~220KB at the depth cap). The decoder's error already says where parsing failed; suggest returning err unwrapped, or wrapping once at the top. Fine as a follow-up.

jpbetz (commented):
Thanks!  A couple edges I noticed then LGTM

value/fields.go
liggitt: added test coverage in https://github.com/kubernetes-sigs/structured-merge-diff/pull/348, added a fix commit in this PR

value/fields.go
liggitt: added test coverage in https://github.com/kubernetes-sigs/structured-merge-diff/pull/348, added a fix commit in this PR

fieldpath/serialize-pe.go
liggitt: added test coverage in https://github.com/kubernetes-sigs/structured-merge-diff/pull/348, added a fix commit in this PR

fieldpath/serialize-pe.go
liggitt: hrmm ... added test coverage in https://github.com/kubernetes-sigs/structured-merge-diff/pull/348 to reveal the behavior change at least

our custom json-iterator impl was already inconsistent with where we would escape (we wouldn't escape in `f:...` or in `k:{"..."` but we would escape elsewhere)

I played with replicating the behavior with a `jsontext.EscapeForHTML(true)` option and it exploded allocs/bytes (like 1000% worse). I tend to think that is not worth the cost, since those decode to the same thing internally in memory.

liggitt:
Ok, I think the only remaining items are to:

- [x] land the go 1.27 update in k/k (https://github.com/kubernetes/kubernetes/pull/141663)
- [x] get the dry-run PR green (I updated https://github.com/kubernetes/kubernetes/pull/141664)
- [x] specifically review and ack test changes in this PR (there are very few):
   * non-utf characters are consistently replaced, which is good
   * the replacement characters are no longer escaped, which I think is ~fine
   * html characters are no longer escaped, which I think is likely ~fine
- [ ] decide if we're going to increment this library to v7 (I think we probably should), and the order of operations to do that (branch v6 from master, merge this PR, then merge a PR incrementing the module name to v7, then tag v7.0.0?)

jpbetz:
> * [ ]  specifically review and ack test changes in this PR (there are very few):
>   
>   * non-utf characters are consistently replaced, which is good
>   * the replacement characters are no longer escaped, which I think is ~fine
>   * html characters are no longer escaped, which I think is likely ~fine

I look through this. It all LGTM.

> * [ ]  decide if we're going to increment this library to v7 (I think we probably should), and the order of operations to do that (branch v6 from master, merge this PR, then merge a PR incrementing the module name to v7, then tag v7.0.0?)

Yes. Once this merges, I'll perform the bump. This repo never got any automation, so I'll do it manually.

liggitt:
anything we fix in the v6 branch at this point would likely be high-sev bugfixes in the v6.4.x stream

I created https://github.com/kubernetes-sigs/structured-merge-diff/tree/release-6.4 from master prior to merging this

liggitt:
v7 bump PR at https://github.com/kubernetes-sigs/structured-merge-diff/pull/349

PR #298, opened 2025-07-16
Add associative list key schema evolution changes

jpbetz:
Builds on https://github.com/kubernetes-sigs/structured-merge-diff/pull/296
Add test coverage for:

- Changing a listType=map key to include a new optional key
- Changing a listType=map key to include a new required + defaulted key

### DIFF
--- a/merge/schema_change_test.go
+++ b/merge/schema_change_test.go
@@ -274,9 +274,6 @@ var associativeListParserOld = func() *typed.Parser {
     - name: name
       type:
         scalar: string
-    - name: id
-      type:
-        scalar: numeric
     - name: value
       type:
         scalar: numeric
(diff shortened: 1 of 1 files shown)


merge/schema_change_test.go
```diff
@@ -252,3 +252,459 @@ func TestAtomicToGranularSchemaChanges(t *testing.T) {
...
+          namedType: associativeList
+- name: associativeList
+  list:
+    elementType:
+      namedType: myElement
+    elementRelationship: associative
+    keys:
+    - name
+- name: myElement
+  map:
```
liggitt, line 272: do we have to include id here for old data to persist, even if we didn't make it a list key?

jpbetz: I split the tests into two groups. The ones here in `TestAssociativeListSchemaChanges` simulate adding a new key field.  The ones in below in `TestPromoteFieldToAssociativeListKey` simulate promoting an existing field to a key.

liggitt: ah, ok, I thought this schema was used as the starting point for both tests

PR #303, opened 2025-09-16
Add more test coverage for searialization and deserialization

lalitc375:
Add cases mentioned in  https://github.com/kubernetes-sigs/structured-merge-diff/pull/292#issuecomment-3271796972.

This pull request adds test coverage for serialization and deserialization of field paths, focusing on correctly handling spaces and invalid escape characters.

### Description

This PR addresses cases mentioned in https://github.com/kubernetes-sigs/structured-merge-diff/pull/292#issuecomment-3271796972. The test suite is expanded to include tests for:

*   **Leading and trailing spaces:** New test cases have been added to `fieldpath/serialize-pe_test.go`, `fieldpath/serialize_test.go`, and `fieldpath/set_test.go` to ensure that leading and trailing spaces in field names and key values are preserved during serialization and deserialization round-trips.
*   **Invalid escape characters:** A new test has been added to `fieldpath/serialize-pe_test.go` to verify that an error is correctly returned when deserializing a path element that contains an invalid JSON escape sequence.

These additions improve the robustness of the serialization and deserialization  logic.

### DIFF
--- a/fieldpath/serialize-pe_test.go
+++ b/fieldpath/serialize-pe_test.go
@@ -16,7 +16,9 @@ limitations under the License.
 
 package fieldpath
 
-import "testing"
+import (
+	"testing"
+)
 
 func TestPathElementRoundTrip(t *testing.T) {
 	tests := []string{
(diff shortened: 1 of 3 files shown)


liggitt:
> In addition to this, I was hoping to merge some tests where the order of map keys and the spacing between delimiters (not just within the contents of strings) is not the same was what the serializer typically outputs. This would tease out issues where a streaming deserializer is accidentally sensitive to the order of the token stream or the whitespace between tokens.

Big +1 to this ... this PR is a great addition, we also want to exercise correct consumption of non-normalized outputs with non-semantic whitespace before / between / after tokens, and make sure the normalized output strips that out.

PR #306, opened 2025-10-03
fix: RemoveItems should preserve empty list and map

### LINKED ISSUE #305: RemoveItems converts empty maps and lists ({} and []) instead to null
I'm experiencing a similar issue as https://github.com/kubernetes-sigs/structured-merge-diff/pull/302 except for `RemoveItems()`

For example in my CronJob jobTempalte i end up with a metadata: null which is not a nullable field

```
  "spec": {
    "concurrencyPolicy": "Allow",
    "failedJobsHistoryLimit": 1,
    "jobTemplate": {
      "metadata": null,
```
We should do the same as Extract items and preserve empty list/maps
```
# Starting value
mapOfMaps:
  b:
    a: "x"
    c: "z"

# Call RemoveItems to remove all nested fields
RemoveItems(mapOfMaps.b.a, mapOfMaps.b.c)

# BUG: Returns null
mapOfMaps:
  b: null
```

In my case the list/map is not empty before calling RemoveItems(). Only after calling RemoveItems() then the list/map is empty and returns null.

I think the correct behavior is we preserve the list or map if it becomes empty after using RemoveItems

pjiang-dev:
fixes https://github.com/kubernetes-sigs/structured-merge-diff/issues/305

Modify tests for new behavior:
RemoveItems will maintain [] or {} instead of changing to null

### DIFF
--- a/merge/multiple_appliers_test.go
+++ b/merge/multiple_appliers_test.go
@@ -1061,14 +1061,14 @@ func TestMultipleAppliersNestedType(t *testing.T) {
 				},
 			},
 			Object: `
-				mapOfMapsRecursive:
-				  a:
-				  c:
-				    d:
-				      e:
-				        f:
-				          g:
-			`,
+			mapOfMapsRecursive:
+			  a: {}
+			  c:
+			    d:
+			      e:
+			        f:
+			          g:
+		`,
 			APIVersion: "v1",
 			Managed: fieldpath.ManagedFields{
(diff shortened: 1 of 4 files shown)


pjiang-dev:
Ping @jpbetz

jpbetz (approved):
/lgtm

jpbetz:
This broke valid, existing use cases. Working on a fix here: https://github.com/kubernetes-sigs/structured-merge-diff/issues/331

PR #310, opened 2025-12-18
Add (*Set).FromJSON benchmarks and switch to post-accumulation sort.

benluddy:
Set deserialization with children in reverse order performs an insertion sort. Sets serialized by structured-merge-diff are sorted, but in practice the serialization of sets is controlled by clients and they can (intentionally or not) exercise this worst case. Switching to a bulk insert followed by a sort appears to perform no worse in the best case (sorted) and is significantly better in the worst case:

```
                                              │     master     │                 bulk                 │
                                              │     sec/op     │    sec/op     vs base                │
SetFromJSON/children_in_ascending_order            1.286m ± 3%   1.341m ±  1%   +4.27% (p=0.000 n=10)
SetFromJSON/children_in_descending_order          20.189m ± 0%   1.385m ±  1%  -93.14% (p=0.000 n=10)
SetFromJSON/grandchildren_in_ascending_order       7.821m ± 1%   8.211m ±  1%   +4.98% (p=0.000 n=10)
SetFromJSON/grandchildren_in_descending_order      75.16m ± 4%   13.31m ± 13%  -82.28% (p=0.000 n=10)
geomean                                            11.11m        3.774m        -66.04%

                                              │     master     │                 bulk                 │
                                              │      B/op      │     B/op      vs base                │
SetFromJSON/children_in_ascending_order           1.320Mi ± 0%   1.320Mi ± 0%        ~ (p=1.000 n=10)
SetFromJSON/children_in_descending_order          1.320Mi ± 0%   1.320Mi ± 0%   +0.00% (p=0.000 n=10)
SetFromJSON/grandchildren_in_ascending_order      5.395Mi ± 0%   5.395Mi ± 0%   +0.00% (p=0.000 n=10)
SetFromJSON/grandchildren_in_descending_order     6.187Mi ± 0%   5.395Mi ± 0%  -12.80% (p=0.000 n=10)
geomean                                           2.761Mi        2.668Mi        -3.37%

                                              │     master     │                 bulk                  │
                                              │   allocs/op    │  allocs/op   vs base                  │
SetFromJSON/children_in_ascending_order            25.99k ± 0%   25.99k ± 0%        ~ (p=1.000 n=10) ¹
SetFromJSON/children_in_descending_order           25.99k ± 0%   25.99k ± 0%        ~ (p=1.000 n=10) ¹
SetFromJSON/grandchildren_in_ascending_order       138.4k ± 0%   138.4k ± 0%        ~ (p=1.000 n=10) ¹
SetFromJSON/grandchildren_in_descending_order      155.7k ± 0%   138.4k ± 0%  -11.11% (p=0.000 n=10)
geomean                                            61.77k        59.98k        -2.90%
¹ all samples are equal
```

This was originally proposed in https://github.com/kubernetes-sigs/structured-merge-diff/pull/292. All I've done here is added benchmarks to demonstrate the worst case performance and separated it from the json-iterator migration.


fieldpath/serialize.go
```diff
@@ -204,34 +205,25 @@ func readIterV1(iter *jsoniter.Iterator) (children *Set, isMember bool) {
...
 				children = &Set{}
 			}
-			m := &children.Members.members
-			// Since we expect that most of the time these will have been
-			// serialized in the right order, we just verify that and append.
-			appendOK := len(*m) == 0 || (*m)[len(*m)-1].Less(pe)
-			if appendOK {
-				*m = append(*m, pe)
-			} else {
-				children.Members.Insert(pe)
```
liggitt: does append + sort do the same thing as Insert (here) and Descend (below)? 

it looks like Insert _can_ be a no-op if the item already exists in the set:
https://github.com/kubernetes-sigs/structured-merge-diff/blob/3392408e4daa0089f96601a9ce96596f610ec813/fieldpath/element.go#L265-L267

and Descend _can_ return the existing set if the path already exists

https://github.com/kubernetes-sigs/structured-merge-diff/blob/3392408e4daa0089f96601a9ce96596f610ec813/fieldpath/set.go#L496-L498

It looks like switching these to unconditional appends can produce duplicates where that was impossible before?

If so, can you add tests that ensure this isn't changing behavior?

liggitt:
cc @jpbetz for detailed review

PR #312, opened 2026-02-19
Jsontagutil: Ignore unexported fields without json tags

alvaroaleman:
Without this, the `Extract` function generated by applyconfig-gen will error for a type that has an unexported field without a json tag:

```
error converting obj to typed: .spec.unexportedField: field not declared in schema
{
    msg: "error converting obj to typed: .spec.unexportedField: field not declared in schema",
    err: <typed.ValidationErrors | len:1, cap:1>[
        {
            Path: ".spec.unexportedField",
            ErrorMessage: "field not declared in schema",
        },
    ],
}
```


liggitt:
this looks reasonable, I verified this had no effect on kubernetes/kubernetes applyconfigs

```
hack/pin-dependency.sh sigs.k8s.io/structured-merge-diff/v6=github.com/alvaroaleman/structured-merge-diff/v6 unexported
hack/update-vendor.sh
make clean
hack/update-codegen.sh applyconfigs
```

/lgtm
/approve

PR #313, opened 2026-02-19
Valuereflect: Add support for arrays

### LINKED ISSUE #311: Sized array unsupported/causes panic
When calling the `Extract` function of a type that has the following field:

```
	// Array [3]int `json:"array,omitempty"`
```

structured-merge-diff panics:
```
  unsupported type: [3]int

  Full Stack Trace
    sigs.k8s.io/structured-merge-diff/v6/value.kind({0x106a5e680?, 0x113ea36359d8?, 0x106cde000?})
        /Users/alvaro/go/pkg/mod/sigs.k8s.io/structured-merge-diff/v6@v6.3.2-0.20260122202528-d9cc6641c482/value/valuereflect.go:191 +0x43c
    sigs.k8s.io/structured-merge-diff/v6/value.(*valueReflect).reuse(0x113ea3e95080, {0x106a5e680?, 0x113ea36359d8?, 0x113ea36359d8?}, 0x105b3b748?, 0x0, 0x0)
        /Users/alvaro/go/pkg/mod/sigs.k8s.io/structured-merge-diff/v6@v6.3.2-0.20260122202528-d9cc6641c482/value/valuereflect.go:82 +0x200
    sigs.k8s.io/structured-merge-diff/v6/value.(*valueReflect).mustReuse(...)
        /Users/alvaro/go/pkg/mod/sigs.k8s.io/structured-merge-diff/v6@v6.3.2-0.20260122202528-d9cc6641c482/value/valuereflect.go:89
    sigs.k8s.io/structured-merge-diff/v6/value.structReflect.IterateUsing.func1(0x113ea430f720?, {0x106990502, 0x5}, {0x106a5e680?, 0x113ea36359d8?, 0x106c72e60?})
        /Users/alvaro/go/pkg/mod/sigs.k8s.io/structured-merge-diff/v6@v6.3.2-0.20260122202528-d9cc6641c482/value/structreflect.go:112 +0x50
    sigs.k8s.io/structured-merge-diff/v6/value.eachStructField({0x106cde000?, 0x113ea3635110?, 0x15?}, 0x113ea3e16f80)
        /Users/alvaro/go/pkg/mod/sigs.k8s.io/structured-merge-diff/v6@v6.3.2-0.20260122202528-d9cc6641c482/value/structreflect.go:123 +0x10c
    sigs.k8s.io/structured-merge-diff/v6/value.structReflect.IterateUsing({{0x0, 0x0, {0x106cde000, 0x113ea3635110, 0x199}, 0x1}}, {0x106d031c0, 0x113ea347e980}, 0x113ea4091c00)
```

alvaroaleman:
Without this, the Extract function generated by applyconfig-gen panics:
```
unsupported type: [3]int
```

Fixes https://github.com/kubernetes-sigs/structured-merge-diff/issues/311

### DIFF
--- a/value/valuereflect.go
+++ b/value/valuereflect.go
@@ -173,6 +173,12 @@ func kind(v reflect.Value) reflectType {
 		return stringType
 	case reflect.Bool:
 		return boolType
+	case reflect.Array:
+		elemKind := typ.Elem().Kind()
+		if elemKind == reflect.Uint8 {
+			return byteStringType
+		}
+		return listType
 	case reflect.Slice:
 		if v.IsNil() {
 			return nullType
--- a/value/valuereflect_test.go
+++ b/value/valuereflect_test.go
@@ -627,6 +627,13 @@ func TestReflectList(t *testing.T) {
 			expectedUnstructured: []interface{}{"converted1", "converted2"},
 			length:               2,
 		},
+		{
+			name:                 "stringArray",
+			val:                  [2]string{"value1", "value2"},
+			expectedIterate:      []interface{}{"value1", "value2"},
+			expectedUnstructured: []interface{}{"value1", "value2"},
+			length:               2,
+		},
 	}
 
 	for _, tc := range cases {


liggitt:
Do you have examples of REST APIs using array types (not slice types)? Wouldn't that also error at decode time with incorrect length inputs?

value/valuereflect.go
```diff
@@ -173,6 +173,12 @@ func kind(v reflect.Value) reflectType {
 		return stringType
 	case reflect.Bool:
 		return boolType
+	case reflect.Array:
+		elemKind := typ.Elem().Kind()
+		if elemKind == reflect.Uint8 {
+			return byteStringType
```
liggitt, line 179: this special-case doesn't apply to arrays ... json marshals [n]byte as an actual list, not a base-64-encoded string

see https://go.dev/play/p/fG8JTbZSMjR

liggitt:
I'm not convinced array support in the API is a good idea... making the generator fail on fields of this type might be a better way to move this error earlier to generate time rather than runtime

alvaroaleman:
> Do you have examples of REST APIs using array types (not slice types)? Wouldn't that also error at decode time with incorrect length inputs?

Sorry for taking so long to come back to this. We have a giant sample CRD in controller-tools where ppl add fields that they found problematic over time and we tried to use it for applyconfig generation and currently have a few of these fields commented out: https://github.com/kubernetes-sigs/controller-tools/blob/e0458524f9c6c22915ccfb7b00311ec49410986e/pkg/applyconfiguration/testdata/cronjob/api/v1/cronjob_types.go#L330-L332

The specific case is here: https://github.com/kubernetes-sigs/controller-tools/blob/e0458524f9c6c22915ccfb7b00311ec49410986e/pkg/crd/testdata/cronjob_types.go#L348
This was added in https://github.com/kubernetes-sigs/controller-tools/pull/648 to fix https://github.com/kubernetes-sigs/controller-tools/issues/613

This all predated my involvement in the project and it looks like that PR fixed two issues and all the linked repos ran into the other issue. So no, I don't have an example of a project using fixed-length arrays in their API.

Should we close this and if/when someone complains that their SSA generation doesn't work and its because of this, we can revisit?

liggitt:
If there's a way to move this from a runtime error to a generate-time error, or from a panic to an error, that seems useful, but that's as far as I would go for now.

PR #323, opened 2026-04-08
fix(value): detect inline fields structurally instead of matching ",inline" tag

aramase:
xref: https://github.com/kubernetes/kubernetes/pull/138260

### DIFF
--- a/value/reflectcache_test.go
+++ b/value/reflectcache_test.go
@@ -253,13 +253,96 @@ func TestTypeReflectEntryOf(t *testing.T) {
 				},
 			},
 		},
-		"StructWithInlinedField": {
+		"StructWithNonEmbeddedInlineTag": {
 			arg: struct {
 				F1 string `json:",inline"`
 			}{},
 			want: &TypeReflectCacheEntry{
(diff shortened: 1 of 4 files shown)


value/reflectcache_test.go
```diff
@@ -253,13 +253,27 @@ func TestTypeReflectEntryOf(t *testing.T) {
 				},
 			},
 		},
-		"StructWithInlinedField": {
+		"StructWithNonEmbeddedInlineTag": {
```
liggitt, line 256: Thanks for fixing this test case

Can you also add tests for an inlined field of this type:
```
type Inlined struct {
  F1 string `json:"f1"`
}
```

with these test cases:
```
arg: struct {
  Inlined `json:",inline"`
}

arg: struct {
  Inlined `json:""`
}

arg: struct {
  Inlined
}
```

(I know that overlaps with the tagutil unit test a bit, but I want to make sure we're recursing fields correctly here when we *hit* an inlined field)

liggitt:
tagged v6.4.0 with this

PR #324, opened 2026-05-08
Disallow trailing data in value.FromJSON and fieldpath.FromJSON and add streaming token tests

### LINKED ISSUE #318: Validate Token Type Errors in readObject
### Background
During the code review for the `encoding/json/v2` migration in PR [#292](https://github.com/kubernetes-sigs/structured-merge-diff/pull/292), it was identified that we need stricter validation around token types during object parsing.

### Description
We need to develop a test case to explicitly confirm that the parser correctly handles unexpected token types. Specifically, if the parser encounters a token type other than `EndObject` or `String` at a position where a JSON key is expected, the system must return the appropriate, well-defined error. 

This ensures that malformed JSON is caught early and produces a standard error matching previous behavior.

### Code Pointers
* **File:** [`typed/parser.go`](https://github.com/kubernetes-sigs/structured-merge-diff/pull/292/files#diff-e55e5828452f14643c75edc316f43e06f9e5b1580f4f9104b2b2361623194a2f)
* **Function to Test:** `readObject`
* **Reference PR Comment:** [Discussion Link](https://github.com/kubernetes-sigs/structured-merge-diff/pull/292#discussion_r2589587228)

### Metadata
* **Priority:** P0
* **Complexity:** Medium

jpbetz:
Fixes https://github.com/kubernetes-sigs/structured-merge-diff/issues/318

### DIFF
--- a/fieldpath/serialize_test.go
+++ b/fieldpath/serialize_test.go
@@ -313,6 +313,60 @@ func TestDeserializeForValidNonNormalized(t *testing.T) {
 		})
 	}
 }
+func TestSetFromJSONTokenTypeErrors(t *testing.T) {
+	cases := []struct {
+		name  string
+		input string
+	}{
+		{"rootArray", `[]`},
+		{"rootNumber", `42`},
+		{"rootString", `"hi"`},
+		{"empty", ``},
+		{"whitespaceOnly", `   `},
+
+		{"numberAsKey", `{1:{}}`},
+		{"trailingCommaAfterEntry", `{"f:a":{},}`},
+		{"unclosedAfterOpen", `{`},
+		{"unclosedAfterValue", `{"f:a":{}`},
+		{"unclosedAfterTrailingComma", `{"f:a":{},`},
+
+		{"closedBeforeComma", `{"f:a" }:{}}`},
+		{"closedBeforeValue", `{"f:a":}`},
+
+		{"valueIsNumber", `{"f:a":1}`},
+		{"valueIsArray", `{"f:a":[]}`},
+		{"valueOpenedAsArrayClosedAsObject", `{"f:a":[}`},
+		{"valueOpenedAsObjectClosedAsArray", `{"f:a":{]}`},
+
+		{"doubleCommaBetweenEntries", `{"f:a":{},,"f:b":{}}`},
+		{"missingCommaBetweenEntries", `{"f:a":{} "f:b":{}}`},
+		{"doubleColonAfterKey", `{"f:a"::{}}`},
+		{"colonWithoutKey", `{:1}`},
+		{"keyWithoutColon", `{"f:a"}`},
+		{"leadingCommaInsideValue", `{"f:a":{,}}`},
+
+		{"bareCloseBrace", `}`},
+
+		{"keyEmptyString", `{"":{}}`},
+		{"keyMissingColon", `{"x":{}}`},
+		{"keyOnlyColon", `{":":{}}`},
+
+		// TODO: Trailing data should not be allowed
+		//{"trailingClosed", `{"f:a":{}}}`},
+		//{"trailingObject", `{"f:a":{}}{}`},
+		//{"trailingValue", `{"f:a":{}}"string"`},
+	}
+
+	for _, tc := range cases {
+		t.Run(tc.name, func(t *testing.T) {
+			s := NewSet()
+			if err := s.FromJSON(strings.NewReader(tc.input)); err == nil {
+				t.Fatalf("expected error for %q, got set: %v", tc.input, s)
+			}
+		})
+	}
+}
+
 func TestDropUnknown(t *testing.T) {
 	input := `{"f:aaa":{},"r:aab":{}}`
 	expect := `{"f:aaa":{}}`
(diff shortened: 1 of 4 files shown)


fieldpath/serialize_test.go
```diff
@@ -313,6 +313,60 @@ func TestDeserializeForValidNonNormalized(t *testing.T) {
...
+
+		{"keyEmptyString", `{"":{}}`},
+		{"keyMissingColon", `{"x":{}}`},
+		{"keyOnlyColon", `{":":{}}`},
+
+		// TODO: Trailing data should not be allowed
+		//{"trailingClosed", `{"f:a":{}}}`},
+		//{"trailingObject", `{"f:a":{}}{}`},
+		//{"trailingValue", `{"f:a":{}}"string"`},
+	}
```
jpbetz: Looks like jsonIter consumes the entire buffer, but our token parsing ignores trailing data.

Patch now?  My understanding is that jsonv2 is strict on this.

cc @liggitt @lalitc375

fieldpath/serialize_test.go
```diff
@@ -313,6 +313,60 @@ func TestDeserializeForValidNonNormalized(t *testing.T) {
...
+		{"empty", ``},
+		{"whitespaceOnly", `   `},
+
+		{"numberAsKey", `{1:{}}`},
+		{"trailingCommaAfterEntry", `{"f:a":{},}`},
+		{"unclosedAfterOpen", `{`},
+		{"unclosedAfterValue", `{"f:a":{}`},
+		{"unclosedAfterTrailingComma", `{"f:a":{},`},
+
+		{"closedBeforeComma", `{"f:a" }:{}}`},
```
liggitt: not sure I understand the "closedBeforeComma" name for this one

fieldpath/serialize_test.go
```diff
@@ -313,6 +313,60 @@ func TestDeserializeForValidNonNormalized(t *testing.T) {
...
+		{"keyWithoutColon", `{"f:a"}`},
+		{"leadingCommaInsideValue", `{"f:a":{,}}`},
+
+		{"bareCloseBrace", `}`},
+
+		{"keyEmptyString", `{"":{}}`},
+		{"keyMissingColon", `{"x":{}}`},
+		{"keyOnlyColon", `{":":{}}`},
+
+		// TODO: Trailing data should not be allowed
```
liggitt: also make sure we have test cases capturing whether internal and trailing whitespace is allowed or treated as an error (if we don't already have that ... we may)

fieldpath/serialize_test.go
liggitt: Yeah, it seems wrong to consume the whole buffer but ignore semantically invalid trailing tokens. Tightening that up to make sure the remaining buffer is either empty or only contains whitespace sounds good to me.

I have a hard time telling if there's any actual exposure from the top-level managedFields or not. Before fixing this, might be worth a sanity check test demonstrating whatever our entry-points are for parsing managedFields would fail if a field path tried to set one of these invalid trailing data things.

fieldpath/serialize_test.go
jpbetz: Very good catch.

fieldpath/serialize_test.go
jpbetz: I think I copy-pasted a test case when adding one and forgot to rename it.  Fixing.

jpbetz:
> Yeah, it seems wrong to consume the whole buffer but ignore semantically invalid trailing tokens. Tightening that up to make sure the remaining buffer is either empty or only contains whitespace sounds good to me.

Added.

> I have a hard time telling if there's any actual exposure from the top-level managedFields or not. Before fixing this, might be worth a sanity check test demonstrating whatever our entry-points are for parsing managedFields would fail if a field path tried to set one of these invalid trailing data things.

I've looked into this. I cannot find any exposure. For Protobuf/CBOR all the data is famed, so if a writer adds trailing data, the reader just ignores it.  For JSON, [FIeldV1](https://github.com/kubernetes/kubernetes/blob/e8742a6ba73fba3729ed9b44b89dd9540f48fd5f/staging/src/k8s.io/apimachinery/pkg/apis/meta/v1/fieldsv1.go#L89) handles the marshallening and there is no opportunity to smuggle bytes into the surrounding metadata fields.

jpbetz:
Fix is in place for trailing and tests are added.

value/value_test.go
```diff
@@ -0,0 +1,87 @@
...
+
+		{"bareCloseBrace", `}`},
+		{"bareCloseBracket", `]`},
+		{"bareComma", `,`},
+		{"bareColon", `:`},
+
+		// TODO: Trailing data should not be allowed
+		//{"trailingClosed", `{"f:a":{}}}`},
+		//{"trailingObject", `{"f:a":{}}{}`},
+		//{"trailingValue", `{"f:a":{}}"string"`},
```
liggitt: should we be able to uncomment these and they error now?

jpbetz: Fixed

jpbetz:
@liggitt I think this one is ready

liggitt:
lgtm, before merge can you open a dry-run bump to your branch in k/k to make sure everything there is happy with this as well?

jpbetz:
Dry run test: https://github.com/kubernetes/kubernetes/pull/139000

liggitt:
dry run was green

/hold cancel

PR #326, opened 2026-05-12
Add a direct unstructured conversion interface that can return errors.

benluddy:
The existing UnstructuredConverter interface method signature `ToUnstructured() any` can only be
implemented by types that are unconditionally convertible to unstructured. However, types that
implement json.Marshaler (`MarshalJSON() ([]byte, error)`) are _conditionally_ convertible to
unstructured via JSON. For example, a value of k8s.io/apimachinery/pkg/util/intstr.IntOrString only
has a JSON representation when its Type fields is one of "Int" or "String".

### DIFF
--- a/value/reflectcache_test.go
+++ b/value/reflectcache_test.go
@@ -18,6 +18,7 @@ package value
 
 import (
 	"encoding/json"
+	"errors"
 	"fmt"
 	"reflect"
 	"testing"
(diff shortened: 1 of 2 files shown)


benluddy:
/retitle Add a direct unstructured conversion interface that can return errors.

value/reflectcache_test.go
```diff
@@ -142,6 +286,22 @@ func TestTimeToUnstructured(t *testing.T) {
...
+	}
+	result, err := entry.ToUnstructured(rv)
+	if err != nil {
+		t.Fatal(err)
+	}
+	if result != "hello" {
+		t.Errorf("expected %q but got %#v", "hello", result)
+	}
+}
+
```
benluddy, line 304: Coverage had been missing for the "addressable value whose pointer-to type implements UnstructuredConverter" case.

jpbetz:
@lalitc375, this might be useful in optimizing CEL lazy conversion as well

PR #327, opened 2026-05-14
Add unicode and escaping test cases

### LINKED ISSUE #319: Test Key Encoding and Unicode in Parser
### Background
When replacing the JSON parsing engine in PR [#292](https://github.com/kubernetes-sigs/structured-merge-diff/pull/292), one of the riskiest areas is the handling of text encoding and Unicode, particularly within JSON keys. 

### Description
We must create extensive test cases to cover various key encoding and Unicode scenarios. These tests must run against both the legacy (`json-iterator`) and the new (`json/v2`) implementations to guarantee they produce the exact same in-memory values.

The scenarios that need coverage include:
* **Escaped/encoded keys:** e.g., keys formatted like `"\u0041"`.
* **Multi-byte Unicode keys:** e.g., `"Iñtërnâtiônàlizætiøn,💝🐹🌇⛔"`.
* **Invalid Unicode characters:** Keys containing byte sequences that do not form valid UTF-8/Unicode characters. We need to verify how both implementations handle/reject these.

### Code Pointers
* **File for New Tests:** [`typed/parser_test.go`](https://github.com/kubernetes-sigs/structured-merge-diff/pull/292/files#diff-1c19b675402a5c1b69657cc17ff582b130a05a417dd0dca6a2342938d8213e9a)
* **Function to Test:** `readObject` in [`typed/parser.go`](https://github.com/kubernetes-sigs/structured-merge-diff/pull/292/files#diff-e55e5828452f14643c75edc316f43e06f9e5b1580f4f9104b2b2361623194a2f)
* **Reference PR Comment:** [Discussion Link](https://github.com/kubernetes-sigs/structured-merge-diff/pull/292#discussion_r2589596681)

### Metadata
* **Priority:** P0
* **Complexity:** Easy

jpbetz:
Fixes https://github.com/kubernetes-sigs/structured-merge-diff/issues/319

### DIFF
--- a/value/value_test.go
+++ b/value/value_test.go
@@ -167,6 +167,21 @@ func TestFromJSONValid(t *testing.T) {
 		{"surroundingWhitespace", " \t\n {\"a\":1} \t\n", `{"a":1}`},
 		{"surroundingWhitespaceAroundNull", " null ", `null`},
 		{"surroundingWhitespaceAroundBool", " true ", `true`},
+
+		// Unicode keys
+		{"multiByteLatin", `{"Iñtërnâtiônàlizætiøn":1}`, `{"Iñtërnâtiônàlizætiøn":1}`},
+		{"cyrillic", `{"Здравствуйте":1}`, `{"Здравствуйте":1}`},
+		{"unifiedChars", `{"こんにちは世界":1}`, `{"こんにちは世界":1}`},
+		{"multiByteUnicode", `{"😂💝🐹🌇⛔":1}`, `{"😂💝🐹🌇⛔":1}`},
+
+		// Unicode escaping
+		{"escapedAsciiKey", `{"\u0041":1}`, `{"A":1}`},
+		{"escapedLatin", `{"\u00c5":1}`, `{"Å":1}`},
+		{"escapedMultiBytePair", `{"\uD83D\uDE02":1}`, `{"😂":1}`},
+
+		// TODO: When we migrate to json/v2 we should use jsontext.AllowInvalidUTF8
+		// and change this test case to expect the bytes to be munged to U+FFFD.
+		{"byteOrderMark", "{\"\xff\xfe\": 1}", "{\"\xff\xfe\": 1}"},
 	}
 	for _, tc := range cases {
 		t.Run(tc.name, func(t *testing.T) {


value/value_test.go
```diff
@@ -167,6 +167,20 @@ func TestFromJSONValid(t *testing.T) {
...
+		{"unifiedChars", `{"こんにちは世界":1}`, `{"こんにちは世界":1}`},
+		{"multiByteUnicode", `{"😂💝🐹🌇⛔":1}`, `{"😂💝🐹🌇⛔":1}`},
+
+		// Unicode escaping
+		{"escapedAsciiKey", `{"\u0041":1}`, `{"A":1}`},
+		{"escapedLatin", `{"\u00c5":1}`, `{"Å":1}`},
+		{"escapedMultiBytePair", `{"\uD83D\uDE02":1}`, `{"😂":1}`},
+
+		// BOM
+		{"byteOrderMark", "{\"\xff\xfe\": 1}", "{\"\xff\xfe\": 1}"}, // This is not-standard JSON
```
jpbetz: @liggitt 

json-iterator is too lenient here.  It looks like json/v2 does not allow this.

It does appear to be possible to write this to storage:

```yaml
apiVersion: apiextensions.k8s.io/v1
kind: CustomResourceDefinition
metadata:
  name: examples.stable.example.com
spec:
  group: stable.example.com
  versions:
    - name: v1
      served: true
      storage: true
      schema:
        openAPIV3Schema:
          type: object
          properties:
            spec:
              type: object
              properties:
                "\xff\xfe":
                  type: integer
                map:
                  type: object
                  additionalProperties:
                    type: integer
  scope: Namespaced
  names:
    plural: examples
    singular: example
    kind: Example
---
apiVersion: stable.example.com/v1
kind: Example
metadata:
  name: example1
spec:
  "\xff\xfe": 1
  map:
    "\xff\xfe": 1
```

```sh
kubectl get Example example1 -oyaml
```

```yaml
apiVersion: stable.example.com/v1
kind: Example
metadata:
  annotations:
    kubectl.kubernetes.io/last-applied-configuration: |
      {"apiVersion":"stable.example.com/v1","kind":"Example","metadata":{"annotations":{},"name":"example1","namespace":"default"},"spec":{"map":{"ÿþ":1},"ÿþ":1}}
  creationTimestamp: "2026-05-14T15:30:28Z"
  generation: 1
  name: example1
  namespace: default
  resourceVersion: "351"
  uid: 64ca736a-0a57-4728-9b33-b706b6b4d42b
spec:
  map:
    ÿþ: 1
  ÿþ: 1
```

jpbetz: I'm checking if I did something wrong or am misunderstanding JSON handling of BOM. I wasn't expecting a CRD to allow the bytes to be used a a property name or map key.

jpbetz: Looks like https://pkg.go.dev/encoding/json/jsontext#AllowInvalidUTF8 would allows the bytes to be read out, but  mangles them to `U+FFFD`: https://go.dev/play/p/x9vjwvhRTDk

jpbetz: I'll put a note in the comment in this PR.  My high level take away so far:

- json-iterator tolerates invalid bytes
- encoding/json (v1) allows invalid bytes but mundges them to valid bytes in some cases
- json/v2 is strict by default but can be configured via AllowInvalidUTF8 to mundge like v1 did

We're probably going to need to configure json/v2 to do this.  And we'll need to do ~everywhere (not just where we use json-iterator)

PR #328, opened 2026-05-19
Add edge-case tests for DeserializePathElement

### LINKED ISSUE #316: Add Edge-Case Tests for DeserializePathElement
### Background
As part of the migration to `encoding/json/v2` in PR [#292](https://github.com/kubernetes-sigs/structured-merge-diff/pull/292), we need to ensure that our deserialization logic remains robust and perfectly backward-compatible with the old `json-iterator` implementation. 

### Description
We need to develop new test cases to cover various edge scenarios for the `DeserializePathElement` function. This is critical to ensure robust error handling and correct behavior under unexpected inputs. 

The scenarios that must be explicitly tested include:
* **Zero-length input:** How does the function handle an empty string?
* **EOF conditions:** What happens when the input stream unexpectedly ends?
* **Leading or trailing whitespace:** Ensure whitespace around the value is handled identically to the previous implementation.
* **Multi-token inputs:** Ensure inputs with multiple tokens fail as expected.

### Goal
Capture the current behavior and prevent us from accidentally changing it. If these tests fail when we switch to `json-iter` (or `json/v2`), we'll be aware and can decide how to handle the differences the tests find.

### Code Pointers
* **File for New Tests:** [`fieldpath/serialize-pe_test.go`](https://github.com/kubernetes-sigs/structured-merge-diff/pull/292/files#diff-3f0a927a76059e6a4b3d75156a0642e05f63fa942f2b5a5912443428a1c93f01)
* **Function to Test:** `DeserializePathElement` in [`fieldpath/serialize-pe.go`](https://github.com/kubernetes-sigs/structured-merge-diff/pull/292/files#diff-23964724a1be7001402377a339a0378b1740921021706637371901a1d948950c)
* **Reference PR Comment:** [Discussion Link](https://github.com/kubernetes-sigs/structured-merge-diff/pull/292#discussion_r2589524546)

### Metadata
* **Priority:** P0
* **Complexity:** Easy

lalitc375:
Fixes #316

This PR introduces test cases to `fieldpath/serialize-pe_test.go` to explicitly document and test edge-case behavior for `DeserializePathElement`. Capturing the current `json-iterator` behavior is a necessary prerequisite to ensure a smooth transition when migrating to `encoding/json/v2`.

#### 1. Zero-length and Short Inputs
- Validates that completely empty strings (`""`) fail immediately.
- Validates that strings with only a single character prefix (`"v"`, `"k"`, `"i"`, `"f"`) fail properly before proceeding.

#### 2. EOF Conditions and Malformed JSON
- Validates that abruptly ended values (e.g. `v:{"a":"b"`, `k:{"a":"b"`) throw the expected EOF/unclosed object errors.
- Validates that a key indicator missing its JSON component (`v:`, `i:`) fails.

#### 3. Leading and Trailing Whitespace
- Validates that values preceded by spaces inside the element (e.g. `v: {"some":"json"}`) are successfully parsed as they were previously.
- Validates that trailing whitespace following a valid JSON object (`v:{"some":"json"} `) is handled seamlessly.

#### 4. Multi-token Inputs
- Validates that trailing, un-associated JSON tokens (e.g. `v:{"some":"json"} {"other":"json"}`) currently **succeed** by effectively ignoring the secondary token. 
*Note: This specific behavior will likely change/break under `encoding/json/v2`, but is captured here as a success to accurately reflect the current baseline logic.*

### DIFF
--- a/fieldpath/serialize-pe.go
+++ b/fieldpath/serialize-pe.go
@@ -97,6 +97,18 @@ func DeserializePathElement(s string) (PathElement, error) {
 		if err != nil {
 			return PathElement{}, err
 		}
+		// Note: For simple JSON primitives (like integers or bare strings) that terminate
+		// exactly at the stream boundary, jsoniter's parser sets iter.Error to io.EOF during
+		// token scan validation, which is a successful termination indicator and not a parsing
+		// error. We explicitly filter out io.EOF to avoid false parsing failure returns.
+		if iter.Error != nil && iter.Error != io.EOF {
+			return PathElement{}, iter.Error
+		}
+		if iter.WhatIsNext(); iter.Error == nil {
+			// Piggy back on scanner aware error reporting here.
+			iter.ReportError("managedFields parsing", "unexpected trailing data after JSON object")
+			return PathElement{}, iter.Error
+		}
 		return PathElement{Value: &v}, nil
 	case peKeySepBytes[0]:
 		iter := readPool.BorrowIterator(b)
@@ -112,7 +124,21 @@ func DeserializePathElement(s string) (PathElement, error) {
 			fields = append(fields, value.Field{Name: key, Value: v})
 			return true
 		})
+		if iter.Error != nil && iter.Error != io.EOF {
+			return PathElement{}, iter.Error
+		}
+		if iter.WhatIsNext(); iter.Error == nil {
+			// Piggy back on scanner aware error reporting here.
+			iter.ReportError("managedFields parsing", "unexpected trailing data after JSON object")
+			return PathElement{}, iter.Error
+		}
 		fields.Sort()
+		// Note: When lookahead detects that there is no unexpected trailing data after
+		// a valid key element, it sets iter.Error to io.EOF at the stream's natural boundary.
+		// io.EOF signifies successful termination and must not be bubbled back as an error.
+		if iter.Error == io.EOF {
+			return PathElement{Key: &fields}, nil
+		}
 		return PathElement{Key: &fields}, iter.Error
 	case peIndexSepBytes[0]:
 		i, err := strconv.Atoi(s[2:])
--- a/fieldpath/serialize-pe.go
+++ b/fieldpath/serialize-pe.go
@@ -97,16 +97,12 @@ func DeserializePathElement(s string) (PathElement, error) {
 		if err != nil {
 			return PathElement{}, err
 		}
-		// Note: For simple JSON primitives (like integers or bare strings) that terminate
-		// exactly at the stream boundary, jsoniter's parser sets iter.Error to io.EOF during
-		// token scan validation, which is a successful termination indicator and not a parsing
-		// error. We explicitly filter out io.EOF to avoid false parsing failure returns.
-		if iter.Error != nil && iter.Error != io.EOF {
-			return PathElement{}, iter.Error
-		}
-		if iter.WhatIsNext(); iter.Error == nil {
-			// Piggy back on scanner aware error reporting here.
-			iter.ReportError("managedFields parsing", "unexpected trailing data after JSON object")
+		// Lookahead validates that there is no unexpected trailing data.
+		// io.EOF is a successful termination indicator; all other errors or trailing data fail.
+		if iter.WhatIsNext(); iter.Error != io.EOF {
+			if iter.Error == nil {
+				iter.ReportError("managedFields parsing", "unexpected trailing data after JSON object")
+			}
 			return PathElement{}, iter.Error
 		}
 		return PathElement{Value: &v}, nil
@@ -124,22 +120,16 @@ func DeserializePathElement(s string) (PathElement, error) {
 			fields = append(fields, value.Field{Name: key, Value: v})
 			return true
 		})
-		if iter.Error != nil && iter.Error != io.EOF {
-			return PathElement{}, iter.Error
-		}
-		if iter.WhatIsNext(); iter.Error == nil {
-			// Piggy back on scanner aware error reporting here.
-			iter.ReportError("managedFields parsing", "unexpected trailing data after JSON object")
+		// Lookahead validates that there is no unexpected trailing data.
+		// io.EOF is a successful termination indicator; all other errors or trailing data fail.
+		if iter.WhatIsNext(); iter.Error != io.EOF {
+			if iter.Error == nil {
+				iter.ReportError("managedFields parsing", "unexpected trailing data after JSON object")
+			}
 			return PathElement{}, iter.Error
 		}
 		fields.Sort()
-		// Note: When lookahead detects that there is no unexpected trailing data after
-		// a valid key element, it sets iter.Error to io.EOF at the stream's natural boundary.
-		// io.EOF signifies successful termination and must not be bubbled back as an error.
-		if iter.Error == io.EOF {
-			return PathElement{Key: &fields}, nil
-		}
-		return PathElement{Key: &fields}, iter.Error
+		return PathElement{Key: &fields}, nil
 	case peIndexSepBytes[0]:
 		i, err := strconv.Atoi(s[2:])
 		if err != nil {
--- a/fieldpath/serialize-pe_test.go
+++ b/fieldpath/serialize-pe_test.go
@@ -55,6 +55,8 @@ func TestPathElementRoundTrip(t *testing.T) {
 		{`k:{"name":"Привет, мир"}`, KeyElement(value.Field{Name: "name", Value: value.NewValueInterface("Привет, мир")})},
 		{`k:{"name":"नमस्ते दुनिया"}`, KeyElement(value.Field{Name: "name", Value: value.NewValueInterface("नमस्ते दुनिया")})},
 		{`k:{"name":"👋"}`, KeyElement(value.Field{Name: "name", Value: value.NewValueInterface("👋")})},
+		{`f:spec-🚀`, FieldNameElement("spec-🚀")},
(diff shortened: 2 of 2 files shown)


fieldpath/serialize-pe_test.go
```diff
@@ -107,3 +114,36 @@ func TestDeserializePathElementError(t *testing.T) {
...
+	tests := []testCase{
+		// Leading whitespace
+		{`v: {"some":"json"}`, ValueElement(value.NewValueInterface(map[string]interface{}{"some": "json"}))},
+		{`k: {"name":"my-container"}`, KeyElement(value.Field{Name: "name", Value: value.NewValueInterface("my-container")})},
+
+		// Trailing whitespace
+		{`v:{"some":"json"} `, ValueElement(value.NewValueInterface(map[string]interface{}{"some": "json"}))},
+		{`k:{"name":"my-container"} `, KeyElement(value.Field{Name: "name", Value: value.NewValueInterface("my-container")})},
+
+		// Multi-token (currently json-iterator ignores trailing tokens)
```
liggitt: is that a problem we need to fix? should we fix that in this PR and expect these trailing token cases to error?

is there any way to persist data with this problem today from k/k use?

fieldpath/serialize-pe_test.go
```diff
@@ -107,3 +114,36 @@ func TestDeserializePathElementError(t *testing.T) {
...
+		{`v:{"some":"json"} {"other":"json"}`, ValueElement(value.NewValueInterface(map[string]interface{}{"some": "json"}))},
+		{`k:{"name":"my-container"} {"other":"my-container"}`, KeyElement(value.Field{Name: "name", Value: value.NewValueInterface("my-container")})},
+	}
```
liggitt, line 162: also add tests for both escaped and unescaped single and multi-byte versions of string characters in the fieldName or key or value to make sure our decoding paths are interpreting escaped characters correctly and consistently (we should always end up with unescaped interpreted values in the in-memory version)

fieldpath/serialize-pe_test.go
liggitt: https://github.com/kubernetes/kubernetes/pull/139202 adds test coverage that this trailing data cannot be persisted, which is great

Let’s fix the parsing to error in these trailing data cases (like https://github.com/kubernetes-sigs/structured-merge-diff/pull/324/changes did), and change the two trailing token cases from success to error cases

lalitc375: Fixed the smd. PTAL.

fieldpath/serialize-pe_test.go
lalitc375: Added a few tests, PTAL. Let me know if it is  not what you were asking for.

fieldpath/serialize-pe.go
```diff
@@ -97,6 +97,18 @@ func DeserializePathElement(s string) (PathElement, error) {
...
+		// token scan validation, which is a successful termination indicator and not a parsing
+		// error. We explicitly filter out io.EOF to avoid false parsing failure returns.
+		if iter.Error != nil && iter.Error != io.EOF {
+			return PathElement{}, iter.Error
+		}
+		if iter.WhatIsNext(); iter.Error == nil {
+			// Piggy back on scanner aware error reporting here.
+			iter.ReportError("managedFields parsing", "unexpected trailing data after JSON object")
+			return PathElement{}, iter.Error
+		}
```
liggitt, line 107: we may need to repeat this block:
```
		if iter.Error != nil && iter.Error != io.EOF {
			return PathElement{}, iter.Error
		}
```

if there was another token, but it was malformed, iter.Error might be non-nil, in which case we want to error as well, right? make sure we have a test for this scenario

edit: or could maybe consolidate into the block above returning on anything other than an EOF error

fieldpath/serialize-pe.go
```diff
@@ -112,7 +124,21 @@ func DeserializePathElement(s string) (PathElement, error) {
 			fields = append(fields, value.Field{Name: key, Value: v})
 			return true
 		})
+		if iter.Error != nil && iter.Error != io.EOF {
+			return PathElement{}, iter.Error
+		}
+		if iter.WhatIsNext(); iter.Error == nil {
```
liggitt: can we simplify this to getting anything other than an EOF error and drop the `if iter.Error == io.EOF` block below?

```suggestion
		if iter.WhatIsNext(); iter.Error != io.EOF {
```

fieldpath/serialize-pe.go
```diff
@@ -97,6 +97,18 @@ func DeserializePathElement(s string) (PathElement, error) {
...
 			return PathElement{}, err
 		}
+		// Note: For simple JSON primitives (like integers or bare strings) that terminate
+		// exactly at the stream boundary, jsoniter's parser sets iter.Error to io.EOF during
+		// token scan validation, which is a successful termination indicator and not a parsing
+		// error. We explicitly filter out io.EOF to avoid false parsing failure returns.
+		if iter.Error != nil && iter.Error != io.EOF {
+			return PathElement{}, iter.Error
+		}
+		if iter.WhatIsNext(); iter.Error == nil {
```
liggitt: might be able to consolidate/simplify all three blocks to:

```suggestion
		if iter.WhatIsNext(); iter.Error != io.EOF {
```

fieldpath/serialize-pe_test.go
```diff
@@ -96,8 +98,19 @@ func TestDeserializePathElementError(t *testing.T) {
...
+		`v`,
+		`k`,
+		`f`,
+		`i`,
+		`v:{"a":"b"`,
+		`k:{"a":"b"`,
+		`i: 0`,
+		`i:0 `,
+		`v:{"some":"json"} {"other":"json"}`, // multiple values
+		`k:{"name":"my-container"} {"other":"my-container"}`, // multiple keys
```
liggitt, line 110: also add cases that have malformed trailing data so iter.Error would be non-nil but not io.EOF:

```
		`v:{"some":"json"} {"other":"json"`, // multiple values with malformed trailing data
		`k:{"name":"my-container"} {"other":"my-container"`, // multiple keys with malformed trailing data
```

fieldpath/serialize-pe_test.go
```diff
@@ -107,3 +120,54 @@ func TestDeserializePathElementError(t *testing.T) {
...
+		{`k: {"name":"my-container"}`, KeyElement(value.Field{Name: "name", Value: value.NewValueInterface("my-container")})},
+
+		// Trailing whitespace
+		{`v:{"some":"json"} `, ValueElement(value.NewValueInterface(map[string]interface{}{"some": "json"}))},
+		{`k:{"name":"my-container"} `, KeyElement(value.Field{Name: "name", Value: value.NewValueInterface("my-container")})},
+
+		// Single-byte escapes in map key of key element (`k`)
+		{`k:{"name\u002dcontainer":"my-container"}`, KeyElement(value.Field{Name: "name-container", Value: value.NewValueInterface("my-container")})},
+		{`k:{"name\nwith\nnewlines":"my-container"}`, KeyElement(value.Field{Name: "name\nwith\nnewlines", Value: value.NewValueInterface("my-container")})},
+		{`k:{"name\"quoted\"":"my-container"}`, KeyElement(value.Field{Name: "name\"quoted\"", Value: value.NewValueInterface("my-container")})},
```
liggitt: just to make it easier to see the escaped version gets interpreted:
```suggestion
		{`k:{"name\"quoted\"":"my-container"}`, KeyElement(value.Field{Name: `name"quoted"`, Value: value.NewValueInterface("my-container")})},
```

fieldpath/serialize-pe_test.go
```diff
@@ -107,3 +120,54 @@ func TestDeserializePathElementError(t *testing.T) {
...
+
+		// Multi-byte escapes in map key of key element (`k`)
+		{`k:{"name-\ud83d\ude80":"my-container"}`, KeyElement(value.Field{Name: "name-🚀", Value: value.NewValueInterface("my-container")})},
+		{`k:{"\u4f60\u597d":"\u4e16\u754c"}`, KeyElement(value.Field{Name: "你好", Value: value.NewValueInterface("世界")})},
+
+		// Single-byte escapes in value element (`v`)
+		{`v:"value\u002dcontainer"`, ValueElement(value.NewValueInterface("value-container"))},
+		{`v:"value\nwith\nnewlines"`, ValueElement(value.NewValueInterface("value\nwith\nnewlines"))},
+		{`v:"value\"quoted\""`, ValueElement(value.NewValueInterface("value\"quoted\""))},
```
liggitt:
```suggestion
		{`v:"value\"quoted\""`, ValueElement(value.NewValueInterface(`value"quoted"`))},
```

fieldpath/serialize-pe_test.go
liggitt: yup, exactly that to show us escape sequences get interpreted

fieldpath/serialize-pe.go
lalitc375: done

fieldpath/serialize-pe.go
lalitc375: done

fieldpath/serialize-pe_test.go
lalitc375: done

fieldpath/serialize-pe_test.go
lalitc375: done

fieldpath/serialize-pe_test.go
lalitc375: done

liggitt (commented):
/lgtm
/approve

Issue #331, opened 2026-07-01
RemoveItems incorrectly transforms null values to empty containers for nullable schemas


jpbetz:
In #306 `RemoveItems` was changed to treat a map or slice that becomes empty during removal as an empty container (`{}` / `[]`). Before, it would treat such a value as `null`.

This can result in validation errors on CRs that previously passed validation, since `{}` is interpreted as a present-but-empty struct, and if that struct has any required fields, `{}` will fail with a "field required" error.

Due to the severity of the bug we will revert the change and then look for ways to fix the underlying problem described in #305 another way. 

This issue is present k/k v1.36 and so a fix to revert #306 will need to be backported to the 1.36 branch.

### Fix

- [x] Revert #306 on master
  - [x] publish a v6.4.2 tag with the revert: https://github.com/kubernetes-sigs/structured-merge-diff/pull/334
- [x] Create a release-6.3 branch
  - [x] revert #306 on the branch: https://github.com/kubernetes-sigs/structured-merge-diff/commits/release-6.3
  - [x] publish a v6.3.3 tag with the revert: https://github.com/kubernetes-sigs/structured-merge-diff/tree/v6.3.3
- [x] Pin k/k master to v6.4.2 - https://github.com/kubernetes/kubernetes/pull/140294
- [ ] Pin k/k release-1.36 to v6.3.3 - https://github.com/kubernetes/kubernetes/pull/140296
- [ ] Find an alternative fix to #305

### How to reproduce

```yaml
apiVersion: apiextensions.k8s.io/v1
kind: CustomResourceDefinition
metadata:
  name: widgets.example.com
spec:
  group: example.com
  names:
    kind: Widget
    plural: widgets
    singular: widget
  scope: Namespaced
  versions:
  - name: v1
    served: true
    storage: true
    schema:
      openAPIV3Schema:
        type: object
        properties:
          spec:
            type: object
            properties:
              foo:
                type: object
                nullable: true
                required: [a, b]
                properties:
                  a:
                    type: string
                  b:
                    type: string
```

First create a CR:

```
apiVersion: example.com/v1
kind: Widget
metadata:
  name: w
spec:
  foo:
    a: "1"
    b: "2"
```

Then use SSA (`kubectl apply --server-side`) to patch it:

```
apiVersion: example.com/v1
kind: Widget
metadata:
  name: w
spec:
  foo: null
```

jpbetz:
[...]

PR #332, opened 2026-07-01
Add nullable type support

jpbetz:
Step 1 in fixing https://github.com/kubernetes-sigs/structured-merge-diff/issues/331

### DIFF
--- a/schema/elements.go
+++ b/schema/elements.go
@@ -62,6 +62,13 @@ type TypeRef struct {
 	// If this field is nil, then it has no effect.
 	// See `Map` and `List` for more information about `ElementRelationship`
 	ElementRelationship *ElementRelationship `yaml:"elementRelationship,omitempty"`
+
+	// Nullable indicates that an explicit null is a valid value,
+	// corresponding to OpenAPI's `nullable: true`. During removal,
+	// empty map/list values are handled differently if this is true.
+	// When true, an empty map or slice becomes null.
+	// When false, an empty map or slice is preserved as an empty container (`{}`/`[]`).
+	Nullable bool `yaml:"nullable,omitempty"`
 }
 
 // Atom represents the smallest possible pieces of the type system.
--- a/typed/remove.go
+++ b/typed/remove.go
@@ -26,6 +26,7 @@ type removingWalker struct {
 	toRemove      *fieldpath.Set
 	allocator     value.Allocator
 	shouldExtract bool
+	nullable      bool
 }
 
 // removeItemsWithSchema will walk the given value and look for items from the toRemove set.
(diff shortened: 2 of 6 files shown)


jpbetz:
/hold cancel
https://github.com/kubernetes/kubernetes/pull/140173 demonstrates that this works as intended. See https://github.com/kubernetes-sigs/structured-merge-diff/issues/331#issuecomment-4861176179 for details.

schema/elements.go
```diff
@@ -62,6 +62,13 @@ type TypeRef struct {
 	// If this field is nil, then it has no effect.
 	// See `Map` and `List` for more information about `ElementRelationship`
 	ElementRelationship *ElementRelationship `yaml:"elementRelationship,omitempty"`
+
+	// Nullable indicates that an explicit null is a valid value,
+	// corresponding to OpenAPI's `nullable: true`. During removal,
+	// empty map/list values are handled differently if this is true.
+	// When true, an empty map or slice becomes is left as null.
```
lalitc375: "an empty map or slice becomes is left as null."
->
"When true, an empty map or slice becomes null."

typed/remove.go
```diff
@@ -105,8 +107,8 @@ func (w *removingWalker) doList(t *schema.List) (errs ValidationErrors) {
 				wasList := item.IsList()
 				item = removeItemsWithSchema(item, w.toRemove.WithPrefix(pe), w.schema, t.ElementType, w.shouldExtract)
 				// If item returned null but we're removing items within the structure(not the item itself),
-				// preserve the empty container structure
-				if item.IsNull() && !w.shouldExtract {
+				// preserve the empty container structure unless the type is nullable.
```
lalitc375, line 110: L112-L116 is not tested. Can you add corresponding tests for map and list.

schema/elements.go
jpbetz: SG. In the future, you can use github suggestions for this type of suggestion and then I can apply it quickly.

typed/remove.go
jpbetz: OK, I've expanded the test coverage.

typed/remove_test.go
lalitc375, line 951: Multiple test cases are expected to return "unparsable". This test is not performing removal when it sees the expected value "unparsable".
This test needs fixing. It should run the removal and fail if there is no error.

jpbetz: OK. Is this pre-existing? Does it need fixing in this PR? I'm a bit unclear on what test gap you are pointing out for this PR specifically vs. a deficiency you've noticed in this test coverage overall..

lalitc375: The last test case is setting both expected outputs to unparsable. So that test case is not adding any value. You can remove it.

As you said, it is preexisting. You don't need to fix it.

yongruilin (commented):
/lgtm

PR #333, opened 2026-07-06
Add lalitc375 as one of the reviewers

lalitc375:
Add lalitc375 as one of the reviewers


jpbetz:
Looks like we need to merge https://github.com/kubernetes/org/issues/6488 and then we should be able to merge this.

/ok-to-test

PR #334, opened 2026-07-07
Revert #306 and #332 but keep Nullable schema field

jpbetz:
This is the first step in plan to fix the regression described in https://github.com/kubernetes-sigs/structured-merge-diff/issues/331

This reverts #306 and most of #332 but keeps the nullable field since that is consumed by kube-openapi today.


jpbetz:
/hold cancel
e2e test passes locally

PR #338, opened 2026-08-25
Test behavior of empty and null key elements

liggitt:
cc @jpbetz 

Hoists the test case from https://github.com/kubernetes-sigs/structured-merge-diff/pull/292/changes#r3856962677 into master to capture json-iterator behavior


liggitt:
sorry, repushed the commit, needs lgtm again

PR #340, opened 2026-08-25
Test behavior of duplicate sets

liggitt:
/assign @cici37

### DIFF
--- a/fieldpath/serialize_test.go
+++ b/fieldpath/serialize_test.go
@@ -19,6 +19,7 @@ package fieldpath_test
 import (
 	"bytes"
 	"fmt"
+	"reflect"
 	"strings"
 	"testing"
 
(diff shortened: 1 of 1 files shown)


fieldpath/serialize_test.go
```diff
@@ -390,35 +391,40 @@ func TestSetFromJSONTokenTypeErrors(t *testing.T) {
...
+		{name: "whitespaceAfterColon", input: `{"f:a":     {}}`, expect: NewSet(MakePathOrDie("a")), output: `{"f:a":{}}`},
 
 		// A bare null is accepted and produces an empty set.
-		{"bareNull", `null`},
-		{"nullWithSurroundingWhitespace", ` null `},
+		{name: "bareNull", input: `null`, expect: NewSet(), output: `{}`},
+		{name: "nullWithSurroundingWhitespace", input: ` null `, expect: NewSet(), output: `{}`},
+
+		// Duplicates are accepted, last one wins.
+		{name: "duplicates", input: `{"f:a":{"f:b":{},"f:b":{}},"f:a":{"f:c":{},"f:c":{}}} `, expect: NewSet(MakePathOrDie("a", "c")), output: `{"f:a":{"f:c":{}}}`},
```
liggitt, line 427: this is the only new test case

the other changes just added assertions of in-memory state after parsing, and resulting JSON from round-tripping back out

PR #344, opened 2026-08-27
Make benchmark deterministic

liggitt:
When running benchstat on https://github.com/kubernetes-sigs/structured-merge-diff/pull/292, we need the benchmarks to be deterministic. This adjusts the generates paths and path selection to be deterministic for the runs, makes sure they run successfully, and temporarily comments out a duplicate key value from the random generator which failed mid-PR commits in https://github.com/kubernetes-sigs/structured-merge-diff/pull/292 (the duplicate behavior is exercised in a non-benchmark unit test)

cc @jpbetz

### DIFF
--- a/fieldpath/set_test.go
+++ b/fieldpath/set_test.go
@@ -27,16 +27,29 @@ import (
 	"sigs.k8s.io/structured-merge-diff/v6/value"
 )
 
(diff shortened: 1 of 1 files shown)


jpbetz (commented):
/approve
/lgtm

PR #347, opened 2026-09-01
Fix 32-bit int overflow in fieldpath set_test.go

arthurbdiniz:
How to reproduce:

```shell
GOARCH=386 go vet ./fieldpath/
```

The random path alphabet in `TestFieldSet` used the untyped constant `2147483648` (2^31) as a list index argument to MakePathOrDie.

On 32-bit platforms (GOARCH=386, arm) Go's int is 32-bit, so 2^31 overflows int and the fieldpath test package fails to compile.

Use `math.MaxInt` instead. On 64-bit platforms this exceeds **2^31**, so it still exercises an index beyond the 32-bit range (the original intent); on 32-bit platforms it is the largest representable int, so the package compiles on every architecture.

### DIFF
--- a/fieldpath/set_test.go
+++ b/fieldpath/set_test.go
@@ -19,6 +19,7 @@ package fieldpath_test
 import (
 	"bytes"
 	"fmt"
+	"math"
 	"math/rand"
 	"testing"
 
@@ -99,7 +100,7 @@ var randomPathMaker = randomPathAlphabet(MakePathOrDie(
 	_V(`\`),
 	_V(`\\`),
 	// Indices
-	1, 2, 3, 4, 0, 5, 100, 2147483648,
+	1, 2, 3, 4, 0, 5, 100, math.MaxInt,
 ))
 
 func BenchmarkFieldSet(b *testing.B) {


fieldpath/set_test.go
```diff
@@ -99,7 +100,7 @@ var randomPathMaker = randomPathAlphabet(MakePathOrDie(
 	_V(`\`),
 	_V(`\\`),
 	// Indices
-	1, 2, 3, 4, 0, 5, 100, 2147483648,
```
liggitt, line 102: would marking this as an int64 explicitly (`int64(2147483648)`) let the value stay the same?

liggitt: we could do `math.MaxInt` to get the largest untyped int, AND int64(2147483648) to make sure a larger than unsized int `int64` works properly on 32-bit platforms

arthurbdiniz: `int64(2147483648)` unfortunately doesn't work here.

`MakePathOrDie` feeds each argument through `MakePath`, which classifies it with a type switch.

The switch matches on the exact dynamic type. There's no case `int64`, and Go doesn't implicitly convert between `int` and `int64` here, so an `int64` argument falls through to default, returns an error, and `MakePathOrDie` panics on it. Because `randomPathMaker` is a package-level var, that panic fires during test-package init, before any test runs:

```
panic: unable to make 2147483648 into a path element
    fieldpath.MakePathOrDie(...) path.go:115
    fieldpath_test.init() set_test.go:58
FAIL sigs.k8s.io/structured-merge-diff/v6/fieldpath
```

liggitt: hmm... looks like the issue is the underlying PathElement is using an unsized int

```
	// Index selects a list element by its index number. The containing
	// object must be an atomic list.
	Index *int
```

Servers are 64-bit, so their unsized ints are already taking that space, and can store/return values greater than 32-bits. I'm not sure about dropping the test that demonstrates a 32-bit client couldn't handle that.

What if we made the index value explicitly an int64?

```diff
diff --git a/fieldpath/element.go b/fieldpath/element.go
index 98ecb1b..4c944b4 100644
--- a/fieldpath/element.go
+++ b/fieldpath/element.go
@@ -45,7 +45,7 @@ type PathElement struct {
 
 	// Index selects a list element by its index number. The containing
 	// object must be an atomic list.
-	Index *int
+	Index *int64
 }
 
 // FieldNameElement creates a new FieldName PathElement.
@@ -74,7 +74,7 @@ func ValueElement(value value.Value) PathElement {
 }
 
 // IndexElement creates a new Index PathElement.
-func IndexElement(index int) PathElement {
+func IndexElement(index int64) PathElement {
 	return PathElement{Index: &index}
 }
 
diff --git a/fieldpath/fromvalue.go b/fieldpath/fromvalue.go
index e89077c..366a0c7 100644
--- a/fieldpath/fromvalue.go
+++ b/fieldpath/fromvalue.go
@@ -64,7 +64,7 @@ func (w *objectWalker) walk() {
 		for iter.Next() {
 			i, value := iter.Item()
 			w2 := *w
-			w2.path = append(w.path, w.GuessBestListPathElement(i, value))
+			w2.path = append(w.path, w.GuessBestListPathElement(int64(i), value))
 			w2.value = value
 			w2.walk()
 		}
@@ -104,7 +104,7 @@ var AssociativeListCandidateFieldNames = []string{
 // referencing by index is acceptable. Currently this is done by checking
 // whether item has any of the fields listed in
 // AssociativeListCandidateFieldNames which have scalar values.
-func (w *objectWalker) GuessBestListPathElement(index int, item value.Value) PathElement {
+func (w *objectWalker) GuessBestListPathElement(index int64, item value.Value) PathElement {
 	if !item.IsMap() {
 		// Non map items could be parts of sets or regular "atomic"
 		// lists. We won't try to guess whether something should be a
diff --git a/fieldpath/path.go b/fieldpath/path.go
index c8b97c4..b50d1bb 100644
--- a/fieldpath/path.go
+++ b/fieldpath/path.go
@@ -86,6 +86,10 @@ func MakePath(parts ...interface{}) (Path, error) {
 		case PathElement:
 			fp = append(fp, t)
 		case int:
+			// TODO: Understand schema and object and convert this to the
+			// FieldSpecifier below if appropriate.
+			fp = append(fp, PathElement{Index: new(int64(t))})
+		case int64:
 			// TODO: Understand schema and object and convert this to the
 			// FieldSpecifier below if appropriate.
 			fp = append(fp, PathElement{Index: &t})
diff --git a/fieldpath/serialize-pe.go b/fieldpath/serialize-pe.go
index b845f24..989be2d 100644
--- a/fieldpath/serialize-pe.go
+++ b/fieldpath/serialize-pe.go
@@ -102,7 +102,7 @@ func DeserializePathElement(s string) (PathElement, error) {
 			return PathElement{}, err
 		}
 		return PathElement{
-			Index: &i,
+			Index: new(int64(i)),
 		}, nil
 	default:
 		return PathElement{}, ErrUnknownPathElementType
@@ -175,7 +175,7 @@ func (pes *pathElementSerializer) serialize(pe PathElement) error {
 		if _, err := pes.builder.Write(peIndexSepBytes); err != nil {
 			return err
 		}
-		if _, err := pes.builder.WriteString(strconv.Itoa(*pe.Index)); err != nil {
+		if _, err := pes.builder.WriteString(strconv.FormatInt(*pe.Index, 10)); err != nil {
 			return err
 		}
 	default:
diff --git a/fieldpath/set.go b/fieldpath/set.go
index 5894b0f..00bfbb8 100644
--- a/fieldpath/set.go
+++ b/fieldpath/set.go
@@ -194,6 +194,9 @@ func PrefixMatcher(parts ...interface{}) (*SetMatcher, error) {
 			// a plain field name
 			pattern = PathElementMatcher{PathElement: PathElement{FieldName: &t}}
 		case int:
+			// a plain list index
+			pattern = PathElementMatcher{PathElement: PathElement{Index: new(int64(t))}}
+		case int64:
 			// a plain list index
 			pattern = PathElementMatcher{PathElement: PathElement{Index: &t}}
 		default:
diff --git a/fieldpath/set_test.go b/fieldpath/set_test.go
index f17e48e..148778d 100644
--- a/fieldpath/set_test.go
+++ b/fieldpath/set_test.go
@@ -965,7 +965,7 @@ func TestSetNodeMapIterate(t *testing.T) {
 	toAdd := 5
 	addedElements := make([]string, toAdd)
 	for i := 0; i < toAdd; i++ {
-		p := i
+		p := int64(i)
 		pe := PathElement{Index: &p}
 		addedElements[i] = pe.String()
 		_ = set.Descend(pe)
diff --git a/typed/validate.go b/typed/validate.go
index 1da8e1f..144fd76 100644
--- a/typed/validate.go
+++ b/typed/validate.go
@@ -130,7 +130,7 @@ func (v *validatingObjectWalker) visitListItems(t *schema.List, list value.List)
 		defer v.allocator.Free(child)
 		var pe fieldpath.PathElement
 		if t.ElementRelationship != schema.Associative {
-			pe.Index = &i
+			pe.Index = new(int64(i))
 		} else {
 			var err error
 			pe, err = listItemToPathElement(v.allocator, v.schema, t, child)
```

@jpbetz, thoughts on that?

